PR #5, opened 2022-03-24
Add h2o benchmarks

ncclementi:
- [ ] Partially resolves https://github.com/coiled/coiled-distribution/pull/5

This is PR includes GHA that creates two software environments (py38 and py39) equivalent to the coiled distribution package and runs some of the groupby operations of the h2o benchmark for the 0.5 GB dataset on a coiled cluster. Currently, I'm just printing in the action two different times: time to get a cluster, and time to run all the operations. This is definitely something to change

Pending:
- [x] Get a coiled bot account to run this (currently using my token)
- [x] Run this with all the data sets on s3
- [x] Include all the operations of the groupby benchmark. I was just testing a few
- [x] Currently, I'm using the default instance as it's enough for the small data, Revisit this.  

Decisions to make 
- [x] how to profile / benchmark 
- [ ] how to report results of the benchmark.

For the report of the results @ian-r-rose  mentioned the possibility of dumping things on html and using a githubpage. I like this idea, but we need to think it trhough once we have the resutls. 

cc: @hayesgb


ncclementi:
@jrbourbeau I added the remaining queries but in a very hacky way, it's not neat and clean. But since we will be changing this and probably use parametrize I just put them in there to have a reference. 

But here is the first benchmark by task-group, now let's make this code pretty. 
https://github.com/coiled/coiled-runtime/runs/5981303681?check_suite_focus=true#step:6:15

ncclementi:
@jrbourbeau 

So it looks like it shugged at the 50GB with only 4 workers. The clusters are still running, I'm going to close them manually. Not sure what exactly happened, in the potential issues I see for the scheduler:

```
[ 6898.654172] cloud-init[1003]: Exception in callback None()
[ 6898.656181] cloud-init[1003]: Traceback (most recent call last):
[ 6898.661226] cloud-init[1003]: ConnectionResetError: [Errno 104] Connection reset by peer
[ 6898.661446] cloud-init[1003]: During handling of the above exception, another exception occurred:
[ 6898.661701] cloud-init[1003]: Traceback (most recent call last):
[ 6898.676147] cloud-init[1003]:     self._ssl_connect_future.exception()
[ 6898.676836] cloud-init[1003]: asyncio.exceptions.CancelledError
[ 6898.677020] cloud-init[1003]: Exception in callback None()
[ 6898.677474] cloud-init[1003]: Traceback (most recent call last):
[ 6898.681917] cloud-init[1003]: ConnectionResetError: [Errno 104] Connection reset by peer
[ 6898.689927] cloud-init[1003]: During handling of the above exception, another exception occurred:
[ 6898.692103] cloud-init[1003]: Traceback (most recent call last):
[ 6898.705029] cloud-init[1003]:     self._ssl_connect_future.exception()
[ 6898.705259] cloud-init[1003]: asyncio.exceptions.CancelledError
[ 6898.705459] cloud-init[1003]: Exception in callback None()
[ 6898.707976] cloud-init[1003]: Traceback (most recent call last):
[ 6898.712932] cloud-init[1003]: ConnectionResetError: [Errno 104] Connection reset by peer
[ 6898.713258] cloud-init[1003]: During handling of the above exception, another exception occurred:
[ 6898.713463] cloud-init[1003]: Traceback (most recent call last):
[ 6898.727327] cloud-init[1003]:     self._ssl_connect_future.exception()
[ 6898.728015] cloud-init[1003]: asyncio.exceptions.CancelledError
```

jrbourbeau:
Temporarily commenting out the 50 GB dataset for now. I'm curious to see how long the 5 GB tests took (we'll get a report after the test suite finishes)

jrbourbeau:
FWIW I think the only thing left for this PR is to determine which datasets we want to run over. @ncclementi could you take a look at the current state of things here when you get a chance?

ncclementi:
@jrbourbeau This is looking good to me, The only thing I'd suggest to see if all the tests finish is bumping from 4 to 10 workers to get all the data running. But if this is not something we want to do. I'm ok skipping the 50GB case. 

It looks like the slowest case is q8 5GB taking approximately 10min
https://github.com/coiled/coiled-runtime/runs/6000314198?check_suite_focus=true#step:7:34

I wonder if there is a chance to get the report to be more readable? Currently, we have the report order by slowest durations, I wonder if this is what we want or something else. We don't need to solve this now, though.

tests/test_h2o_benchmarks.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.fixture(scope="module")
+def cluster():
+    with coiled.Cluster(
+        software=runtime_software_env,
+        name="h2o-groupby-benchmark_" + str(uuid.uuid4()),
+        account="dask-engineering",
+        n_workers=4,
+        backend_options={"spot": False},
+    ) as cluster:
```
mrocklin, line 24: If possible I recommend using Coiled Declarative.  It's significantly more reliable.

jrbourbeau: Being handled separately over in https://github.com/coiled/coiled-runtime/pull/34

jrbourbeau:
> The only thing I'd suggest to see if all the tests finish is bumping from 4 to 10 workers to get all the data running

I'll push a commit up to try this. If the 50 GB dataset doesn't work / if the 5 GB datasets works, but takes a long time, let's just temporarily comment them out for now and address in a follow-up PR.

> I wonder if there is a chance to get the report to be more readable? Currently, we have the report order by slowest durations, I wonder if this is what we want or something else. We don't need to solve this now, though.

We'll definitely want things reported different in the future, but for now I think this is fine. While we're getting things off the ground, I'm most curious about answering "What tests are taking the longest?" / "How long does each CI build take?", which the current format helps with

jrbourbeau:
CI builds were timing out after 120 minutes when the 50 GB dataset was included. I've re-commented that dataset out for now as I'd like to continue to iterate quickly at this stage, so would like to see us have faster CI builds. Re-introducing the 50 GB dataset is just a couple of lines of code away -- let's reevaluate the value of adding it back as a follow-up item. 

Let's see how long the 5 GB dataset takes with a 10-worker cluster.

ncclementi:
WARNING! DO NOT MERGE
This is looking green although I don't see the h2o clusters running on coiled. I wonder if this is just running on the CI box. For example this is running at the moment https://github.com/coiled/coiled-runtime/runs/6015014233?check_suite_focus=true 

But there are no clusters in the coiled dashboard. I can't see the other clusters either.
cc: @jrbourbeau

jrbourbeau:
Hrm, I thought I saw those clusters earlier -- thanks for surfacing @ncclementi. I just pushed an empty commit to see if we can reproduce

jrbourbeau:
I see Coiled cluster spinning up for this action

tests/test_h2o_benchmarks.py
```diff
@@ -0,0 +1,129 @@
...
+
+@pytest.fixture(scope="module")
+def cluster(runtime_software_env):
+    with Cluster(
+        software=runtime_software_env,
+        name="h2o-groupby-benchmark_" + str(uuid.uuid4()),
+        account="dask-engineering",
+        n_workers=10,
+    ) as cluster:
+        yield cluster
```
mrocklin, line 22: I didn't see this until after I submitted https://github.com/coiled/coiled-runtime/pull/40

I recommend the changes there for a few reasons:

1.  software seems simpler than runtime_software_env, unless we think that there are going to be other kinds of software environments
2. we should specify VM types I think.  They an float otherwise and get different kinds of performance
3. I think that we should just use generic names, but I'm open to other thoughts.

However, I also see that you want a module-level cluster.  If so then yeah, this should maybe be separate.

ncclementi:
> I see Coiled cluster spinning up for this PR
> 
> EDIT: But I should note that running / pending clusters aren't always at the top of the cluster dashboard

It might have been that they were way down the list. It seems the order of the clusters is whatever was launched latest at the top, even if it finishes. I think once this CI completes we should be able to merge.

jrbourbeau:
> I think once this CI completes we should be able to merge

+1. I temporarily commented out the 5 GB dataset in order to keep CI build times down while we're quickly iterating on this repo. We can re-introduce them in the future easily

PR #19, opened 2022-04-07
Update `README` and move recipe location

jrbourbeau:
This PR:

- Refreshes the `README` by updating `coiled-distribution` references to `coiled-runtime` along with a few other (minor) changes
- Moves the `recipe/` directory to the root of the repo. As this content isn't strictly concerned with CI, it feels weird to have it in a `continuous_integration/` directory


README.md
```diff
@@ -1,25 +1,50 @@
-# Coiled Distribution
+# Coiled Runtime
```
ncclementi, line 2: Should we include the CI badges?  Not sure if we can do something like this?
```suggestion

[![Conda Build](https://github.com/coiled/coiled-runtime/actions/workflows/conda.yml/badge.svg)](https://github.com/coiled/coiled-runtime/actions/workflows/conda.yml) [![Linting](https://github.com/coiled/coiled-runtime/actions/workflows/pre-commit.yml/badge.svg)](https://github.com/coiled/coiled-runtime/actions/workflows/pre-commit.yml)
```

ncclementi (approved):
I had only one suggestion add the CI badges, outside that it looks great to me. Thanks @jrbourbeau

README.md
jrbourbeau: Adding badges would be good. I held off on this because currently the conda build badge is red. I think this is because the most recent GHA build for `main` was cancelled due to us also pushing a release tag. I'd like to see that fixed before adding badges

ncclementi: Whoops, didn't notice that. That makes sense. I'll go ahead and merge this one then

PR #26, opened 2022-04-11
Update GitHub actions cancelling condition

jrbourbeau:
This matches what's used over in `distributed` and, I think, might fix some of the GHA job cancellation issues were were seeing earlier (xref https://github.com/coiled/coiled-runtime/pull/19#discussion_r846298772) 

cc @ncclementi 

EDIT: See https://docs.github.com/en/github-ae@latest/actions/learn-github-actions/contexts#github-context for the difference between `github.head_ref` and `github.ref`


ncclementi:
Thanks, @jrbourbeau This makes sense, I'll merge it right when the CI passes.

PR #27, opened 2022-04-11
[WIP] Test against multiple runtimes

jrbourbeau:
This adds the logic necessary for us to run our existing CI builds against multiple `coiled-runtime` versions. This way we can make sure that any tests, benchmarks, etc. which are added here run against the `latest` `coiled-runtime` version (i.e. whatever is in the `main` branch of this repo) as well as all currently support `coiled-runtime` releases.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -51,7 +51,7 @@ jobs:
         run: |
           if [[ ${{ matrix.runtime-version }} = 'latest' ]]
           then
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Going to close this PR and open a new one that's not from a fork, so it will have access to repo secrets...

jrbourbeau:
Going to close this PR and open a new one that's not from a fork, so it will have access to repo secrets...

PR #29, opened 2022-04-12
Add nyc quickstart to the test suite

ncclementi:


tests/test_quickstart.py
```diff
@@ -0,0 +1,36 @@
+import os
```
jrbourbeau, line 1: Let's rename this file `test_coiled.py` as we'll want to add more related tests here shortly

tests/test_quickstart.py
```diff
@@ -0,0 +1,36 @@
...
+        "RatecodeID": "UInt8",
+    },
+    storage_options={"anon": True},
+    blocksize="16 MiB",
+).persist()
+
+result = ddf.groupby("passenger_count").tip_amount.mean().compute()
+
+assert isinstance(result, pandas.core.series.Series)
+assert not result.empty
```
jrbourbeau, line 36: Let's wrap this all in a `def test_quickstart():` function

jrbourbeau (commented):
Thanks for working on this @ncclementi

jrbourbeau:
Looks like we're getting `Invalid Coiled token encountered.` in CI

.github/workflows/conda.yml
```diff
@@ -53,6 +54,9 @@ jobs:
           mamba env export | grep -E -v '^prefix:.*$'
 
       - name: Run tests
+        env:
+            DASK_COILED__TOKEN: ${{ secrets.COILED_BENCHMARK_BOT_TOKEN }}
```
jrbourbeau, line 58: Unfortunately I don't think secrets are available from forks. Can you re-submit this PR with a branch directly from the `coiled/coiled-runtime` repo?

ncclementi: This makes sense, I forgot about this. Thanks for pointing this out. Will do!

ncclementi:
Closing as we can't test secrets from a fork.

PR #30, opened 2022-04-12
Basic stability test outline

sjperkins:
This introduces a basic dataframe shuffle stability test outline. Note that this is not the final version.


sjperkins:
Closing in favour of #36 which has access to a coiled account

PR #36, opened 2022-04-13
Basic stability test

sjperkins:


tests/stability/test_stability.py
```diff
@@ -0,0 +1,32 @@
...
+
+import coiled
+import dask
+from dask.distributed import Client
+
+SOFTWARE = f"dask-engineering/coiled_dist-py{sys.version_info[0]}{sys.version_info[1]}"
+
+
+def test_shuffle_stability():
+    with ExitStack() as stack:
```
jrbourbeau, line 13: I'm curious why `ExitStack` is needed. Naively I would have expected us to just use 

```python
with coiled.Cluster(...) as cluster:
    with Client(cluster) as client:
        ...
```

tests/stability/test_stability.py
```diff
@@ -0,0 +1,32 @@
...
+
+        client = stack.enter_context(Client(cluster))  # noqa F841
+        df = dask.datasets.timeseries(
+            start="2000-01-01", end="2000-12-31", freq="1s", partition_freq="1D"
+        )
+
+        sdf = df.shuffle(on="x")
+
+        sdf_size = sdf.memory_usage(index=True).sum() / (1024.0**3)
+        assert sdf_size.compute() > 1.0  # 1.0 GB
```
jrbourbeau, line 32: Is this memory usage check meaningful? Or are we just wanting to trigger the shuffle computation?

jrbourbeau (commented):
Thanks for pushing this up @sjperkins!

tests/stability/test_stability.py
sjperkins: This is a personal preference on my part to avoid deeply nested with statements, but I don't have a strong opinion if there are objections to the pattern.

tests/stability/test_stability.py
sjperkins: It's not meaningful at present. I wanted to check that a reasonable amount of data had been created, but not so much that the shuffle takes a long time at this stage and blocks the CI at this stage.

sjperkins:
Closes https://github.com/coiled/oss-engineering/issues/92

tests/stability/test_stability.py
mrocklin: Yeah, personally I don't know how ExitStack works, so this code would be fairly inaccessible to me (unless I go and learn things, which I probably should).  More generally, we should endeavor to [write dumb code](https://matthewrocklin.com/blog/work/2018/01/27/write-dumb-code) I think

tests/stability/test_stability.py
```diff
@@ -0,0 +1,28 @@
...
+
+SOFTWARE = f"dask-engineering/coiled_dist-py{sys.version_info[0]}{sys.version_info[1]}"
+
+
+def test_shuffle_stability():
+    with coiled.Cluster(
+        software=SOFTWARE,
+        name="coiled-stability-tests_" + str(uuid.uuid4()),
+        account="dask-engineering",
+        n_workers=10,
```
ncclementi, line 16: Out of curiosity do we need 10 workers here? I know we use 10 for the quickstart because is a decent chunk of data, but can we run this with just `n_workers=4`?

tests/stability/test_stability.py
```diff
@@ -0,0 +1,28 @@
...
+import dask
+from dask.distributed import Client
+
+SOFTWARE = f"dask-engineering/coiled_dist-py{sys.version_info[0]}{sys.version_info[1]}"
+
+
+def test_shuffle_stability():
+    with coiled.Cluster(
+        software=SOFTWARE,
+        name="coiled-stability-tests_" + str(uuid.uuid4()),
```
ncclementi, line 14: All of the tests will run on coiled, so we can probably remove the word coiled. Just a detail.

mrocklin:
There are some improvements that I would make here, but I also think that this is better than what we had before, so I'm +1 :)

ncclementi:
> I don't understand why the dask type dispatcher doesn't understand a pandas Dataframe in macos-latest Python 3.8. Small traceback snippet below:
> 
> ```
>  /opt/conda/envs/coiled/lib/python3.8/site-packages/dask/layers.py:36: in __call__
>     ???
> /opt/conda/envs/coiled/lib/python3.8/site-packages/dask/dataframe/core.py:104: in _concat
>     ???
> /opt/conda/envs/coiled/lib/python3.8/site-packages/dask/dataframe/core.py:7176: in has_parallel_type
>     ???
> /opt/conda/envs/coiled/lib/python3.8/site-packages/dask/utils.py:622: in __call__
>     ???
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> >   ???
> E   TypeError: No dispatch for <class 'pandas.core.frame.DataFrame'>
> ```
> 
> Does anyone have any thoughts here?


@sjperkins It looks like the fail is on Ubuntu and py38 actually, but this is very strange to me
https://github.com/coiled/coiled-runtime/runs/6011102330?check_suite_focus=true

@jrbourbeau Do you have any ideas on what could be happening?

jrbourbeau (approved):
Thanks @sjperkins! There were a bunch of changes yesterday that introduced some merge conflicts here. I pushed directly to this branch to resolve them -- hope you don't mind.

PR #34, opened 2022-04-12
switch to v2 clusters (declarative)

dchudz:
It would help out the clusters team if you can use v2 clusters -- those are the clusters under active development, so it's better to be dogfooding that.

This removes the `spot=False` b/c (1) that's the default anyway, and (2) v2 clusters don't support that argument quite yet.

Closes https://github.com/coiled/coiled-runtime/issues/35

PR #39, opened 2022-04-13
Use declarative clusters

mrocklin:
Following on from #34 

Closes https://github.com/coiled/coiled-runtime/issues/35


jrbourbeau (approved):
For those following along, I confirmed this `import` works on both older (0.0.71) and newer (0.0.73) versions of `coiled`

PR #40, opened 2022-04-13
Move client and cluster creation to fixture

mrocklin:


mrocklin:
I do not currently understand why the failing tests fail.  Maybe a bug?

```
Software environment deleted successfully.
Error: Process completed with exit code 1.
```

ncclementi:
> I do not currently understand why the failing tests fail. Maybe a bug?
> 
> ```
> Software environment deleted successfully.
> Error: Process completed with exit code 1.
> ```

This deletion of Software environment happens only for when we run with "latest" see GHA code added on this PR https://github.com/coiled/coiled-runtime/pull/33

I don't think it's a bug. I think there is something about the fixture  that it's not quite working  see https://github.com/coiled/coiled-runtime/runs/6015475603?check_suite_focus=true#step:8:1

I'm off for the day as it's late for me, but things seem to be working outside this PR, so I don't think it's a bug on main.

jrbourbeau:
> I do not currently understand why the failing tests fail. Maybe a bug?

This is intentional. If you look at the section in the build where tests are run, you'll see they're failing with 

```python
 E       fixture 'client' not found
```

Unfortunately this isn't obvious because we want to keep continuing to the next step in the workflow so we can delete the software environment we created earlier on in the workflow

https://github.com/coiled/coiled-runtime/blob/7ec5f08c1a6d21b881249ec9897f8c37ff2057af/.github/workflows/conda.yml#L92-L94

I still wanted to overall CI build to fail if tests fail so I added 

https://github.com/coiled/coiled-runtime/blob/7ec5f08c1a6d21b881249ec9897f8c37ff2057af/.github/workflows/conda.yml#L118-L121

which will `exit 1` if the test suite failed. I'll think about how this can be improved (suggestions welcome)

conftest.py
```diff
@@ -21,8 +24,29 @@ def pytest_collection_modifyitems(config, items):
...
         "COILED_SOFTWARE_NAME",
         f"dask-engineering/coiled_dist-py{sys.version_info[0]}{sys.version_info[1]}",
     )
+
+
+@pytest.fixture(scope="module")
+def small_cluster(software, request):
+    module = os.path.basename(request.fspath).split(".")[0]
+    with Cluster(
+        name=f"{module}-{uuid.uuid4().hex}",
```
jrbourbeau, line 38: This creates clusters with names like

```
test_h2o_benchmarks-95ccc91cc81f4ffe8816ee30f5bc36ba
```

and 

```
test_coiled-fd7000ad6fe340139f1ebcb43848ab03
```

jrbourbeau (approved):
This looks like a nice addition to me, thanks @mrocklin

PR #41, opened 2022-04-13
Add basic array tests

mrocklin:
So, the in-memory one works great.  The out-of-memory one not so much.  I'm skipping for now because I suspect that this will kill things.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import uuid
 
 import pytest
 from coiled._beta import ClusterBeta as Cluster
@@ -22,7 +23,7 @@ def pytest_collection_modifyitems(config, items):
             item.add_marker(skip_latest)
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
 def software():
     return os.environ.get(
         "COILED_SOFTWARE_NAME",
@@ -30,9 +31,11 @@ def software():
     )
 
 
-@pytest.fixture
-def small_cluster(software):
+@pytest.fixture(scope="module")
+def small_cluster(software, request):
+    module = os.path.basename(request.fspath).split(".")[0]
     with Cluster(
+        name=f"{module}-{uuid.uuid4().hex[:8]}",
         software=software,
         account="dask-engineering",
         n_workers=10,
@@ -45,4 +48,7 @@ def small_cluster(software):
 @pytest.fixture
 def small_client(small_cluster):
     with Client(small_cluster) as client:
+        small_cluster.scale(10)
+        client.wait_for_workers(10)
+        client.restart()
         yield client
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,9 @@
+import os
+import sys
+
 import pytest
+from coiled._beta import ClusterBeta as Cluster
+from dask.distributed import Client
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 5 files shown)


conftest.py
```diff
@@ -15,3 +20,29 @@ def pytest_collection_modifyitems(config, items):
...
+@pytest.fixture
+def small_cluster(software):
+    with Cluster(
+        software=software,
+        account="dask-engineering",
+        n_workers=10,
+        worker_vm_types=["m5.large"],
+        scheduler_vm_types=["m5.large"],
+    ) as cluster:
+        yield cluster
```
gjoseph92, line 42: What shuts down the cluster when we're done?

conftest.py
```diff
@@ -15,3 +20,29 @@ def pytest_collection_modifyitems(config, items):
...
+@pytest.fixture
+def software():
+    return os.environ.get(
+        "COILED_SOFTWARE_NAME",
+        f"dask-engineering/coiled_dist-py{sys.version_info[0]}{sys.version_info[1]}",
+    )
+
+
+@pytest.fixture
+def small_cluster(software):
```
gjoseph92, line 34: Do we want to allow multiple tests to use the same cluster? Should be be restarted each time? Recreated each time? Wondering about fixture scope.

conftest.py
```diff
@@ -15,3 +20,29 @@ def pytest_collection_modifyitems(config, items):
...
+    with Cluster(
+        software=software,
+        account="dask-engineering",
+        n_workers=10,
+        worker_vm_types=["m5.large"],
+        scheduler_vm_types=["m5.large"],
```
gjoseph92, line 40:
```suggestion
        scheduler_vm_types=["m5.large"],
        wait_for_workers=True,
```
Debatable. Leaving it out increases chaos. Adding it increases determinism. Looks like the default is 0.3.

conftest.py
mrocklin: The context manager hopefully.  Cluster.__exit__ should call Cluster.close

conftest.py
mrocklin: Maybe.  This is what they're doing with the H2O benchmarks.  We'll want things clean, but maybe we can achieve that with restart, scale, and wait_for_workers.

The platform integration tests do a single cluster per test, and they just run all of the tests in parallel.  It's less efficient, but works ok.

I'm open here.  I think that we'll probably want to iterate.

conftest.py
mrocklin: This repository is used both for stability and for performance testing.  My guess is that we're going to want to split this out pretty soon to different options, but for now I don't care that much.

conftest.py
jrbourbeau: I'm suggesting we start with module-scoped cluster fixtures + function-scoped client fixtures with restarts over in https://github.com/coiled/coiled-runtime/pull/40. This is something that seems to work for our current needs and, as @mrocklin said, we can iterate as needed. 

+1 for parallelizing tests. Opening a separate issue for that.

mrocklin: In platform integration tests we asked pytest-xdist to parallelize between modules, but not within them.  That would work will with the module-scoped cluster fixtures.

mrocklin:
@jrbourbeau ok to merge?

ncclementi:
@mrocklin This looks good to me, I quickly check on the `test_rechunk_out_of_memory` I noticed that with the default instances and 10 workers and it is chugging. But I haven't tried with the `"m5.large"` I wonder if we want to keep skipping this or take a different approach. 

Screen Shot 2022-04-14 at 11 26 51 AM
Screen Shot 2022-04-14 at 11 27 46 AM

mrocklin:
What sort of different approach would you propose?  Run it?  Change the test?

ncclementi:
> What sort of different approach would you propose? Run it? Change the test?

1. If this is a test that is expected to run with the Cluster as is, but it doesn't we can skip it and open an issue. 
2. If the only reason why this doesn't run is that there is not enough memory then:
    a. change the test and make x smaller.
    b. use a different cluster configuration

cc: @jrbourbeau Do you have any comments? If not I'd be ok merging it as is and open a separate issue to sort out the reason why the test chugs.

mrocklin:
> If the only reason why this doesn't run is that there is not enough memory then:
> a. change the test and make x smaller.
> b. use a different cluster configuration

Yeah, so my intention here is to run a computation that is uncomfortably large for Dask.  Dask should grow so that is can handle it gracefully.  Thus the skip + issue approach

jrbourbeau (approved):
Let's go with the skip + issue approach

PR #46, opened 2022-04-14
Groupby Agg

hayesgb:
This turns out to really stress the system.  At both small and large partition sizes, I'm seeing killed workers with some regularity.  At 100MB partition sizes, it sometimes completes, and sometimes ends up with tasks being unevenly distributed on just a few workers

This is with 10MB partitions.  Started with 10 workers and now I'm down to 4.
Screen Shot 2022-04-14 at 9 27 29 AM
.


tests/test_groupby_agg.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@pytest.mark.parametrize("partition_size", ["10MB", "100MB", "1GB"])
+def test_quickstart(partition_size):
+    with Cluster(
+        software=SOFTWARE,
+        name="test_groupby_agg" + str(uuid.uuid4()),
+        account="dask-engineering",
+        n_workers=10,
+    ) as cluster:
```
mrocklin, line 23: See the recent fixtures work. You might want to check out newer PRs like #41 for a model.

tests/test_groupby_agg.py
```diff
@@ -0,0 +1,56 @@
...
+                    "passenger_count": ["sum", "std"],
+                    "PULocationID": ["count"],
+                    "total_amount": ["min"],
+                    "travel_time": ["mean"],
+                }
+            )
+            ddf2.columns = ddf2.columns.droplevel(0)
+            mem = ddf2.memory_usage(deep=True).sum().compute()
+
+            assert hasattr(ddf2, "dask")
```
mrocklin, line 55: ?

tests/test_groupby_agg.py
hayesgb: Updated

tests/test_groupby_agg.py
hayesgb: Removed.

jrbourbeau:
There have been several updates since this PR was open, so I took the liberty of merging `main`. Hope you don't mind @hayesgb

PR #52, opened 2022-04-14
Make sure when tests fail CI jobs always fail

jrbourbeau:
Currently we're setting `continue-on-error: true` for our testing step so we can make sure we delete any Coiled software environments we created as part of a push to `main` or a PR. However, since we only ever create Coiled software environments on the latest build, we should only set `continue-on-error: true` on latest builds. Otherwise, we end up with tests failing and a green check mark for the GHA build. 

Closes https://github.com/coiled/coiled-runtime/issues/42

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -105,7 +105,7 @@ jobs:
             export EXTRA_OPTIONS=" "
           fi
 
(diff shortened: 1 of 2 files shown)


jrbourbeau:
Looks like this thing works. Will merge after CI is green

ncclementi (approved):

PR #53, opened 2022-04-14
[WIP] Run tests in parallel

jrbourbeau:
Closes https://github.com/coiled/coiled-runtime/issues/48

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -44,7 +44,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install dependencies
-        run: mamba install boa conda-verify jinja2 packaging pytest
+        run: mamba install boa conda-verify jinja2 packaging pytest pytest-xdist
 
       - name: Install coiled-runtime
         run: |
@@ -105,7 +105,7 @@ jobs:
             export EXTRA_OPTIONS=" "
           fi
 
-          python -m pytest $EXTRA_OPTIONS tests
+          python -m pytest -n auto --dist loadscope $EXTRA_OPTIONS tests
 
       - name: Remove Coiled software environment
         if: ${{ matrix.runtime-version == 'latest' }}


.github/workflows/conda.yml
```diff
@@ -105,7 +105,7 @@ jobs:
             export EXTRA_OPTIONS=" "
           fi
 
-          python -m pytest -n auto --dist loadscope $EXTRA_OPTIONS tests
+          python -m pytest -n 4 --dist loadscope $EXTRA_OPTIONS tests
```
ncclementi, line 108: Interesting to see that `auto` is only giving us 2.

mrocklin: We probably only have 2 cores, but almost all of our computation is happening remotely.  The platform cranks this up pretty high, into the 10s I think.

jrbourbeau: Yeah, `auto` giving us 2 means we have 2 cores on the GHA runners. Parallelizing has gotten tests (just the `pytest` part of the full CI build) down to ~5 minutes compared to ~12 minutes on `main` :tada:. Bumping the number of `pytest-xdist` workers from 2 to 4 didn't change the overall test runtime because we're currently bottlenecked by the h2o benchmarks (which is the module that takes the longest to run at ~5 minutes). I'm going to bump to `-n 10` to make things a bit more future-proof as we add more test modules, but otherwise I think this is good to go

mrocklin:
Nice.

jrbourbeau:
We're seeing 

```python
RuntimeError: There is no current event loop in thread 'Dummy-1'.
```

in CI. I suspect we're running into https://github.com/pytest-dev/pytest-xdist/issues/620 (cc @graingert who is engaged on that `pytest-xdist` issue).

mrocklin:
I've just pushed up what we use in Coiled.  Maybe it helps?

jrbourbeau:
Thanks! Let's give it a try -- I pushed a commit to install `pytest-asyncio` as I think we need that for the `event_loop` fixture to take effect

jrbourbeau:
@graingert if you have a moment to look at this, that would be appreciated

graingert:
looks like there's two bugs here:

https://github.com/coiled/coiled-runtime/runs/6039817847?check_suite_focus=true#step:8:133 looks like both the pytest-xdist bug and a legit code bug

it's not valid to create an `asyncio.Event()` with the `asynchronous=False` flag in the `__init__` as it will be bound to the wrong loop

conftest.py
```diff
@@ -57,6 +59,16 @@ def get_software():
             )
 
 
+@pytest.fixture(scope="session")
+def event_loop():
+    if threading.current_thread() is not threading.main_thread():
+        os.exit(1)
```
jrbourbeau, line 65: I added this as a workaround based on a suggestion from @graingert. All tests passed on the most recent commit, I'm going to run tests a few more times to make sure things work consistently (previous failures only happened occasionally)

jrbourbeau:
Alright, tests have all passed for 4 workflow runs in a row. I'm going to merge this in -- thanks for your help @graingert

jrbourbeau:
IIRC we tried that but still ran into https://github.com/pytest-dev/pytest-xdist/issues/620

PR #57, opened 2022-04-14
Solve with latest `coiled` release

jrbourbeau:
Updating this `click` pinning leads to me getting the latest `coiled==0.0.73` release when I build `coiled-runtime` locally and then attempt to install it. We need this `click` pin as `click=8.1.0` broke `distributed`. We've since included a fix in `distributed` itself (xref https://github.com/dask/distributed/pull/6018), but we can't use that version of `distributed` as it has known stability issues. 

TODO:

- [x] Add a test to make sure we always default to using the latest `coiled` release

Closes https://github.com/coiled/coiled-runtime/issues/22

### DIFF
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -1,12 +1,16 @@
 from __future__ import annotations
 
 import pathlib
+import shlex
+import subprocess
 
+import coiled
 import conda.cli.python_api as Conda
 import pytest
 import yaml
 from jinja2 import Environment, FileSystemLoader, select_autoescape
 from packaging.requirements import Requirement, SpecifierSet
+from packaging.version import InvalidVersion, Version
 
 
 def get_conda_installed_versions() -> dict[str, str]:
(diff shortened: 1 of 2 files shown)


dchudz:
cool

ncclementi:
Should we add this to conda-forge and make a new release or not yet?

hayesgb:
Nice

tests/test_build.py
```diff
@@ -49,3 +53,23 @@ def test_install_dist():
         specifier.contains(installed_versions[package])
         for package, specifier in meta_specifiers.items()
     )
+
+
+@pytest.mark.latest_runtime
+def test_latest_coiled():
```
jrbourbeau, line 59: This is a little gnarly, but works. To minimize local test suite failures, I'm marking this to only run on our latest `coiled-runtime` build. I think this is okay for now, we can always relax that contsraint later if needed

jrbourbeau:
> Should we add this to conda-forge and make a new release or not yet?

Yeah, let's update conda-forge / release after this PR and https://github.com/coiled/coiled-runtime/pull/54 are merged

PR #58, opened 2022-04-14
Raise issue when CI on `main` fails

### LINKED ISSUE #43: Raise issue when CI on `main` fails
If CI fails for some reason on a push to `main`, it'd be good for us to have that visible (today there's just a red X that most folks won't notice). Dask's `upstream` CI build opens an issue when there's a failure on `main` and leaves a comment linking to the corresponding CI build. We can reuse that GHA workflow here.

ncclementi:
This probably has problems, and I'm not quite sure how to test it works. But dropping something here to get it started

- [x] Closes #43

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -105,7 +105,7 @@ jobs:
             export EXTRA_OPTIONS=" "
           fi
 
-          python -m pytest $EXTRA_OPTIONS tests
+          python -m pytest $EXTRA_OPTIONS tests/test_build.py::test_bad
 
       - name: Remove Coiled software environment
         if: ${{ matrix.runtime-version == 'latest' }}
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -165,9 +165,9 @@ jobs:
               const workflow_url = `https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}`
               const issue_body = `[Workflow Run URL](${workflow_url})`
               // Run GraphQL query against GitHub API to find the most recent open issue used for reporting failures
-              const query = `query($owner:String!, $name:String!, $creator:String!, $label:String!){
+              const query = `query($owner:String!, $name:String!, $creator:String!){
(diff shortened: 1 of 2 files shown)


ncclementi:
I'm trying to make this Fail on purpose and also making a condition to report on PR, to see if it works. But It seems the report workflow didn't run and it was skipped. Not sure what I'm doing wrong yet, any ideas are welcome!

.github/workflows/conda.yml
```diff
@@ -140,4 +140,64 @@ jobs:
...
+          anaconda upload dist/conda/noarch/*.tar.bz2
+
+  report:
+    name: report
+    needs: conda
+    if: |
+      always()
+      && github.event_name == 'pull_request'
+      && github.repository == 'coiled/coiled-runtime'
+      && needs.conda.steps.test.outcome == 'failure'
```
jrbourbeau, line 152: xref https://github.com/dask/dask/blob/bfc76afdcdc43c575a10ffefda94aaba424fe347/.github/workflows/upstream.yml#L75

```suggestion
      && needs.conda.result == 'failure'
```

jrbourbeau:
✨ https://github.com/coiled/coiled-runtime/issues/59 ✨

.github/workflows/conda.yml
```diff
@@ -140,4 +140,64 @@ jobs:
...
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v2
+      - name: Report failures
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const title = "⚠️ Main CI failed ⚠️"
```
jrbourbeau, line 164: Nit: Even though this has less content, it somehow seems more clear to me. What do you think? 

```suggestion
            const title = "⚠️ CI failed ⚠️"
```

ncclementi:
Reverted the Intentional Fail. 
Once this is green I think it is ready to merge, unless we need to add anything.

jrbourbeau (approved):
This looks great -- thanks @ncclementi!

PR #61, opened 2022-04-15
Declarative clusters only 

jrbourbeau:
I hadn't realized how frequently we'd run into errors with using non-declarative Coiled cluster (good job declarative team!). Now that we're pulling in the latest `coiled` by default, let's just switch back to only using new v2 clusters

xref https://github.com/coiled/coiled-runtime/issues/60

cc @dchudz

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,19 +2,9 @@ import os
 import sys
 import uuid
 
-import coiled
 import pytest
+from coiled.v2 import ClusterBeta as Cluster
 from dask.distributed import Client
-from packaging.version import Version
-
-# Use non-declarative clusters on Python 3.7 with `coiled <= 0.0.73`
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -2,20 +2,15 @@
...
-    "0.0.73"
-):
-    from coiled import Cluster
-else:
+try:
+    from coiled.v2 import Cluster
+except ImportError:
     from coiled._beta import ClusterBeta as Cluster
 
+from dask.distributed import Client
```
jrbourbeau, line 12: Forgot the existing `coiled-runtime=0.0.3` release doesn't contain https://github.com/coiled/coiled-runtime/pull/57 so is still pulling in the an older version of `coiled`

PR #64, opened 2022-04-15
Make Coiled software environment name more robust in CI

jrbourbeau:
We want the software environment to be different for pushes. This PR also makes the names a bit more compact

Closes https://github.com/coiled/coiled-runtime/issues/63

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -78,14 +78,15 @@ jobs:
           # If testing the latest `coiled-runtime`, we need to build a Coiled software environment
           # that can be used when running tests
 
-          # Coiled software environment names can't contain "." characters
+          # Coiled software environment names can't contain "." or uppercase characters
           export PYTHON_VERSION_FORMATTED=$(echo "${{ matrix.python-version }}" | sed 's/\.//g' )
+          export OS_FORMATTED$(python -c "import os; print(os.environ['RUNNER_OS'].lower())")
 
           if [[ ${{ github.event_name }} = 'pull_request' ]]
           then
-            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-${{ github.event.number }}-$GITHUB_RUN_ID-$RUNNER_OS-py$PYTHON_VERSION_FORMATTED
+            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-${{ github.event.number }}-$GITHUB_RUN_ID-$OS_FORMATTED-py$PYTHON_VERSION_FORMATTED
           else
-            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-$GITHUB_REF_NAME-$GITHUB_RUN_ID-$RUNNER_OS-py$PYTHON_VERSION_FORMATTED
+            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-$GITHUB_REF_NAME-$GITHUB_RUN_ID-$OS_FORMATTED-py$PYTHON_VERSION_FORMATTED
           fi
 
           echo "Creating Coiled software environment for $COILED_SOFTWARE_NAME"
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -81,12 +81,15 @@ jobs:
           # Coiled software environment names can't contain "." or uppercase characters
           export PYTHON_VERSION_FORMATTED=$(echo "${{ matrix.python-version }}" | sed 's/\.//g' )
           export OS_FORMATTED=$(python -c "import os; print(os.environ['RUNNER_OS'].lower())")
+          export REF_NAME_FORMATTED=$(echo "$GITHUB_REF_NAME" | sed 's/\./-/g' )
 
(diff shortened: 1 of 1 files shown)


.github/workflows/conda.yml
```diff
@@ -78,14 +78,15 @@ jobs:
...
           export PYTHON_VERSION_FORMATTED=$(echo "${{ matrix.python-version }}" | sed 's/\.//g' )
+          export OS_FORMATTED=$(python -c "import os; print(os.environ['RUNNER_OS'].lower())")
 
           if [[ ${{ github.event_name }} = 'pull_request' ]]
           then
-            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-${{ github.event.number }}-$GITHUB_SHA-${{ matrix.os }}-py$PYTHON_VERSION_FORMATTED
+            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-${{ github.event.number }}-$GITHUB_RUN_ID-$OS_FORMATTED-py$PYTHON_VERSION_FORMATTED
           else
-            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-main-$GITHUB_SHA-${{ matrix.os }}-py$PYTHON_VERSION_FORMATTED
+            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-$GITHUB_REF_NAME-$GITHUB_RUN_ID-$OS_FORMATTED-py$PYTHON_VERSION_FORMATTED
```
ncclementi, line 89: GITHUB_REF_NAME will give us only the name of the branch or tag, If we still want a bit more info we could use GITHUB_REF instead, from the docs: "The branch or tag ref that triggered the workflow run. For branches this is the format refs/heads/<branch_name>, and for tags it is refs/tags/<tag_name>."

.github/workflows/conda.yml
```diff
@@ -78,14 +78,18 @@ jobs:
           # If testing the latest `coiled-runtime`, we need to build a Coiled software environment
           # that can be used when running tests
 
-          # Coiled software environment names can't contain "." characters
+          # Coiled software environment names can't contain "." or uppercase characters
           export PYTHON_VERSION_FORMATTED=$(echo "${{ matrix.python-version }}" | sed 's/\.//g' )
+          export OS_FORMATTED=$(python -c "import os; print(os.environ['RUNNER_OS'].lower())")
+          export REF_NAME_FORMATTED=$(echo "$GITHUB_REF_NAME" | sed 's/\./-/g' )
```
jrbourbeau, line 84: This replaces `.` with `-` in branch and tag names (e.g. `0.0.3` -> `0-0-3`) since Coiled software environment names can't contain `.`

.github/workflows/conda.yml
```diff
@@ -78,14 +78,18 @@ jobs:
...
 
+          export COILED_SOFTWARE_NAME_HEAD=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME
+          export COILED_SOFTWARE_NAME_TAIL=$GITHUB_RUN_ID-$OS_FORMATTED-py$PYTHON_VERSION_FORMATTED
```
jrbourbeau, line 87: Just pulling these out into variables so we only have to change them in one place instead both `if`/`else` branches below

.github/workflows/conda.yml
```diff
@@ -78,14 +78,18 @@ jobs:
...
           if [[ ${{ github.event_name }} = 'pull_request' ]]
           then
-            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-${{ github.event.number }}-$GITHUB_SHA-${{ matrix.os }}-py$PYTHON_VERSION_FORMATTED
+            export COILED_SOFTWARE_NAME=$COILED_SOFTWARE_NAME_HEAD-${{ github.event.number }}-$COILED_SOFTWARE_NAME_TAIL
```
jrbourbeau, line 90: Software environments from PRs now look like `coiled-runtime-pull_request-64-2185260993-linux-py39` (a bit more compact than what they are today)

.github/workflows/conda.yml
```diff
@@ -78,14 +78,18 @@ jobs:
...
           else
-            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$GITHUB_EVENT_NAME-main-$GITHUB_SHA-${{ matrix.os }}-py$PYTHON_VERSION_FORMATTED
+            export COILED_SOFTWARE_NAME=$COILED_SOFTWARE_NAME_HEAD-$GITHUB_REF_TYPE-$REF_NAME_FORMATTED-$COILED_SOFTWARE_NAME_TAIL
```
jrbourbeau, line 92: These _should_ look like `coiled-runtime-push-branch-main-2185260993-linux-py39`

.github/workflows/conda.yml
jrbourbeau: Just pushed a commit to add [`GITHUB_REF_TYPE`](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) which will be `branch` for pushing commits and `tag` when pushing a tag. I think that covers the additional information you're talking about here, but let me know if I'm missing something

ncclementi (approved):
This looks good to me, thanks @jrbourbeau for making it more clear

PR #67, opened 2022-04-18
Larger dataframe shuffle

sjperkins:
Increase dataframe size to ~80GB. Introduce:

1. Read Operations
2. Shuffle
3. Write operations

Prior to this, it would be nice to able to write the ~80GB to some temporary AWS location that gets cleaned up once the tests complete. The written data should also be cleaned up after the tests complete.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -62,7 +62,7 @@ def small_client(small_cluster):
 
 
 S3_REGION = "us-east-2"
-S3_BUCKET = "dask-io"
+S3_BUCKET = "s3://dask-io"
 
 
 @pytest.fixture(scope="session")
@@ -82,23 +82,16 @@ def s3():
 @pytest.fixture(scope="session")
 def s3_scratch(s3):
     # Ensure that the test-scratch directory exists,
-    # but do NOT reomove it as multiple test runs could be
+    # but do NOT remove it as multiple test runs could be
     # accessing it at the same time
-    stability_url = f"{S3_BUCKET}/test-scratch"
-    s3.mkdirs(stability_url, exist_ok=True)
-    return stability_url
+    scratch_url = f"{S3_BUCKET}/test-scratch"
+    s3.mkdirs(scratch_url, exist_ok=True)
+    return scratch_url
 
 
 @pytest.fixture(scope="function")
-def s3_url_factory(s3, s3_scratch):
-    urls = []
-
-    def factory(prefix="test"):
-        url = f"{s3_scratch}/{prefix}-{uuid.uuid4().hex}"
-        s3.mkdirs(url, exist_ok=False)
-        return f"s3://{url}"
-
-    yield factory
-
-    for url in urls:
-        s3.rm(url, recursive=True)
+def s3_url(s3, s3_scratch, request):
+    url = f"{s3_scratch}/{request.node.name}-{uuid.uuid4().hex}"
+    s3.mkdirs(url, exist_ok=False)
+    yield url
+    s3.rm(url, recursive=True)
--- a/tests/test_shuffle.py
+++ b/tests/test_shuffle.py
@@ -32,7 +32,4 @@ def test_shuffle_parquet(small_client, s3_url, s3_storage_options):
     shuffled_url = s3_url + "/shuffled.parquet"
     df_shuffled = dd.read_parquet(dataset_url, storage_options=s3_storage_options)
     df_shuffled = df_shuffled.shuffle(on="x")
-    out = df_shuffled.to_parquet(
-        shuffled_url, compute=False, storage_options=s3_storage_options
-    )
-    dask.compute(out)
+    df_shuffled.to_parquet(shuffled_url, storage_options=s3_storage_options)
--- a/tests/test_shuffle.py
+++ b/tests/test_shuffle.py
@@ -4,7 +4,19 @@ import pytest
 
 
 @pytest.mark.stability
-def test_shuffle_simple(small_client, s3_url_factory, s3_storage_options):
+def test_shuffle_simple(small_client):
+    df = dask.datasets.timeseries(
+        start="2000-01-01", end="2000-12-31", freq="1s", partition_freq="1D"
+    )
+
+    sdf = df.shuffle(on="x")
+
+    sdf_size = sdf.memory_usage(index=True).sum() / (1024.0**3)
+    assert sdf_size.compute() > 1.0  # 1.0 GB
+
+
+@pytest.mark.stability
+def test_shuffle_parquet(small_client, s3_url_factory, s3_storage_options):
     s3_url = s3_url_factory("shuffle-test-output")
     write_url = s3_url + "/shuffled.parquet"
     test_url = s3_url + "/dataset.parquet"
@@ -14,11 +26,11 @@ def test_shuffle_simple(small_client, s3_url_factory, s3_storage_options):
     # 100ms ~12GB
     # 75ms ~15GB
     # 50ms ~23.5GB
-    test_df = dask.datasets.timeseries(
+    df = dask.datasets.timeseries(
         start="2000-01-01", end="2000-12-31", freq="50ms", partition_freq="1D"
     )
 
-    write = test_df.to_parquet(
+    write = df.to_parquet(
         test_url,
         overwrite=True,
         compute=False,
--- a/tests/test_shuffle.py
+++ b/tests/test_shuffle.py
@@ -32,4 +32,7 @@ def test_shuffle_parquet(small_client, s3_url, s3_storage_options):
     shuffled_url = s3_url + "/shuffled.parquet"
     df_shuffled = dd.read_parquet(dataset_url, storage_options=s3_storage_options)
     df_shuffled = df_shuffled.shuffle(on="x")
-    df_shuffled.to_parquet(shuffled_url, storage_options=s3_storage_options)
+    out = df_shuffled.to_parquet(
+        shuffled_url, compute=False, storage_options=s3_storage_options
+    )
+    dask.compute(out)
--- a/tests/test_shuffle.py
+++ b/tests/test_shuffle.py
@@ -18,35 +18,20 @@ def test_shuffle_simple(small_client):
 @pytest.mark.stability
 def test_shuffle_parquet(small_client, s3_url_factory, s3_storage_options):
     s3_url = s3_url_factory("shuffle-test-output")
-    write_url = s3_url + "/shuffled.parquet"
-    test_url = s3_url + "/dataset.parquet"
-    # write_url = s3_url_factory("shuffle-test-output")
-    # test_url = s3_url_factory("shuffle-test-data")
 
-    # 100ms ~12GB
-    # 75ms ~15GB
-    # 50ms ~23.5GB
+    # Write synthetic dataset to S3
+    # Notes on how `freq` impacts total dataset size:
+    #   - 100ms ~12GB
+    #   - 75ms ~15GB
+    #   - 50ms ~23.5GB
+    dataset_url = s3_url + "/dataset.parquet"
     df = dask.datasets.timeseries(
         start="2000-01-01", end="2000-12-31", freq="50ms", partition_freq="1D"
     )
+    df.to_parquet(dataset_url, storage_options=s3_storage_options)
 
-    write = df.to_parquet(
-        test_url,
-        overwrite=True,
-        compute=False,
-        storage_options=s3_storage_options,
-    )
-
-    dask.compute(write)
-
-    # Shuffle test starts here
-    shuffle_df = dd.read_parquet(test_url, storage_options=s3_storage_options)
-    shuffle_df = shuffle_df.shuffle(on="x")
-    write = shuffle_df.to_parquet(
-        write_url,
-        overwrite=True,
-        compute=False,
-        storage_options=s3_storage_options,
-    )
-
-    dask.compute(write)
+    # Test `read_parquet` + `shuffle` + `to_parquet` works
+    shuffled_url = s3_url + "/shuffled.parquet"
+    df_shuffled = dd.read_parquet(dataset_url, storage_options=s3_storage_options)
+    df_shuffled = df_shuffled.shuffle(on="x")
+    df_shuffled.to_parquet(shuffled_url, storage_options=s3_storage_options)
--- a/tests/test_shuffle.py
+++ b/tests/test_shuffle.py
@@ -16,9 +16,7 @@ def test_shuffle_simple(small_client):
 
 
 @pytest.mark.stability
-def test_shuffle_parquet(small_client, s3_url_factory, s3_storage_options):
-    s3_url = s3_url_factory("shuffle-test-output")
-
+def test_shuffle_parquet(small_client, s3_url, s3_storage_options):
     # Write synthetic dataset to S3
     # Notes on how `freq` impacts total dataset size:
     #   - 100ms ~12GB
@@ -34,4 +32,7 @@ def test_shuffle_parquet(small_client, s3_url_factory, s3_storage_options):
     shuffled_url = s3_url + "/shuffled.parquet"
     df_shuffled = dd.read_parquet(dataset_url, storage_options=s3_storage_options)
     df_shuffled = df_shuffled.shuffle(on="x")
-    df_shuffled.to_parquet(shuffled_url, storage_options=s3_storage_options)
+    out = df_shuffled.to_parquet(
+        shuffled_url, compute=False, storage_options=s3_storage_options
+    )
+    dask.compute(out)
--- a/tests/test_shuffle.py
+++ b/tests/test_shuffle.py
@@ -32,4 +32,9 @@ def test_shuffle_parquet(small_client, s3_url, s3_storage_options):
     shuffled_url = s3_url + "/shuffled.parquet"
     df_shuffled = dd.read_parquet(dataset_url, storage_options=s3_storage_options)
     df_shuffled = df_shuffled.shuffle(on="x")
-    df_shuffled.to_parquet(shuffled_url, storage_options=s3_storage_options)
+    # FIXME: Setting `compute=True` below causes the `to_parquet` call
+    # to take much, much longer. This is unexpected and we should fix this upstream.
+    result = df_shuffled.to_parquet(
+        shuffled_url, compute=False, storage_options=s3_storage_options
+    )
+    dask.compute(result)
--- a/tests/test_shuffle.py
+++ b/tests/test_shuffle.py
@@ -5,8 +5,11 @@ import pytest
 
 @pytest.mark.stability
 def test_shuffle_simple(small_client, s3_url_factory, s3_storage_options):
-    write_url = s3_url_factory("shuffle-test-output")
-    test_url = s3_url_factory("shuffle-test-data")
+    s3_url = s3_url_factory("shuffle-test-output")
+    write_url = s3_url + "/shuffled.parquet"
+    test_url = s3_url + "/dataset.parquet"
+    # write_url = s3_url_factory("shuffle-test-output")
+    # test_url = s3_url_factory("shuffle-test-data")
 
     # 100ms ~12GB
     # 75ms ~15GB
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,10 +28,6 @@ jobs:
         os: ["ubuntu-latest", "windows-latest", "macos-latest"]
         python-version: ["3.7", "3.8", "3.9"]
         runtime-version: ["latest", "0.0.3"]
-        exclude:
-          # FIXME: Building Coiled software environments is currently broken on Windows
-          - os: windows-latest
-            runtime-version: "latest"
     
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,10 +25,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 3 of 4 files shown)


tests/test_shuffle.py
```diff
@@ -1,12 +1,59 @@
...
+        warnings.warn("Total worker memory limit > 85GB")
+
+    if sum(mem_limit) < 75 * ONE_GB:
+        warnings.warn("Total worker memory limit < 75GB")
 
-def test_shuffle_simple(small_client):
     df = dask.datasets.timeseries(
-        start="2000-01-01", end="2000-12-31", freq="1s", partition_freq="1D"
+        start="2000-01-01", end="2000-12-31", freq="50ms", partition_freq="1D"
     )
```
mrocklin, line 36: I don't think that we need a fixture for this.  I would just create `df` in the test and leave the checks out.  Dask should be stable to these operations for a variety of different worker sizes.  Also, this fixture seems to couple the dataset and the client, which I think isn't ideal (we would ideally be able to mix and match these.

tests/test_shuffle.py
```diff
@@ -1,12 +1,59 @@
...
 
-    sdf = df.shuffle(on="x")
+    yield df
+    # url = f"{BASE_URL}/test-data"
```
mrocklin, line 39: I think that the address to store things in though should *definitely* be a fixture, and we should verify that we clean it up after the `yield`

tests/test_shuffle.py
```diff
@@ -1,12 +1,59 @@
...
+    # size, _ = dask.compute(write, size)
+
+    # yield dd.read_parquet(url, storage_options=storage_options)
+
 
-    sdf_size = sdf.memory_usage(index=True).sum() / (1024.0**3)
-    assert sdf_size.compute() > 1.0  # 1.0 GB
+def test_shuffle_simple(shuffle_dataset):
+    sdf = shuffle_dataset.shuffle(on="x")
+    dask.compute(sdf)
```
mrocklin, line 52: Beware, dask.compute brings everything into memory.  I would expect this to fail.

tests/test_shuffle.py
```diff
@@ -1,12 +1,59 @@
...
+
+
+@pytest.fixture
+def shuffle_dataset(small_client):
+    # NOTE(sjperkins)
+    # I think the dataset and worker memory limits are related for this test
+    # as the dataset is large enough to provoke some spilling when it is
+    # persisted. Warn if the worker memory limits change egregiously.
+    # Alternatively, a separate cluster fixture may be appropriate
+    # for shuffle tests.
```
mrocklin, line 17: Let's just fail if things fail.  I don't think that we should be this cautious.  (it also adds more code complexity than I think we want to add to every test.)

tests/test_shuffle.py
```diff
@@ -1,12 +1,59 @@
...
+    yield df
+    # url = f"{BASE_URL}/test-data"
+    # write = df.to_parquet(
+    #     url, compute=False, overwrite=True, storage_options=storage_options
+    # )
+    # size = df.memory_usage(index=True).sum() / ONE_GB
+
+    # size, _ = dask.compute(write, size)
+
+    # yield dd.read_parquet(url, storage_options=storage_options)
```
mrocklin, line 47: Ah, it looks like you didn't actually intend this fixture to just yield a single dataframe.  I propose that you don't comment things out, and instead propose the PR that you want to go in.  It's ok if it doesn't work at first.  I think it's more important that we align on the objective and the path to get there than that CI passes initially.

tests/test_shuffle.py
sjperkins: It was succeeding in the jupyter notebook I was testing on, although there was a lot of spillage. In fact, this must be on an AWS bucket somewhere?

BTW I'm emphasising the AWS bucket because I could never get

```python
import tempfile

with tempfile.TemporaryDirectory() as tmpdirname:
  df.to_parquet(tmpdirname, ...)
```

to work (it complains it can't open one of the partitioned parquet files) and I've assumed that's because one doesn't necessarily have access to local store. However, if the worker's are spilling to disk then they must have access to some AWS file system. Any tips to unblock me here?

mrocklin: So, `dask.compute(df.to_parquet())` will be just fine because the output is small (a bunch of filenames)

`dask.compute(df)` will fail because the output is big (a lot of actual data)

mrocklin: I recommend proceding with this PR as if you had write access.  CI may still work (CI may have write access).  If it doesn't, then we'll ask other folks (like @ncclementi ) to make it so.  Let's give her something to work with though that is actually what we want to work.

sjperkins:
Just discussed this with Naty and James

> Increase dataframe size to ~80GB. Introduce:

> -  Read Operations
> -  Shuffle
> - Write operations

> Prior to this, it would be nice to able to write the ~80GB to some temporary AWS location that gets cleaned up once the tests complete. The written data should also be cleanup after the tests cleanup.

Naty is going to chat to the infra people and see if this is possible. I'll pick up the baton in the morning. I'm fine with other people changing this PR in the interim.

mrocklin:
I would do something like the following:

```python
def s3():
    with s3fs.S3FileSystem(... whatever we need to get good access ...) as s3:
        yield s3

def super_tmp_directory(s3):
    name = bucket + "/coiled-runtime-" + some_random_thing()
    s3.mkdir(name)

    yield name

    s3.rmdir(name)

def s3_directory(super_tmp_directory):
    directory = super_tmp_directory + "/" + some_random_thing()
    s3.mkdir(directory)

    yield directory
    
    s3.rmdir(directory)
```

There are probably better names that one could come up with.  If we can get an S3 bucket with a short lifetime that's good too.  I think that we probably don't need to block on that though.

sjperkins:
Sounds good.

Also I scoped the `shuffle_dataframe` fixture as `session` initially but it depends on `small_client` which is `function` scoped so that doesn't work. If anyone has immediate thoughts on how to deal with this conflict, it would be good to hear. I've dealt with this kind of stuff before dask-ms but it's been a while so I thought I'd ask. Otherwise I'll investigate in the morning

jrbourbeau:
Here's an outline for how we might use @mrocklin's fixtures in a test:

```python
import dask
import dask.dataframe as dd

def test_shuffle(small_client, s3_directory):
    # Write dataset to S3
    df = dask.datasets.timeseries(...)
    dataset_path = s3_directory + "/timseries"
    df.to_parquet(dataset_path)

    # Test `read_parquet` + `shuffle` + `to_parquet` works
    df = dd.read_parquet(dataset_path)
    df_shuffled = df.shuffle(on="x")
    df_shuffled.to_parquet(s3_directory + "/timeseries-shuffled")
```

If we write several tests that all need to read the `timeseries` dataset, then we can move the

```python
    # Write dataset to S3
    df = dask.datasets.timeseries(...)
    dataset_path = s3_directory + "/timseries"
    df.to_parquet(dataset_path)
```

part to a fixture (probably module scoped), but for now inlining directly in the test seems fine

sjperkins:
Screenshot 2022-04-19 142315

sjperkins:
OK, I think I've got to a place where I think this is meeting the requirements of https://github.com/coiled/oss-engineering/issues/94, so review would be welcomed at this point.

I know that there were a bunch of initial review comments, but could I ask people to look over the PR with a fresh eye now that its in a working form?

- @mrocklin and @jrbourbeau, I've retained the `shuffle_dataframe` fixture, mostly because I think it might be nice to run the stability tests multiple times against the same dataframe on s3 in future, but I don't have a strong opinion here.
- I've removed all the memory check warnings.
- I've also added `stability` pytest markers.

tests/test_shuffle.py
sjperkins: PR updated now that it has AWS read/write access

tests/test_shuffle.py
sjperkins: I've removed this warnings

tests/test_shuffle.py
sjperkins: My intention was to create a single dataframe on disk that could be re-used for multiple stability tests. I've left the `shuffle_dataset` fixture in to demonstrate my original idea. However, I don't have a strong opinion and could inline it into the test.

tests/test_shuffle.py
sjperkins: This is also discussed here: https://github.com/coiled/coiled-runtime/pull/67#discussion_r853091732. Shall we keep the thread in that location?

conftest.py
```diff
@@ -57,3 +58,27 @@ def small_client(small_cluster):
...
+
+
+@pytest.fixture(scope="session")
+def s3_bucket():
+    return "dask-io"
+
+
+@pytest.fixture(scope="session")
+def s3_storage_options(s3_region):
+    return {"config_kwargs": {"region_name": s3_region}}
```
ncclementi, line 75: If this is the only option we are providing, do we need one fixture to define the region and one to pass the region?

tests/test_shuffle.py
```diff
@@ -1,12 +1,61 @@
...
+@pytest.fixture(scope="session")
+def s3_stability_url(s3, s3_bucket):
+    # Unique, because multiple tests are accessing the bucket
+    stability_url = f"{s3_bucket}/stability-scratch-{uuid.uuid4().hex}"
+
+    try:
+        s3.makedirs(stability_url)
+        yield stability_url
+
+    finally:
```
ncclementi, line 17: I see that couple of the folders weren't deleted on S3 I wonder if it's from previous tries, or if there is something not entirely working. This are still in S3
Screen Shot 2022-04-19 at 11 12 30 AM

sjperkins: I think this is because a new commit was pushed up and the test run was forcibly cancelled by github actions. @jrbourbeau would you know how to handle this by any chance?

In any case, I think it's worth making all stability run data happens within a single root folder (but with unique sub-folders). I'll implement this change.

conftest.py
sjperkins: `s3_region` is used by the `s3` fixture, but also in other fixtures where it's passed in the `storage_options`.

I guess we could drop `s3_storage_options` and just use `s3_region` in the multiple places where storage options are required, but there's more possibility for something to go wrong IMO.

conftest.py
```diff
@@ -57,3 +58,27 @@ def small_client(small_cluster):
         client.wait_for_workers(10)
         client.restart()
         yield client
+
+
+@pytest.fixture(scope="session")
+def s3_region():
+    return "us-east-1"
```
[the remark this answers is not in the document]
sjperkins: I'll change it to us-east-2

tests/test_shuffle.py
```diff
@@ -1,12 +1,70 @@
...
+
+
+@pytest.fixture(scope="session")
+def s3_stability_url(s3, s3_bucket):
+    # Ensure that the stability-scratch directory exists,
+    # but do NOT reomove it as multiple test runs could be
+    # accessing it at the same time
+    stability_url = f"{s3_bucket}/stability-scratch"
+    s3.mkdirs(stability_url, exist_ok=True)
+    return stability_url
```
jrbourbeau, line 15: Nice!

Is there a reason we this top-level `stability-scratch/` directory? If not, my preference would be to have each test session create its own unique bucket and then have each test create its own bucket inside the session bucket that it uses (similar to what's outlined in https://github.com/coiled/coiled-runtime/pull/67#issuecomment-1101567089). 

These s3 fixtures will also be used by other tests outside of this module. Let's move them over to `conftest.py`, and remove any stability references, so they can be used throughout the test suite, similar to our existing client fixture.

tests/test_shuffle.py
```diff
@@ -1,12 +1,70 @@
...
+
+    try:
+        s3.makedirs(test_url)
+        yield test_url
+    finally:
+        s3.rm(test_url, recursive=True)
+
 
+@pytest.fixture
+def s3_write_url(s3, s3_stability_url):
```
jrbourbeau, line 31: Nitpick: I think we can drop `write` and still maintain the existing meaning. It would be good for the fixture tests use to say "give me an s3 bucket I can use" to have concise (but still meaningful) name

```suggestion
def s3_url(s3, s3_stability_url):
```

tests/test_shuffle.py
```diff
@@ -1,12 +1,70 @@
...
+    )
+
+
+@pytest.mark.stability
+def test_shuffle_simple(shuffle_dataset, s3_storage_options, s3_write_url):
+    sdf = shuffle_dataset.shuffle(on="x")
+    write = sdf.to_parquet(
+        f"s3://{s3_write_url}",
+        compute=False,
+        overwrite=True,
```
jrbourbeau, line 67: Are these options needed? I would have expected the default values for these parameters to work

tests/test_shuffle.py
```diff
@@ -1,12 +1,70 @@
...
+        start="2000-01-01", end="2000-12-31", freq="50ms", partition_freq="1D"
     )
 
-    sdf = df.shuffle(on="x")
+    df.to_parquet(
+        f"s3://{s3_test_data_url}",
+        compute=True,
+        overwrite=True,
+        storage_options=s3_storage_options,
+    )
```
jrbourbeau, line 54: Just curious, how long does this write take?

tests/test_shuffle.py
```diff
@@ -1,12 +1,70 @@
...
-def test_shuffle_simple(small_client):
+    try:
+        s3.makedirs(write_url)
+        yield write_url
+    finally:
+        s3.rm(write_url, recursive=True)
+
+
+@pytest.fixture
+def shuffle_dataset(small_client, s3_test_data_url, s3_storage_options):
```
jrbourbeau, line 43: This feels a bit early to me. Can we just inline this down in `test_shuffle_simple` for now? We can always factor this out into a fixture later if we need to, but I'd like to wait until there's a need before doing so

tests/test_shuffle.py
```diff
@@ -1,12 +1,70 @@
...
+    """Produces a ~80GB dataset which is about the memory limit of the cluster"""
     df = dask.datasets.timeseries(
-        start="2000-01-01", end="2000-12-31", freq="1s", partition_freq="1D"
+        start="2000-01-01", end="2000-12-31", freq="50ms", partition_freq="1D"
     )
 
-    sdf = df.shuffle(on="x")
+    df.to_parquet(
+        f"s3://{s3_test_data_url}",
```
jrbourbeau, line 50: Let's include the `s3://` prefix higher up in the s3-path generation stack (e.g. what's currently the `s3_bucket` fixture). That'll reduce the likelihood someone forgets to prepend `s3://` and since we're already using `s3_*` in fixture names, it's already clear that folks writing tests are requesting a path on S3

conftest.py
```diff
@@ -57,3 +58,27 @@ def small_client(small_cluster):
...
+
+
+@pytest.fixture(scope="session")
+def s3_region():
+    return "us-east-2"
+
+
+@pytest.fixture(scope="session")
+def s3_bucket():
+    return "dask-io"
```
jrbourbeau, line 70: Once we move the other S3 path-related fixtures here, we can probably make this a module-level `S3_BUCKET` variable too as I don't think we'll be using an feature features anymore

jrbourbeau (commented):
Thanks for working on this @sjperkins -- great to see a green build : )

tests/test_shuffle.py
sjperkins: > If not, my preference would be to have each test session create its own unique bucket

Do you mean a unique directory within the `dask-io` bucket or specifically a unique bucket `dask-test-1234abcd`? If the latter, I'm not sure if the AWS credentials have the requisite permissions and this would involve the use of boto3. 

If the former:

- Each test does have it's own unique directory, within the now renamed  top-level `test-scratch` directory
- I've created the `test-scratch` dir because Naty [pointed out](https://github.com/ncclementi) that if the GHA is cancelled lots of test data cruft can be left lying around.
- I wanted to avoid potential cruft in the bucket top-level directory structure.
- I wanted to signal that `test-scratch` can be deleted.

> These s3 fixtures will also be used by other tests outside of this module. Let's move them over to conftest.py, and remove any stability references, so they can be used throughout the test suite, similar to our existing client fixture.

I've done this and will discuss this in related comments.

tests/test_shuffle.py
sjperkins: - I've created a `s3_url_factory` fixture in `conftest.py`.
- It's a factory fixture because with the removal of the `shuffle_dataset` fixture, multiple scratch directories are necessary in `test_shuffle` to:

  1. Generate the test dataset
  2. Write the shuffled dataset

- It's not possible to use a standard fixture twice in a test

tests/test_shuffle.py
sjperkins: I've removed them and used the defaults

tests/test_shuffle.py
sjperkins: done

conftest.py
sjperkins: I've made an `S3_BUCKET` variable.

conftest.py
```diff
@@ -57,3 +58,27 @@ def small_client(small_cluster):
         client.wait_for_workers(10)
         client.restart()
         yield client
+
+
+@pytest.fixture(scope="session")
+def s3_region():
+    return "us-east-2"
```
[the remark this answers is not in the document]
sjperkins: fixed.

tests/test_shuffle.py
sjperkins: The whole test takes less than a minute, so it seems negligble at present -- so inlining seems sensible.
```bash
============================= test session starts ==============================
platform linux -- Python 3.7.12, pytest-7.1.1, pluggy-1.0.0 -- /usr/share/miniconda3/envs/test/bin/python
cachedir: .pytest_cache
rootdir: /home/runner/work/coiled-runtime/coiled-runtime, configfile: setup.cfg
plugins: anyio-3.5.0
collecting ... collected 1 item

tests/test_shuffle.py::test_shuffle_simple PASSED

=============================== warnings summary ===============================
../../../../../usr/share/miniconda3/envs/test/lib/python3.7/site-packages/coiled/_beta/cluster.py:11
  /usr/share/miniconda3/envs/test/lib/python3.7/site-packages/coiled/_beta/cluster.py:11: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    from collections import Iterable

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================== slowest durations ===============================
3060.41s call     tests/test_shuffle.py::test_shuffle_simple
109.42s setup    tests/test_shuffle.py::test_shuffle_simple
0.35s teardown tests/test_shuffle.py::test_shuffle_simple
================== 1 passed, 1 warning in 3170.43s (0:52:50) ===================
```

tests/test_shuffle.py
```diff
@@ -1,12 +1,31 @@
 import dask
+import dask.dataframe as dd
+import pytest
 
 
-def test_shuffle_simple(small_client):
-    df = dask.datasets.timeseries(
-        start="2000-01-01", end="2000-12-31", freq="1s", partition_freq="1D"
+@pytest.mark.stability
+def test_shuffle_simple(small_client, s3_url_factory, s3_storage_options):
```
fjetter, line 7: I would actually consider having both tests around is valuable. The original test doesn't involve IO, nor parquet.

How aobut

test_shuffle_simple
and
test_shuffle_parquet

tests/test_shuffle.py
ncclementi: @sjperkins I think you might have missed the comment above, do you have an estimate?

tests/test_shuffle.py
sjperkins: So I found something interesting. when I omit `compute=True` and `overwrite=False` there's a lot more spilling to disk and the tests can take up to an hour!

I'm wondering if `compute=True` results in a persist somehow in the distributed case somehow.

sjperkins:
@jrbourbeau I've addressed most of your comments. I ran into something strange: when I omit the `compute=False` and `overwrite=True` keywords from `to_parquet` the cluster seems to not release the data written to disk. Perhaps it persists in the distributed case? As of https://github.com/coiled/coiled-runtime/pull/67/commits/d28daf1fb4b3aa5856bc4cfa12d1528ad5acbcc3, the tests take about 5m30s to complete.

jrbourbeau:
I ran CI a couple of times both with `compute=True` and `compute=False` and consistently got the behavior you ran into (i.e. `compute=False` + `dask.compute(…)` is much faster). This is definitely unexpected. I think we should:

1. Make sure the cluster that’s being spun up in both cases is the same (I suspect it is, but we should confirm as that’s the only thing I can think that could lead to an expected difference in performance in this situation)
2. If the cluster resources are the same, file an upstream issue in `dask/dask`
3. Set `compute=False` in this test and move onto other stability tests. It’ll be good to circle back to `test_shuffle_parquet` after we figure out why there’s a performance difference, but it’s should block merging this PR

tests/test_shuffle.py
sjperkins: Apologies, I was holding back because of the run time discrepancies James and I were seeing with `compute={True,False}` on `to_parquet`.

When the test is working as expected it takes ~5m30s. I estimate from watching the cluster that it takes roughly 30s to write the testing dataset.

jrbourbeau:
Updating to the latest `dask` + `distributed` release didn't fix the performance issue. I updated this PR to use `compute=False` with a comment about wanting to figure out what's going on and eventually move to `compute=True` and opened https://github.com/coiled/coiled-runtime/issues/79 to track the performance issue

sjperkins:
Thanks for pushing this over finishing line @jrbourbeau!

PR #69, opened 2022-04-18
Lint conda recipe

jrbourbeau:
This PR adds a new conda recipe linting check (the same one that's run over in conda-forge)

EDIT: I included a temporary commit which should fail the linting check to confirm that the recipe linting in CI is working as expected


jrbourbeau:
Getting 

```
recipe has some lint:
  ``requirements: run: numba==0.55.1`` must contain a space between the name and the pin, i.e. ``numba ==0.55.1``
```

in the recipe linting job, so things are working as expected

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,39 @@
+name: Linting
```
ncclementi, line 1: Since we are modifying the name, can we make all the extensions match? The other workflow has `yml`

jrbourbeau: Np, done

ncclementi:
This looks great, going in!

PR #70, opened 2022-04-18
GHA workflow for managing Coiled software environments

### LINKED ISSUE #32: Create GHA workflow for managing Coiled software environments
As we start testing `coiled-runtime` on Coiled clusters, we'll need to have Coiled software environments with various versions of `coiled-runtime` installed. We should create a GHA workflow which manages these software environments. This workflow should be separate from our existing workflows that run on PRs to avoid software environments being unintentionally changed when someone is working on a PR

ncclementi:
- [x] Closes #32 

The action is set up to only be run on push to main, so we can't test it actually on this PR, I wonder if we should slightly change the name of the soft env, to not override the existing ones and run it on this PR.?

@jrbourbeau any suggestions?


.github/workflows/software-environments.yml
```diff
@@ -0,0 +1,42 @@
...
+on:
+  push:
+    branches: main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 4
```
jrbourbeau, line 13: Curious why this is needed

.github/workflows/software-environments.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Create Runtime ${{ matrix.runtime-version }} Python ${{ matrix.python-version }} software environment
+        env:
+          DASK_COILED__SERVER: ${{ matrix.server }}
+          DASK_COILED__TOKEN: ${{ secrets[matrix.token-name] }}
+        run: |
+          # Coiled software environment names can't contain "." or uppercase characters
+          export RUNTIME_VERSION_FORMATTED=$(echo "${{ matrix.runtime-version }}" | sed 's/\./-/g' )
+          export PYTHON_VERSION_FORMATTED=$(echo "${{ matrix.python-version }}" | sed 's/\.//g' )
+          export COILED_SOFTWARE_ENV_NAME=coiled-runtime-$RUNTIME_VERSION_FORMATTED-py$PYTHON_VERSION_FORMATTED
+          export COILED_SOFTWARE_ENV_PATH=environments/coiled-runtime-${{ matrix.runtime-version }}-py$PYTHON_VERSION_FORMATTED.yml
```
jrbourbeau, line 41: What do you think about unifying the environment name (which needs to use `-`) and the YAML file name? I.e. update paths like:

```
environments/coiled-runtime-0.0.3-py37.yml
```

to 

```
environments/coiled-runtime-0-0-3-py37.yml
```
That would help increase self-consistency a bit

jrbourbeau: We would also need one less `COILED_SOFTARE_ENV_*` variable here

.github/workflows/software-environments.yml
```diff
@@ -0,0 +1,43 @@
+name: Create software environments
+
+on:
+  push:
+    branches: main
```
jrbourbeau, line 5: Since we won't need to run this workflow very often (normally only when we issue a new `coiled-runtime` release), and the build matrix below creates several CI jobs, we could use [GHAs `path` filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) to only run this workflow when a file in `environments/` is added / removed / modified. It should look something like

```yaml
on:
  push:
    branches: main
    paths:
      - environments/**
```

jrbourbeau (commented):
Thanks @ncclementi

.github/workflows/software-environments.yml
ncclementi: I like this.

.github/workflows/software-environments.yml
ncclementi: You just read my mind. I was about to change this to make it easier.

.github/workflows/software-environments.yml
ncclementi: I copied it over from https://github.com/coiled/software-environments/blob/main/.github/workflows/ci-update-envs.yml Should we remove those 4 lines? or just the strategy part?

jrbourbeau: Just the `max-parallel: 4` part. I don't think we need it here. I see you removed that line -- thanks

jrbourbeau (approved):
Nice, thanks for handling this @ncclementi. This looks good to go

We spoke about this offline, but just noting for others that a follow-up to this PR will be to start using these new `coiled-runtime-*` software environments in CI by updating our `software` fixture here

https://github.com/coiled/coiled-runtime/blob/10d53ae77d7c0c24e4972e333ead6004f1cb3e4f/conftest.py#L31-L36

PR #72, opened 2022-04-19
[WIP] Test fix for windows ci

### LINKED ISSUE #49: Fix Windows CI issue when testing latest `coiled-runtime`
In https://github.com/coiled/coiled-runtime/pull/33 we explicitly excluded running Windows builds against the latest `coiled-runtime` in CI

https://github.com/coiled/coiled-runtime/blob/95ba3528faefc6302274ecc3f4155e836c3ba54b/.github/workflows/conda.yml#L29-L32

due to the issue described in https://github.com/coiled/coiled-runtime/pull/33#discussion_r849753447. Opening this issue so we remember to go back and fix that and re-introduce Windows builds for the latest `coiled-runtime`

ncclementi:
- [ ] Closes #49

Running only on windows, for now, to see if we can fix the problem. I will revert changes when done.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,7 +25,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest", "windows-latest", "macos-latest"]
+        # ["ubuntu-latest", "windows-latest", "macos-latest"]
+        #test windows for now
+        os: ["windows-latest"]
         python-version: ["3.7", "3.8", "3.9"]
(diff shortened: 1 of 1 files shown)


ncclementi:
This "runs" properly with pip, but I don't see the software env created on the coiled dashboard. I also have commented out the clean-up part so I would expect to see the `dask-engineering/coiled-runtime-pull_request-72-2192292739-windows-py37` but it's not there. Any idea what could it be?

This will only be needed if fix over the feedstock doesn't work

ncclementi:
Closing in favor of #74

PR #73, opened 2022-04-19
Switch to coiled-runtime default envs forn non-latest runs

ncclementi:
Follow up from https://github.com/coiled/coiled-runtime/pull/70 tackling https://github.com/coiled/coiled-runtime/pull/70#pullrequestreview-944881051
 
Trying to use the environments generated with the `software-environments.yml` GHA

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,8 @@ jobs:
       matrix:
         os: ["ubuntu-latest", "windows-latest", "macos-latest"]
         python-version: ["3.7", "3.8", "3.9"]
-        runtime-version: ["latest", "0.0.3"]
+        # ["latest", "0.0.3"]
+        runtime-version: ["0.0.3"]
         exclude:
           # FIXME: Building Coiled software environments is currently broken on Windows
           - os: windows-latest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -111,7 +111,7 @@ jobs:
             export EXTRA_OPTIONS=" "
           fi
 
-          python -m pytest $EXTRA_OPTIONS tests
+          python -m pytest $EXTRA_OPTIONS tests/test_coiled.py
 
       - name: Remove Coiled software environment
         if: ${{ matrix.runtime-version == 'latest' }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -109,9 +109,6 @@ jobs:
             export COILED_SOFTWARE_NAME=${{ env.COILED_SOFTWARE_NAME }}
           else
             export EXTRA_OPTIONS=" "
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -109,6 +109,9 @@ jobs:
             export COILED_SOFTWARE_NAME=${{ env.COILED_SOFTWARE_NAME }}
           else
             export EXTRA_OPTIONS=" "
+            export RUNTIME_VERSION_FORMATTED=$(echo "${{ matrix.runtime-version }}" | sed 's/\./-/g' )
+            export PYTHON_VERSION_FORMATTED=$(echo "${{ matrix.python-version }}" | sed 's/\.//g' )
+            export DEFAULT_SOFTWARE_NAME=dask-engineering/coiled-runtime-$RUNTIME_VERSION_FORMATTED-py$PYTHON_VERSION_FORMATTED
```
ncclementi, line 114: We could just use `COILED_SOFTWARE_NAME` I was wondering if we wanted to be more explicit about the env name, But happy to change it to with the respective change on the conf test
```suggestion
            export COILED_SOFTWARE_NAME=dask-engineering/coiled-runtime-$RUNTIME_VERSION_FORMATTED-py$PYTHON_VERSION_FORMATTED
```

conftest.py
```diff
@@ -31,8 +30,7 @@ def pytest_collection_modifyitems(config, items):
 @pytest.fixture(scope="session")
 def software():
     return os.environ.get(
-        "COILED_SOFTWARE_NAME",
-        f"dask-engineering/coiled_dist-py{sys.version_info[0]}{sys.version_info[1]}",
+        "COILED_SOFTWARE_NAME", os.environ.get("DEFAULT_SOFTWARE_NAME")
```
ncclementi, line 33:
```suggestion
        "COILED_SOFTWARE_NAME"
```

jrbourbeau: This works fine in CI, but means that folks will need to set an environment variable locally to run tests, which is easy to forget. I'd prefer for this to "just work" (or fail with an informative error message) these environment variables aren't set. 

For now, let's just strictly swap out the `dask-engineering/coiled_dist-py*` environments here with the newer `dask-engineering/coiled-runtime*` environments. This is already an improvement as we'll be using software environments which are automatically build for us (instead of adhoc like before). We should still think more about how we want to fix this longer-term, but let's do that separate from this PR

ncclementi:
@jrbourbeau and I chatted about not using an env variable and parsing the version of the coiled-runtime for the soft-envs I tried what we discussed but something is not quite working for the case of latest.

jrbourbeau (approved):
Thanks @ncclementi -- I think if CI passes, then this is good to merge

ncclementi:
Sounds good to me. Thank you @jrbourbeau!

PR #78, opened 2022-04-21
Temporarily use conda to export env for debug statement

ncclementi:
- [x] Closes https://github.com/coiled/coiled-runtime/issues/77


jrbourbeau (approved):
Thanks for the quick workaround! Not sure why this isn't working for `mamba`, but this seems fine. The CI failure seems coiled-related -- could you open a GitLab issue with that traceback?

PR #81, opened 2022-04-22
Drop `coiled` conda channel

jrbourbeau:
Since `coiled-runtime` is on `conda-forge` now, we no longer need to specify `-c coiled`

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import os
 import shlex
 import subprocess
@@ -7,13 +8,16 @@ import uuid
 
 import pytest
 import s3fs
+from dask.distributed import Client
 
 try:
     from coiled.v2 import Cluster
 except ImportError:
     from coiled._beta import ClusterBeta as Cluster
 
-from dask.distributed import Client
+
+# So coiled logs can be displayed on test failure
+logging.getLogger("coiled").setLevel(logging.INFO)
 
 
 def pytest_addoption(parser):
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,12 +25,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest"]
-        python-version: ["3.9"]
-        runtime-version: ["0.0.3"]
-        # os: ["ubuntu-latest", "windows-latest", "macos-latest"]
-        # python-version: ["3.7", "3.8", "3.9"]
-        # runtime-version: ["latest", "0.0.3"]
+        os: ["ubuntu-latest", "windows-latest", "macos-latest"]
+        python-version: ["3.7", "3.8", "3.9"]
+        runtime-version: ["latest", "0.0.3"]
     
     steps:
       - uses: actions/checkout@v2
@@ -113,7 +110,7 @@ jobs:
             export EXTRA_OPTIONS=" "
           fi
 
-          python -m pytest $EXTRA_OPTIONS tests/test_coiled.py::test_quickstart
+          python -m pytest $EXTRA_OPTIONS tests
 
       - name: Remove Coiled software environment
         if: ${{ matrix.runtime-version == 'latest' }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,9 +25,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest", "windows-latest", "macos-latest"]
-        python-version: ["3.7", "3.8", "3.9"]
-        runtime-version: ["latest", "0.0.3"]
+        os: ["ubuntu-latest"]
+        python-version: ["3.9"]
+        runtime-version: ["0.0.3"]
+        # os: ["ubuntu-latest", "windows-latest", "macos-latest"]
+        # python-version: ["3.7", "3.8", "3.9"]
+        # runtime-version: ["latest", "0.0.3"]
     
     steps:
       - uses: actions/checkout@v2
@@ -110,7 +113,7 @@ jobs:
             export EXTRA_OPTIONS=" "
           fi
 
-          python -m pytest $EXTRA_OPTIONS tests
+          python -m pytest $EXTRA_OPTIONS tests/test_coiled.py::test_quickstart
 
       - name: Remove Coiled software environment
         if: ${{ matrix.runtime-version == 'latest' }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,9 +52,9 @@ jobs:
           if [[ ${{ matrix.runtime-version }} = 'latest' ]]
           then
             python ci/create_latest_runtime_meta.py
-            mamba install -c coiled -c conda-forge --file latest.txt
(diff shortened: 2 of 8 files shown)


ncclementi:
@jrbourbeau This looks good to me, the only question I have is if this PR should include removing coiled as a channel from the `environment.yml` files we use to build the software envs.

jrbourbeau:
Ah, I missed those -- good catch

ncclementi:
Is the CI failure OSError a known error on distributed or is this something new?
see https://github.com/coiled/coiled-runtime/runs/6131631876?check_suite_focus=true#step:8:1884

ntabris:
That's a generic timeout when trying to connect to scheduler. Maybe the cluster died? I don't see where in your test logs there's any info about the cluster (like cluster name, link to details page, that sort of thing) that would make it easier to troubleshoot further.

ncclementi:
@jrbourbeau Nat makes a good point, I can't find any info either, do we print any of this? Maybe we should print out some cluster info, like the name at least? What do you think

ntabris:
Here's an example of what we have in coiled integration tests:

```
DEBUG:coiled.context:Entering coiled.core.Cloud.get_account_provider_name-q6opt9op
DEBUG:coiled.context:Exiting coiled.core.Cloud.get_account_provider_name-q6opt9op - DD URL: https://app.datadoghq.com/apm/traces?query=%40coiled-operation-id%3Aq6opt9op&paused=true&streamTraces=false&showAllSpans=true&start=1650651418851&end=1650651539095
DEBUG:coiled.context:Entering coiled._beta.cluster.ClusterBeta._start-h4r79qbk
INFO:coiled:Creating Cluster (name: cluster-zespm, https://cloud.coiled.io/gcp-vm-testbot-account/clusters/12078/details). This might take a few minutes...
ERROR:coiled:   | Worker Process         | cluster-zespm-worker-41d57f048f                | error      at 18:28:34 (UTC) | Process never phoned home
```

From the cluster name that was in the logs (`test_h2o_benchmarks-f40a48c9`) I was able to find the cluster pk id, which we can then use to get to the details page:

https://cloud.coiled.io/dask-engineering/clusters/11993/details

Looks like those instances self-terminated, don't currently have an idea why.

shughes-uk:
Ensuring the `coiled` logger is configured at the `INFO` level should capture enough details to help debug things (primarily a link to the cluster details page)

ntabris:
Hm, actually it just looks like your cluster had an idle timeout at 20 minutes.


```
Apr 22 16:42:03 ip-10-0-4-24 cloud-init[1018]: distributed.scheduler - INFO - Scheduler closing after being idle for 1200.00 s
--
Apr 22 16:42:03 ip-10-0-4-24 cloud-init[1018]: distributed.scheduler - INFO - Scheduler closing...
Apr 22 16:42:03 ip-10-0-4-24 cloud-init[1018]: distributed.scheduler - INFO - Connecting to coiled - I am i-0ae4ec40af7c8a95c
Apr 22 16:42:03 ip-10-0-4-24 cloud-init[1018]: distributed.scheduler - INFO - my auth token ends with KzIA
Apr 22 16:42:03 ip-10-0-4-24 cloud-init[1018]: distributed.scheduler - INFO - Scheduler closing all comms
Apr 22 16:42:03 ip-10-0-4-24 cloud-init[1018]: distributed.scheduler - INFO - Remove worker <WorkerState 'tls://10.0.0.236:46525', name: test_h2o_benchmarks-f40a48c9-worker-870e1f88f4, status: running, memory: 0, processing: 0>
Apr 22 16:42:03 ip-10-0-4-24 cloud-init[1018]: distributed.core - INFO - Removing comms to tls://10.0.0.236:46525
```

ncclementi:
Are there any coiled docs on how to get the links  you would find useful so I can add a step on the GHA to have them handy? 
Something like `echo coiled.cluster.logs_link` or something like that?

ntabris:
> Are there any coiled docs on how to get the links you would find useful so I can add a step on the GHA to have them handy?
> Something like `echo coiled.cluster.logs_link` or something like that?

You'll get cluster details link if you enable INFO level logging (like @shughes-uk said), and from the cluster details page there are links to download instance logs.

ncclementi:
> Hm, actually it just looks like your cluster had an idle timeout at 20 minutes.

That's the default and we haven't had a problem before. I wonder if it's because we did not get the 10 workers we requested. We only got 8, and we do have a  wait for workers' call. 

https://github.com/coiled/coiled-runtime/blob/304ae9db862e23d38f17d73ce7a3f7ca965eeff2/conftest.py#L75

shughes-uk:
the `client.wait_for_workers` is definitely not ideal for coiled. It has no information about coiled state, so using it will often lead to a forever timeout instead of exiting early if the number of workers you're waiting for will never happen (ie because one crashed). We're looking at implementing `wait_for_workers` on the cluster level (or having dask hooks for the client.wait_for_workers). 

In the meantime my extremely hacky workaround is

```python
    td_seconds = timeout.total_seconds()
    cluster._wait_for_workers_arg = n_workers
    cluster.timeout = td_seconds
    await cluster._wait_until_ready(is_new_cluster=True)
    await client.wait_for_workers(n_workers=n_workers)
```
This uses private methods we might change so no guarantee of stability but I think we'd only change them to implement a proper `wait_for_workers`

shughes-uk:
https://github.com/dask/distributed/pull/6041 The request for hooks is here if you can help us help you!

conftest.py
```diff
@@ -7,13 +8,16 @@
...
 
 try:
     from coiled.v2 import Cluster
 except ImportError:
     from coiled._beta import ClusterBeta as Cluster
 
-from dask.distributed import Client
+
+# So coiled logs can be displayed on test failure
+logging.getLogger("coiled").setLevel(logging.INFO)
```
jrbourbeau, line 20: This worked locally for me. @shughes-uk @ntabris suggestions are welcome if there's a more recommended way of doing this

ncclementi: So we have a new failure but I can't seem to find the logs on CI. 
https://github.com/coiled/coiled-runtime/runs/6135051159?check_suite_focus=true

ntabris: Looks like it didn't even get as far as connecting to cloud.coiled.io so there wouldn't be logs about creating cluster. Did DNS break in your CI or something? You might just try running again and see if anything changes.

```
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host cloud.coiled.io:443 ssl:default [nodename nor servname provided, or not known]
```

shughes-uk: For pytest you'll ned to configure a logging handler that outputs to stdout 

https://realpython.com/python-logging/#basic-configurations this is a nice fast way to add a logger for that

shughes-uk: You can also run pytest with ```pytest --log-cli-level INFO` , but be warned this will set ALL loggers to `INFO` level

jrbourbeau:
This PR seems good to go to me. We've dropped outdated references to the `coiled` conda channel and are now outputting Coiled (info level) logs when a test doesn't pass (getting the cluster details link is quite nice). Planning to merge in a bit if no further comments

ncclementi:
This goes in, thank you @jrbourbeau :tada:

PR #84, opened 2022-04-22
Add repeated merge spill deadlock test

jrbourbeau:
xref https://github.com/coiled/coiled-runtime/issues/83


jrbourbeau:
Alright, we've got a deadlock (yey?). So now we can ensure that it doesn't reemerge once a `distributed` release with a fix if out

tests/test_deadlock.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_deadlock(software):
+    # See https://github.com/dask/distributed/issues/6110
+
+    with coiled.v2.Cluster(
+        name=f"test_deadlock-{uuid.uuid4().hex}",
+        n_workers=20,
+        account="dask-engineering",
+        software=software,
```
mrocklin, line 15: Can we throw these into `dask.config` with a fixture?  It would be good for folks to not have to think about these keywords.

tests/test_deadlock.py
```diff
@@ -0,0 +1,51 @@
+import uuid
+
+import coiled.v2
+import dask
+from distributed import Client
+
+
+def test_deadlock(software):
```
mrocklin, line 8: I'd rather we aimed towards correctness, and called this "test shuffle" or something similar.  When things are working this should test that a certain user-focused workload is solid.  I think that that should influence the name.

tests/test_deadlock.py
```diff
@@ -0,0 +1,51 @@
...
+            while True:
+                print(f"Iteration {i}")
+                client.restart()
+                demo1 = ddf.map_partitions(slowident)
+                (demo1.x + demo1.y).mean().compute()
+
+                demo2 = ddf.merge(ddf2)
+                demo2 = demo2.map_partitions(slowident)
+                (demo2.x + demo2.y).mean().compute()
+                i += 1
```
mrocklin, line 51: Before we encode this into a test we should verify that we need all of this.  I don't think that this was intended to be minimal.  For example, I wouldn't be surprised if we didn't need the sleep in here.

tests/test_deadlock.py
jrbourbeau: Sure, this was just a temporary filler name

tests/test_deadlock.py
jrbourbeau: +1

tests/test_deadlock.py
jrbourbeau: Yeah, agreed we want to remove portions that aren't needed. Right now I'm just verifying we can trigger the deadlock in CI. Still need to clean things up

jrbourbeau:
cc @sjperkins for visibility

sjperkins:
@sjperkins to:

- Improve the assert on the test
- Checking the state on scheduler/worker
- A good place to start:

  - The transition counter
  - is `WorkerState.last_seen`
  - /cc @gjoseph92 if he has further ideas on things to try.

conftest.py
```diff
@@ -57,13 +58,22 @@ def software():
...
+            "coiled.account": "dask-engineering",
+            "coiled.software": software,
+        }
+    ):
+        yield
+
+
 @pytest.fixture(scope="module")
-def small_cluster(software, request):
+def small_cluster(request):
```
jrbourbeau, line 73: Moving `dask.config.set(...)` into an `autouse` fixture is causing us to no longer use `account="dask-engineering"` in this fixture and are running into errors due to us hitting Coiled CPU limits 

```suggestion
def small_cluster(dask_config, request):
```

tests/test_deadlock.py
```diff
@@ -0,0 +1,39 @@
+import uuid
+
+import coiled.v2
+import dask
+from distributed import Client, wait
+
+
+def test_repeated_merge_spill():
```
jrbourbeau, line 8: We should add something like `@pytest.mark.skip(reason="https://github.com/dask/distributed/issues/6110")` given that this test currently fails

conftest.py
sjperkins: Thanks, I'll change this

tests/test_deadlock.py
sjperkins: I think it should be `pytest.mark.xfail` instead?

sjperkins: Should we not be strictly reporting failure here? It's a failed stability test.

jrbourbeau: `xfail` would also work here. The main thing we want is for this to not cause CI builds to fail because this is a known stability issue we're working to solve. I proposed `skip` to avoid an extra 4 minutes on each CI build (`xfail` will still run the test)

jrbourbeau:
Whoops, looks like my suggestion ran into some fixture-scoping mismatch error. Thank you for fixing @sjperkins

PR #88, opened 2022-04-25
Update release procedure

jrbourbeau:
This PR updates our release procedure to reflect that we're now issuing releases on conda-forge (instead of the `coiled` conda channel) as well as adds a few more details. 

cc @ncclementi @fjetter @hayesgb

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -125,28 +125,6 @@ jobs:
             python -c "raise Exception('Tests failed. Please see the output from the previous step for more details.')"
           fi
 
-      - name: Build coiled-runtime
-        if: ${{ matrix.runtime-version == 'latest' }}
-        run: |
-          conda mambabuild recipe \
-                --output-folder dist/conda \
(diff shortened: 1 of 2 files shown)


ncclementi:
Does this mean we are dropping the upload to the anaconda channel completely? For some reason, I thought we were maintaining both, but I'm happy to only keep conda-forge if that's the plan.

If that's the case, I think this PR looks good and ready to go.

jrbourbeau:
> Does this mean we are dropping the upload to the anaconda channel completely?

At the moment we're just using `conda-forge` for releases, which is why I dropped those GHA steps here. However, as part of https://github.com/coiled/coiled-runtime/issues/86, we'll start pushing nightly releases of `coiled-runtime` to the `coiled` conda channel. So eventually we'll be using both channels, but for now I think it's safe to remove these steps.

ncclementi:
SOunds good, then. This goes in. Thanks for updating this @jrbourbeau

PR #90, opened 2022-04-25
Test against development version of `dask` and `distributed`

jrbourbeau:
This PR makes it so when a commit message on a PR contains the phrase `test-upstream`, we use the current `main` branch version of `dask` and `distributed` in the `latest` CI build 

Closes https://github.com/coiled/coiled-runtime/issues/89

### DIFF
--- a/ci/create_latest_runtime_meta.py
+++ b/ci/create_latest_runtime_meta.py
@@ -4,7 +4,7 @@ import json
 import os
 import pathlib
 import sys
-from urllib.request import urlopen
+from urllib.request import Request, urlopen
 
 import yaml
 from jinja2 import Environment, FileSystemLoader, select_autoescape
@@ -12,9 +12,13 @@ from jinja2 import Environment, FileSystemLoader, select_autoescape
 
 def get_latest_commit(repository):
     org, repo = repository.split("/")
-    response = urlopen(
-        f"https://api.github.com/repos/{org}/{repo}/commits?per_page=1"
-    ).read()
+    headers = {}
+    if os.environ.get("GITHUB_TOKEN", False):
+        headers = {"Authorization": f"token {os.environ['GITHUB_TOKEN']}"}
+    request = Request(
+        f"https://api.github.com/repos/{org}/{repo}/commits?per_page=1", headers=headers
+    )
+    response = urlopen(request).read()
     data = json.loads(response.decode())
     return data[0]["sha"]
 
--- a/ci/create_latest_runtime_meta.py
+++ b/ci/create_latest_runtime_meta.py
@@ -12,8 +12,9 @@ from jinja2 import Environment, FileSystemLoader, select_autoescape
 
 def get_latest_commit(repository):
     org, repo = repository.split("/")
-    url = f"https://api.github.com/repos/{org}/{repo}/commits?per_page=1"
-    response = urlopen(url).read()
+    response = urlopen(
+        f"https://api.github.com/repos/{org}/{repo}/commits?per_page=1"
+    ).read()
     data = json.loads(response.decode())
     return data[0]["sha"]
 
@@ -31,7 +32,6 @@ def main():
     # Ensure Python is pinned to X.Y.Z version currently being used
     python_version = ".".join(map(str, tuple(sys.version_info)[:3]))
     assert sum([req.startswith("python ") for req in requirements]) == 1
-
     for idx, req in enumerate(requirements):
         if req.startswith("python "):
             requirements[idx] = f"python =={python_version}"
--- a/ci/create_latest_runtime_meta.py
+++ b/ci/create_latest_runtime_meta.py
@@ -1,11 +1,21 @@
 from __future__ import annotations
 
+import json
 import os
 import pathlib
 import sys
+from urllib.request import urlopen
 
 import yaml
 from jinja2 import Environment, FileSystemLoader, select_autoescape
+
+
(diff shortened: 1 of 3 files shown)


jrbourbeau:
Note, it's expected for all the upstream Python 3.7 builds to fail with `ERROR: Package 'dask' requires a different Python: 3.7.12 not in '>=3.8'` as `dask` and `distributed` have dropped support for 3.7

ci/create_latest_runtime_meta.py
```diff
@@ -24,9 +40,15 @@ def main():
...
-    # File compatible with `mamba install --file <...>`
-    with open("latest.txt", "w") as f:
-        f.write("\n".join(requirements))
+    # Optionally install the development version of `dask` and `distributed`
+    if os.environ.get("TEST_UPSTREAM", False):
+        dev_req = {
+            "pip": [
+                f"git+https://github.com/dask/dask.git@{get_latest_commit('dask/dask')}",
+                f"git+https://github.com/dask/distributed.git@{get_latest_commit('dask/distributed')}",
+            ]
```
jrbourbeau, line 49: It would be good to eventually move from `pip` here to use directly install with conda from the https://anaconda.org/dask/ channel. I tried this locally but ran into https://github.com/dask/dask/issues/8984. 

I think this is good enough for now, and we can open a ticket to switch to pure-conda once https://github.com/dask/dask/issues/8984 is resolved.

jrbourbeau (commented):
@ncclementi would you mind taking a look at this when you get a chance?

ci/create_latest_runtime_meta.py
jrbourbeau: For reference, something like 

```diff
diff --git a/ci/create_latest_runtime_meta.py b/ci/create_latest_runtime_meta.py
index 908f0e8..780ba13 100644
--- a/ci/create_latest_runtime_meta.py
+++ b/ci/create_latest_runtime_meta.py
@@ -3,24 +3,20 @@ from __future__ import annotations
 import json
 import os
 import pathlib
+import shlex
+import subprocess
 import sys
-from urllib.request import Request, urlopen
 
 import yaml
 from jinja2 import Environment, FileSystemLoader, select_autoescape
+from packaging.requirements import Requirement
 
 
-def get_latest_commit(repository):
-    org, repo = repository.split("/")
-    headers = {}
-    if os.environ.get("GITHUB_TOKEN", False):
-        headers = {"Authorization": f"token {os.environ['GITHUB_TOKEN']}"}
-    request = Request(
-        f"https://api.github.com/repos/{org}/{repo}/commits?per_page=1", headers=headers
-    )
-    response = urlopen(request).read()
-    data = json.loads(response.decode())
-    return data[0]["sha"]
+def get_latest_conda(package):
+    cmd = f"conda search --json --override-channels -c dask/label/dev {package}"
+    out = subprocess.check_output(shlex.split(cmd), text=True)
+    result = json.loads(out)[package][-1]
+    return f"{result['version']}={result['build']}"
 
 
 def main():
@@ -35,23 +31,26 @@ def main():
 
     # Ensure Python is pinned to X.Y.Z version currently being used
     python_version = ".".join(map(str, tuple(sys.version_info)[:3]))
-    assert sum([req.startswith("python ") for req in requirements]) == 1
     for idx, req in enumerate(requirements):
-        if req.startswith("python "):
+        package_name = Requirement(req).name
+        if package_name == "python":
             requirements[idx] = f"python =={python_version}"
 
+    # File compatible with `mamba env create --file <...>`
+    env = {"channels": ["conda-forge"]}
+
     # Optionally install the development version of `dask` and `distributed`
+    # from `dask/label/dev` conda channel
     if os.environ.get("TEST_UPSTREAM", False):
-        dev_req = {
-            "pip": [
-                f"git+https://github.com/dask/dask.git@{get_latest_commit('dask/dask')}",
-                f"git+https://github.com/dask/distributed.git@{get_latest_commit('dask/distributed')}",
-            ]
-        }
-        requirements.append(dev_req)
+        env["channels"].append("dask/label/dev")
+
+        upstream_packages = {"dask"}
+        # upstream_packages = {"dask", "distributed"}
+        for idx, req in enumerate(requirements):
+            package_name = Requirement(req).name
+            if package_name in upstream_packages:
+                requirements[idx] = f"{package_name} =={get_latest_conda(package_name)}"
 
-    # File compatible with `mamba env create --file <...>`
-    env = {"channels": ["conda-forge"]}
     env["dependencies"] = requirements
     with open("latest.yaml", "w") as f:
         yaml.dump(env, f)
````

is what we want to switch over to conda-only

ncclementi:
@jrbourbeau this looks good to me, I just curious what change since 

> Note, it's expected for all the upstream Python 3.7 builds to fail with `ERROR: Package 'dask' requires a different Python: 3.7.12 not in '>=3.8'` as `dask` and `distributed` have dropped support for 3.7

that now, py3.7  latest are passing? this is more of a question to understand, outside that I think this is good to go.

jrbourbeau:
Sorry I should have been more clear. I expect to see the Python 3.7 builds fail for commits which include `test-upstream` in their commit because that's the condition in which we'll install `dask` and `distributed` `main` branches (which don't support Python 3.7). The latest commit I pushed didn't include `test-upstream` in its commit because I noticed that all the commits I've been pushing to this PR contain `test-upstream` and I wanted to check that things behave normally when we're not using the `upstream` build

PR #91, opened 2022-04-26
Split tests into separate GHA jobs

sjperkins:


jrbourbeau (commented):
Thanks @sjperkins. It looks like we're moving these to separate steps, but still sequential steps in the same GHA job. My understanding is if we want to have a separate build matrix for each of these subsets of tests, then we'll need to move each to a separate job (though I could be missing something)

.github/workflows/tests.yml
```diff
@@ -110,7 +109,47 @@ jobs:
...
+          # Ensure we run additional tests when testing the latest coiled-runtime
+          if [[ ${{ matrix.runtime-version }} = 'latest' ]]
+          then
+            export EXTRA_OPTIONS="--run-latest"
+            export COILED_SOFTWARE_NAME=${{ env.COILED_SOFTWARE_NAME }}
+          else
+            export EXTRA_OPTIONS=" "
+          fi
+
+          python -m pytest $EXTRA_OPTIONS tests/h2o
```
jrbourbeau, line 131: Today there are only h2o benchmarks, but we'll soon be adding more (e.g. some parquet-related benchmarks). Given this, can make the directory structure / names use the more generic term "benchmarks" instead of "h2o benchmarks", etc.

```suggestion
          python -m pytest $EXTRA_OPTIONS tests/benchmarks
```

sjperkins: I've placed the h2o benchmarks in a separate subdirectory. /cc @MrPowers @rrpelgrim. The test directory now looks as follows:

```bash
$ tree tests/
tests/
├── __init__.py
├── __pycache__
├── benchmarks
│   ├── __init__.py
│   └── h2o
│       ├── __init__.py
│       └── test_h2o_benchmarks.py
├── runtime
│   ├── __init__.py
│   ├── test_build.py
│   └── test_coiled.py
├── stability
│   ├── __init__.py
│   ├── test_array.py
│   └── test_shuffle.py
└── test_deadlock.py
```

.github/workflows/tests.yml
```diff
@@ -53,71 +56,224 @@ jobs:
...
 
       - name: Install coiled-runtime
+        env:
+          COILED_RUNTIME_VERSION: ${{ matrix.runtime-version }}
+        run: bash ci/scripts/install_coiled_runtime.sh
+
+      - name: Export environment
+        env:
+          COILED_RUNTIME_VERSION: ${{ matrix.runtime-version }}
+          PYTHON_VERSION: ${{ matrix.python-version }}
```
jrbourbeau, line 69: I don't think these environment variables are needed here (or other places we run `export_conda_env.sh`)

```suggestion
```

.github/workflows/tests.yml
```diff
@@ -53,71 +56,224 @@ jobs:
...
+        run: bash ci/scripts/export_conda_env.sh
+
+      - name: Build Coiled Software Environment
+        if: ${{ matrix.runtime-version == 'latest' }}
+        env:
+          DASK_COILED__TOKEN: ${{ secrets.COILED_BENCHMARK_BOT_TOKEN }}
+          GITHUB_EVENT_NAME: ${{ github.event_name }}
+          GITHUB_EVENT_NUMBER: ${{ github.event.number }}
+          PYTHON_VERSION: ${{ matrix.python-version }}
+          TEST_NAME: runtime
```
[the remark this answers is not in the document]
jrbourbeau: What if we `source <script>` instead of `bash <script>`? Will that remove the need to manually specify these `GITHUB_*` environment variables?

.github/workflows/tests.yml
```diff
@@ -53,71 +56,224 @@ jobs:
...
+          COILED_RUNTIME_VERSION: ${{ matrix.runtime-version }}
+        run: bash ci/scripts/install_coiled_runtime.sh
+
+      - name: Export environment
+        env:
+          COILED_RUNTIME_VERSION: ${{ matrix.runtime-version }}
+          PYTHON_VERSION: ${{ matrix.python-version }}
+        run: bash ci/scripts/export_conda_env.sh
+
+      - name: Build Coiled Software Environment
```
jrbourbeau, line 72: We're now building three times as many Coiled software environment. We're already building more than we need to since Coiled clusters always run on linux, there's no need to build a separate software environment for each os in the job matrix. This might be a good opportunity to move the Coiled software environment creation / deletion into their own jobs. Something like (psuedo-yaml below):

```yaml
jobs:
  software: <- build one software environment for each version of Python (with `runs-on: ["ubuntu-latest"]`)
  test_runtime: <- `needs: software`
  test_stability: <- `needs: software`
  test_benchmark: <- `needs: software`
  cleanup: <- We delete the Coiled software environments built in the `software` (`needs: [test_runtime, test_stability, test_benchmark]`)
```

That will both help us reduce the number of software environments we're building and cut down on repeated steps in each of the `test_*` jobs

.github/workflows/tests.yml
```diff
@@ -53,71 +56,224 @@ jobs:
...
           then
-            python ci/create_latest_runtime_meta.py
-            mamba env update --file latest.yaml
+            export EXTRA_OPTIONS="--run-latest"
           else
-            mamba install -c conda-forge coiled-runtime=${{ matrix.runtime-version }}
+            export EXTRA_OPTIONS=" "
           fi
 
+          python -m pytest $EXTRA_OPTIONS tests/runtime
```
jrbourbeau, line 100: It seems like we should move this to a separate script as well

.github/workflows/tests.yml
```diff
@@ -53,71 +56,224 @@ jobs:
...
+
+
+  tests_benchmarking:
+    name: Benchmarks ${{ matrix.os }}, Python ${{ matrix.python-version }}, Runtime ${{ matrix.runtime-version }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 120
+    # Required shell entrypoint to have properly activated conda environments
+    defaults:
+      run:
+        shell: bash -l {0}
```
jrbourbeau, line 118: I know this predates the changes made in this PR, but we should be able to move this to the top level of the workflow file https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaults. That will avoid us needing to repeat it in each job

jrbourbeau (commented):
Thanks @sjperkins!

.github/workflows/tests.yml
sjperkins: Removed

.github/workflows/tests.yml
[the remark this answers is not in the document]
sjperkins: I didn't manage to investigate this option.

.github/workflows/tests.yml
sjperkins: done

jrbourbeau (approved):
Thanks for all your work on this @sjperkins. Things look like they're in good shape, so let's merge this in to test out the periodic scheduled builds

sjperkins:
Likewise @jrbourbeau!

PR #93, opened 2022-04-26
Confirm upstream `to_parquet` performance fix

jrbourbeau:
Test to see that https://github.com/dask/dask/pull/8982 resolves the issue we were seeing in https://github.com/coiled/coiled-runtime/issues/79

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -131,7 +131,7 @@ jobs:
             export EXTRA_OPTIONS=" "
           fi
 
-          python -m pytest $EXTRA_OPTIONS tests
+          python -m pytest $EXTRA_OPTIONS tests/test_shuffle.py::test_shuffle_parquet
 
       - name: Remove Coiled software environment
         if: ${{ matrix.runtime-version == 'latest' }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,18 +26,15 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.9"]
-        runtime-version: ["latest"]
-        # os: ["ubuntu-latest"]
-        # python-version: ["3.7", "3.8", "3.9"]
-        # runtime-version: ["latest", "0.0.3"]
-        # include:
-        #   - python-version: "3.9"
-        #     runtime-version: "latest"
-        #     os: "windows-latest"
-        #   - python-version: "3.9"
-        #     runtime-version: "latest"
-        #     os: "macos-latest"
+        python-version: ["3.7", "3.8", "3.9"]
+        runtime-version: ["latest", "0.0.3"]
+        include:
+          - python-version: "3.9"
+            runtime-version: "latest"
+            os: "windows-latest"
+          - python-version: "3.9"
+            runtime-version: "latest"
+            os: "macos-latest"
 
     steps:
       - uses: actions/checkout@v2
@@ -131,7 +128,7 @@ jobs:
             export EXTRA_OPTIONS=" "
           fi
 
-          python -m pytest $EXTRA_OPTIONS tests/test_shuffle.py::test_shuffle_parquet
+          python -m pytest $EXTRA_OPTIONS tests
 
       - name: Remove Coiled software environment
         if: ${{ matrix.runtime-version == 'latest' }}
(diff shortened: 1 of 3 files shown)


jrbourbeau:
Comparing the `test_shuffle_parquet` times with the current `coiled-runtime` release

```python
1873.59s call     tests/test_shuffle.py::test_shuffle_parquet[True]
195.31s call     tests/test_shuffle.py::test_shuffle_parquet[False]
```

vs. the `main` branch of `dask`

```python
201.05s call     tests/test_shuffle.py::test_shuffle_parquet[False]
196.20s call     tests/test_shuffle.py::test_shuffle_parquet[True]
```

we can see https://github.com/dask/dask/pull/8982 indeed fixed the `to_parquet` performance issue. Nice job @jcrist! 

On an unrelated note, it looks like the changes here are causing our `s3_url` fixture to throw a `FileNotFoundError: []` error. It's not totally clear to me why this is. @sjperkins, if you have some time tomorrow, would you mind taking a look at this?

sjperkins:
> On an unrelated note, it looks like the changes here are causing our s3_url fixture to throw a FileNotFoundError: [] error. It's not totally clear to me why this is. @sjperkins, if you have some time tomorrow, would you mind taking a look at this?

Taking a look now

sjperkins:
Interestingly, each test parametrization is run twice.

https://github.com/coiled/coiled-runtime/runs/6190800692?check_suite_focus=true

```
tests/test_shuffle.py::test_shuffle_parquet[True] PASSED                 [ 94%]
tests/test_shuffle.py::test_shuffle_parquet[True] ERROR                  [ 94%]
tests/test_shuffle.py::test_shuffle_parquet[False] PASSED                [100%]
tests/test_shuffle.py::test_shuffle_parquet[False] ERROR                 [100%]
```

jrbourbeau:
> Interestingly, each test parametrization is run twice

It doesn't look like the test percentage is increasing on between the `PASSED` and `ERROR` lines. IIUC `pytest` setup/call/teardown each get their own line here, but setup/teardown aren't shown by default if they pass https://github.com/pytest-dev/pytest/issues/1004. So I don't think test parametrization is running twice

ian-r-rose:
Sure seems like `compute_as_if_collection` was doing something pretty pathological. I'd like to understand better exactly what was going on there, since it's also used in 10-20 other places throughout the codebase. Does anybody have any insights? If not, I'll poke at it a bit.

jrbourbeau:
Ah, the mechanism we're using to get the test name includes the parametrized value. I.e. 

```
s3://dask-io/test-scratch/test_shuffle_parquet[True]-a28dd2f21adb4125a448c717e5610644
```

Good find @sjperkins

jrbourbeau:
Are files actually getting written to that bucket? Since the test call itself is passing, my guess is yes, but it would be good to confirm. If so, that would mean creating/writing to that bucket works, but deleting doesn't, which we should open an issue over in `s3fs` for

ian-r-rose:
I left a comment on the upstream issue yesterday: I think we found a HLG bug.

sjperkins:
> Ah, the mechanism we're using to get the test name includes the parametrized value. I.e.
> 
> ```
> s3://dask-io/test-scratch/test_shuffle_parquet[True]-a28dd2f21adb4125a448c717e5610644
> ```
> 
> Good find @sjperkins

I'll take credit for pushing the print statement up, but I think you're deserve the credit for spotting that we're breaking S3 bucket naming rules? https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html :-)

sjperkins:
> > Ah, the mechanism we're using to get the test name includes the parametrized value. I.e.
> > ```
> > s3://dask-io/test-scratch/test_shuffle_parquet[True]-a28dd2f21adb4125a448c717e5610644
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Good find @sjperkins
> 
> I'll take credit for pushing the print statement up, but I think you deserve the credit for spotting that we're breaking S3 bucket naming rules? https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html :-)

More correctly, square brackets are to be avoided in s3 object naming conventions: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

PR #99, opened 2022-04-27
Unskip `test_repeated_merge_spill`

jrbourbeau:


jrbourbeau:
Alright, it looks like `test_repeated_merge_spill` is passing now when we use the `main` `distributed` branch (see [this CI build](https://github.com/coiled/coiled-runtime/runs/6196671046?check_suite_focus=true)) 🎉 Let's go ahead and unskip this test now that things have been resolved upstream

jrbourbeau:
This failure is due to a recent `traitlets` release but should be fixed soon (xref https://github.com/ipython/traitlets/issues/741)

jrbourbeau:
Re-running CI as I think the `traitlets` issue has been resolved

ncclementi:
Looks like things are working now, but the deadlock test is taking ~15min I wonder if this is a test we want to run all the time, or maybe mark it as slow or something?

PR #102, opened 2022-04-27
h2o-benchmarks with parquet data 

### LINKED ISSUE #4: Create GHA to automatically run & record Coiled Distribution H2O.ai benchmark Performance
Users and maintainers of dask want to understand how dask performs in a variety of benchmarks against a variety of competitive technologies.  One such benchmark is the [H2O.ai benchmark](https://h2oai.github.io/db-benchmark/).

I propose using the H2O.ai repository to create datasets of size 0.5GB, 5GB, and 50GB.  These csvfiles should be stored in s3://coiled-data/h2o-ai/XYZ-SIZE-CSVFILES/*.csv.  Alternatively, MPowers has worked with the H2O.ai team and may have access to the test data.

The benchmark should be executed on the Coiled Cloud infrastructure using default dask worker configurations, and the [sample environment](https://github.com/h2oai/db-benchmark#example-environment) described in the benchmark.

I would propose we record and report the following:

- Time to spin up infrastructure
- Mean time to load the data from S3
- Mean time to execute each test

Failure mode, if/when it occurs.

ncclementi:
- [x] Closes #4


tests/test_h2o_benchmarks.py
```diff
@@ -3,105 +3,123 @@
...
+                "id4": "Int32",
+                "id5": "Int32",
+                "id6": "Int32",
+                "v1": "Int32",
+                "v2": "Int32",
+                "v3": "float64",
+            },
+            storage_options={"anon": True},
+        )
+    elif request.param.endswith("parquet"):
```
jrbourbeau, line 48: Since there's only two options (csv vs. parquet) we can just use `else` here 

```suggestion
    else:
```

ncclementi:
@jrbourbeau So I uncomment the tests to run with the 50GB parquet files, and I think the failures are related to timeout. Would you suggest we only include the 5GB cases? or are they still taking a while? 

Not sure when I'll be able to get back to this, feel free to push to the branch to get it through.

ncclementi:
@jrbourbeau Should we comment out the 5GB case for parquet, or should we leave it uncommented. Otherwise, I think this is ready for a review.

jrbourbeau (approved):
Thanks @ncclementi!

PR #105, opened 2022-04-28
Parquet benchmark

bryanwweber:
Adds some (timing?) benchmarks for Parquet I/O on cloud storage.

TODO:

- [x] Add `gcsfs` to the software environment
- [x] Add `fastparquet` to the software environment
- [x] Handle removal of `require_extensions` kwarg

@jrbourbeau @ncclementi Pointers on how to add new packages to the software environment for testing without borking the `coiled-runtime` build are appreciated 😄

### DIFF
--- a/tests/test_parquet_benchmarks.py
+++ b/tests/test_parquet_benchmarks.py
@@ -1,0 +1,137 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import pytest
+
+import dask.dataframe as dd
+from dask import datasets
+from dask.distributed import wait
+
+if TYPE_CHECKING:
+    from dask.distributed import Client
+
+try:
+    import fastparquet
+except ImportError:
+    fastparquet = False
+
+try:
+    import pyarrow
+except ImportError:
+    pyarrow = False
+
+FASTPARQUET_MARK = pytest.mark.skipif(
+    not fastparquet, reason="fastparquet is not installed"
+)
+PYARROW_MARK = pytest.mark.skipif(not pyarrow, reason="pyarrow is not installed")
+
+
+@pytest.fixture(
+    params=[
+        pytest.param("fastparquet", marks=FASTPARQUET_MARK),
+        pytest.param("pyarrow", marks=PYARROW_MARK),
+    ]
+)
+def engine(request):
+    return request.param
+
+
+def read_write_engines():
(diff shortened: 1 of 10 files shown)


hayesgb:
Why would we add fastparquet

bryanwweber:
@hayesgb I think we should benchmark with fastparquet. If it provides a performance benefit or penalty, we can use that to guide future work or recommendations. I don't think we need to add it to the meta for the Coiled runtime package, only for the tests that run in GH Actions here

jrbourbeau:
The credentials that are forwarded to Coiled clusters are temporary and how long they last until expiring depends on some setting in the corresponding AWS account. I _think_ our current default is 1 hr (or at least it used to be several months back). My best guess is we're hitting this expiration timeout and loosing S3 access. We should figure out if we can increase the default AWS credential timeout (cc @shughes-uk ?), somehow refresh the credentials on the workers, and/or get https://github.com/coiled/coiled-runtime/pull/53 in which will certainly help but may not fully solve this issue

ian-r-rose:
xref https://gitlab.com/coiled/cloud/-/issues/4559

ntabris:
Another option since this is your own AWS account would be to edit the instance profile attached to instances so that it grants S3 access. In general I think identity-based access is better than passing tokens from client to cluster.

shughes-uk:
I'd agree with Nat that identity based access is better for this. Otherwise I'd maybe suggest not relying on the AWS plugin and perhaps just passing your creds to `storage_options` ? That way it won't be generating temporary creds.

tests/test_parquet_benchmarks.py
```diff
@@ -0,0 +1,143 @@
...
+                pytest.fail(res.result(), False)
+
+
+def test_read_parquet(engine: str, small_client: Client) -> None:
+    """
+    This dataset is ~5 GiB on disk, although it will grow as Anaconda add more data.
+    Each file is a few hundred MiB to a few GiB on disk with one row group per file.
+    """
+    df = dd.read_parquet(
+        "s3://anaconda-package-data/conda/hourly/",
```
ncclementi, line 95: To be able to read the public data you need to add the `storage_options={"anon": True}` this should fix the access issue.

jrbourbeau:
Closing in favor of https://github.com/coiled/coiled-runtime/pull/130

PR #107, opened 2022-04-29
Add performance reports

jrbourbeau:
This uploads performance reports for tests


conftest.py
```diff
@@ -79,12 +83,13 @@ def small_cluster(request):
...
 
 @pytest.fixture
-def small_client(small_cluster):
+def small_client(small_cluster, request):
     with Client(small_cluster) as client:
         small_cluster.scale(10)
         client.wait_for_workers(10)
         client.restart()
-        yield client
+        with coiled.performance_report(f"{request.node.originalname}-{UNIQUE_ID}.html"):
```
ncclementi, line 91: With this, we are storing the performance reports on coiled, but I can't seem to find which is the bucket or storage that contains them. We should also maybe think about how to clean this up. 

I see that in the coiled account we have some performance reports but not all of them, not sure exactly what's happening

jrbourbeau: > I can't seem to find which is the bucket or storage that contains them

I chatted with @shughes-uk offline and she mentioned today all performance reports are stored in a centralized S3 bucket that only Coiled staff have access to. I don't think this is a blocker for us, but it is good to know. 

> We should also maybe think about how to clean this up.

@shughes-uk mentioned this shouldn't become a problem unless we create millions of performance reports. I'm inclined to not worry about this for now -- I think it will take us a while before we've created millions of these : )

jrbourbeau: > I see that in the coiled account we have some performance reports but not all of them, not sure exactly what's happening

@koverholt pointed out that there's currently a limit of 5 hosted performance reports per user. Per the [`coiled.performance_report` API docs](https://docs.coiled.io/user_guide/api.html#coiled.performance_report)

> Note each user is limited to 5 hosted reports with each a maximum file size of 10mb.

So if we want to keep performance reports around, then we should upload them to our own S3 bucket

ian-r-rose:
Depending on what you are looking for, the analytics page might get you this information for free

jrbourbeau:
> Depending on what you are looking for, the analytics page might get you this information for free

That's a good point. With both https://github.com/coiled/coiled-runtime/pull/102 and https://github.com/coiled/coiled-runtime/issues/79 it would/would have been nice to see a performance report when we notice that something is taking longer than expected. I think this is definitely in the category of "nice to have" but not required.

conftest.py
```diff
@@ -90,12 +93,15 @@ def small_cluster(request):
...
 @pytest.fixture
-def small_client(small_cluster):
+def small_client(small_cluster, s3, s3_report_url, request, tmp_path):
     with Client(small_cluster) as client:
         small_cluster.scale(10)
         client.wait_for_workers(10)
         client.restart()
-        yield client
+        local_file = str(tmp_path / "preformance-report.html")
+        with performance_report(local_file):
```
ian-r-rose, line 102: If we use the `coiled` version of performance reports, we should get hosting for free

ian-r-rose: Oof, good point, I had forgotten about that

ncclementi:
@jrbourbeau I added a life-cycle to the s3 prefix (`test-scratch/performance-reports`) where the files in this "directory" will be removed after 7 days of being added, I haven't found a way of removing them if not accessed, but I thing 7 days is probably enough time to download a file locally if needed. 

Would you be ok, would that?

ncclementi:
I'm not seeing any new performance report being uploaded to s3, the latest date is 05/27.
I think that some of the changes might have affected this, even though things are not failing...

conftest.py
```diff
@@ -147,3 +145,14 @@ def s3_url(s3, s3_scratch, request):
...
     s3.rm(url, recursive=True)
+
+
+@pytest.fixture
+def upload_performance_report(tmp_path, request, s3, s3_report_url):
+    def func(client):
+        local_file = str(tmp_path / "preformance-report.html")
+        with performance_report(local_file):
+            yield client
+        s3.put(local_file, s3_report_url + f"/{request.node.originalname}.html")
```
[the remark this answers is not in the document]
ncclementi: Never mind, we are providing s3 which has the credentials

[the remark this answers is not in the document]
jrbourbeau: We pass credentials when we create the `s3` instance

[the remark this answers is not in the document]
ncclementi: Then there has to be another problem because the files are not being uploaded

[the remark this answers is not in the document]
jrbourbeau: Yeah, I see what I did wrong. I'll have to push up a commit with a fix, but probably not today

conftest.py
```diff
@@ -133,16 +143,16 @@ def small_client(small_cluster, s3_cluster_dump_url, s3_storage_options, request
 def s3_storage_options():
     return {
         "config_kwargs": {"region_name": S3_REGION},
-        "key": os.environ["AWS_ACCESS_KEY_ID"],
-        "secret": os.environ["AWS_SECRET_ACCESS_KEY"],
+        # "key": os.environ["AWS_ACCESS_KEY_ID"],
```
ncclementi, line 146: Don't we need to uncomment these lines to be able to test things with CI?

conftest.py
```diff
@@ -133,16 +143,16 @@ def small_client(small_cluster, s3_cluster_dump_url, s3_storage_options, request
...
+        # "secret": os.environ["AWS_SECRET_ACCESS_KEY"],
     }
 
 
 @pytest.fixture(scope="session")
 def s3():
     return s3fs.S3FileSystem(
-        key=os.environ["AWS_ACCESS_KEY_ID"],
-        secret=os.environ["AWS_SECRET_ACCESS_KEY"],
+        # key=os.environ["AWS_ACCESS_KEY_ID"],
```
ncclementi, line 154: Same here, do we need to uncomment?

conftest.py
jrbourbeau: Whoop -- left over from local debugging

ncclementi:
Ci It's not finished yet but it seems things are working. I'm seeing performance reports uploaded. I wonder if we should log the link same way we do with cluster dumps, at the moment from CI I can't find where the performance reports are. 

Screen Shot 2022-06-10 at 3 55 03 PM


This is unrelated, but I'm seeing mismatches warnings. Should we worried about these ones? 
Screen Shot 2022-06-10 at 3 58 20 PM

ncclementi:
Current status and problems:

- [ live logging ](https://docs.pytest.org/en/6.2.x/logging.html#live-logs)to log performance report link as warning doesn't work on CI, even though it works locally. Hence, we can't have a link to performance reports unless we have a failure. The performance report gets uploaded to S3 but no link is logged. Not being able to get a performance report link it makes it impossible to find the performance report on S3. 

- When tests that rely on small client fixture fail the logging of the performance report works, but only when it fails.  

- For the case of `tests/stability/test_deadlock` which uses its own cluster/client I can't make the upload performance report work, it's not even uploading the file, even less will get a link

In conclusion, 

- It looks like CI is not giving us all the logs and haven't been able to find a way to do so when tests are passing.

- For the case of `tests/stability/test_deadlock` I'm not too worried in getting perf reports to work, although it'll be nice if it did.

cc: @ian-r-rose  what do you think we should do here?

jrbourbeau:
@ian-r-rose @ncclementi and I chatted about this offline and decided to put this on hold. This is a nice-to-have but also somewhat complicated. If this becomes nicer-to-have in the future we can revive this PR

PR #109, opened 2022-05-03
Update scheduled CI interval

jrbourbeau:
This PR updates how often we run our scheduled upstream build to once a week. Choosing to run the build early in the week to maximize time between when the build is run and when Dask is usual released (on Friday).

cc @sjperkins


sjperkins:
Thanks @jrbourbeau

PR #115, opened 2022-05-11
Upgrade Dask version

jrbourbeau:
This PR moves us closer to a new `coiled-runtime` release 

Closes https://github.com/coiled/coiled-runtime/issues/100

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -212,7 +212,12 @@ jobs:
           - python-version: "3.9"
             runtime-version: "latest"
             os: "macos-latest"
+          - python-version: "3.7"
+            runtime-version: "0.0.3"
+            os: "ubuntu-latest"
         exclude:
+          # `coiled-runtime=0.0.3` pins to a version of Dask
+          # that doesn't support Python 3.10
           - python-version: "3.10"
             runtime-version: "0.0.3"
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,10 +1,10 @@
 name: Tests
 on:
   push:
-    branches:
-      - main
-    tags:
-      - '*'
+    # branches:
+    #   - main
+    # tags:
+    #   - '*'
   pull_request:
   schedule:
     # Runs "At minute 1 on Sunday" (see https://crontab.guru)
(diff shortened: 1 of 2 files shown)


jrbourbeau:
Software environment creation seems to be stalled for Python 3.10... I'm going to cancel the build it's it's already been going for over 1.5 hours

jrbourbeau:
Okay, looks like we can't upgrade to the latest `click=8.1.3` release as it's causing us to pull in an older version of `coiled` because `coiled` requires `click>=7.1,<=8.0` to prevent versions of `distributed<2022.04.0` from failing due to https://github.com/dask/distributed/issues/6013.

@dchudz @shughes-uk @ntabris question, thoughts on:

1. Updating the minimum version of `distributed` in `coiled` to be `2022.04.0`? This would remove the need for pinning `click` at all in `coiled`, but my guess is this is too restrictive for all `coiled` users
2. Changing the click pin in `coiled ` to be `click>=7.1,<8.1`. This would allow us to install `click=8.0.4` (which has a Python 3.10 build on `conda-forge`) with `coiled` while continuing to avoid https://github.com/dask/distributed/issues/6013

dchudz:
(1) seems way too restrictive! It's hard to imagine going from "we don't recommend anything later than 2022.01" to "you must use at least 2022.04.0" in one leap.

any downsides to (2)?

hayesgb:
@jrbourbeau @dchudz -- Suggest we wait for 2022.05.1

jrbourbeau:
> (1) seems way too restrictive! 

Yep, thought so but still wanted to jot the option down

> any downsides to (2)?

Not that I'm aware of

> Suggest we wait for 2022.05.1

This PR isn't the actual release, but does add handling around dropping Python 3.7 and adding Python 3.10 with more recent Dask releases. We can always bump the `dask` / `distributed` version again before releasing

jrbourbeau:
See https://github.com/conda-forge/coiled-feedstock/pull/96 for the latest `coiled=0.2.1` release which should resolve the issues we're seeing in CI here by updating the `click` version constraint for `coiled`

jrbourbeau:
@ncclementi this PR should be ready for review -- could you take a look whenever you get a chance?

.github/workflows/tests.yml
```diff
@@ -218,6 +212,14 @@ jobs:
           - python-version: "3.9"
             runtime-version: "latest"
             os: "macos-latest"
+          - python-version: "3.7"
```
ncclementi, line 215: Did you mean 3.8, or we are still testing 3.7 against 0.0.3 only for ubuntu?

ncclementi: I see that makes sense, that was my only comment/question. I see CI is green so I'll go ahead and merge this one.

PR #123, opened 2022-05-19
Use consistent hash in cluster names

jrbourbeau:
Today for a single GHA workflow run we use a unique hash for each cluster name (for example):

```
test_shuffle-f769bd1c
test_array-23fb79f4
...
```

This PR makes it so that each cluster name uses the same hash:

```
test_shuffle-f769bd1c
test_array-f769bd1c
...
```

which makes it easier to trace clusters back to a specific GHA workflow run. This type of tracing is nice as, for example, it helps us get a better sense for how much each workflow run costs.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -64,12 +64,14 @@ dask.config.set(
     }
 )
 
+ID = uuid.uuid4().hex[:8]
+
 
 @pytest.fixture(scope="module")
 def small_cluster(request):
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Hrm, I forgot that the hash will be different across different jobs, so this isn't the correct approach. @dchudz pointed me to the `coiled.get_billing_activity` method which is more robust way to go

PR #125, opened 2022-05-19
Implement cluster dump for stability test failures

### LINKED ISSUE #45: Automatically capture (and upload) cluster dump on integration test failure
Follow-up to https://github.com/coiled/coiled-runtime/pull/41

I'm not actually sure how useful a cluster dump would be in many cases. But my first question in seeing https://github.com/dask/distributed/issues/6125 was to think, "can we get a cluster dump for this"? The worker event logs might be helpful.

An S3 bucket that the GHA runners have access to might be a logical place to upload (since `dump_cluster_state` already supports it).

ncclementi:
- [x] Closes #45

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -90,22 +90,33 @@ def small_cluster(request):
         yield cluster
 
 
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    # execute all other hooks to obtain the report object
+    outcome = yield
+    rep = outcome.get_result()
+
+    # set a report attribute for each phase of a call, which can
+    # be "setup", "call", "teardown"
+
+    setattr(item, "rep_" + rep.when, rep)
+
+
 @pytest.fixture
-def small_client(small_cluster, s3_cluster_dump_url, s3_storage_options):
+def small_client(small_cluster, s3_cluster_dump_url, s3_storage_options, request):
     with Client(small_cluster) as client:
         small_cluster.scale(10)
         client.wait_for_workers(10)
         client.restart()
-        try:
-            yield client
-        except Exception:
-            cluster_dump = strtobool(os.environ.get("CLUSTER_DUMP", "false"))
-            print(f"{cluster_dump=}")
-            if cluster_dump:
-                dump_path = os.path.join(s3_cluster_dump_url, small_cluster.name)
-                print(f"Cluster state dump can be found at: {dump_path}")
-                client.dump_cluster_state(dump_path, **s3_storage_options)
-            raise
+
+        yield client
+
+        cluster_dump = strtobool(os.environ.get("CLUSTER_DUMP", "false"))
+        print(f"{cluster_dump=}")
+        if cluster_dump and request.node.rep_call.failed:
+            dump_path = os.path.join(s3_cluster_dump_url, small_cluster.name)
+            print(f"Cluster state dump can be found at: {dump_path}")
+            client.dump_cluster_state(dump_path, **s3_storage_options)
 
 
 S3_REGION = "us-east-2"
--- a/conftest.py
+++ b/conftest.py
@@ -93,8 +93,7 @@ def small_client(small_cluster, s3, s3_scratch, s3_cluster_dump_url):
             if cluster_dump:
                 print(f"Cluster state dump can be found at: {s3_cluster_dump_url}")
                 client.dump_cluster_state(s3_cluster_dump_url)
-            else:
-                pass
+            raise
 
 
 S3_REGION = "us-east-2"
--- a/conftest.py
+++ b/conftest.py
@@ -103,7 +103,7 @@ def pytest_runtest_makereport(item, call):
 
 
 @pytest.fixture
-def small_client(small_cluster, s3_cluster_dump_url, s3_storage_options, request):
+def small_client(small_cluster, s3_cluster_dump_url, request):
     with Client(small_cluster) as client:
         small_cluster.scale(10)
         client.wait_for_workers(10)
@@ -116,7 +116,7 @@ def small_client(small_cluster, s3_cluster_dump_url, s3_storage_options, request
         if cluster_dump and request.node.rep_call.failed:
             dump_path = os.path.join(s3_cluster_dump_url, small_cluster.name)
             print(f"Cluster state dump can be found at: {dump_path}")
-            client.dump_cluster_state(dump_path, **s3_storage_options)
+            client.dump_cluster_state(dump_path)
 
 
 S3_REGION = "us-east-2"
--- a/conftest.py
+++ b/conftest.py
@@ -90,6 +90,8 @@ def small_cluster(request):
         yield cluster
 
 
+# this code was taken from pytest docs
+# https://docs.pytest.org/en/latest/example/simple.html#making-test-result-information-available-in-fixtures
 @pytest.hookimpl(tryfirst=True, hookwrapper=True)
 def pytest_runtest_makereport(item, call):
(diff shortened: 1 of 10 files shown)


conftest.py
```diff
@@ -79,12 +80,21 @@ def small_cluster(request):
...
+        try:
+            yield client
+        except Exception:
+            cluster_dump = strtobool(os.environ.get("CLUSTER_DUMP", "false"))
+            print(f"{cluster_dump=}")
+            if cluster_dump:
+                print(f"Cluster state dump can be found at: {s3_cluster_dump_url}")
+                client.dump_cluster_state(s3_cluster_dump_url)
+            else:
+                pass
```
jrbourbeau, line 97: I think we'll still want to `raise` the exception here so that test failures show up 

```suggestion
            raise
```

ncclementi:
Declarative is broken, for now, canceling the workflows until it's fixed.

ncclementi:
@jrbourbeau We made progress, now we are having PermissionError writing the cluster dump, I'm not sure if this is on our end or if it's a bug. See https://github.com/coiled/coiled-runtime/runs/6682060242?check_suite_focus=true#step:6:177

Currently only doing the dump for "latest" because the url dump was added after dask=2022.1.0

ncclementi:
So I did a bit more digging and found this issue https://github.com/fsspec/s3fs/issues/139 which was technically solved in this PR  https://github.com/fsspec/s3fs/pull/140/files. According to this code https://github.com/fsspec/s3fs/pull/140/files#diff-044e3dd45e47667275a210a6ac0033c31687eaea7d4e33e3b21b34baa1a050daR975-R978
I would expect "LocationConstraint" to match `region_name` because of how it's designed. But it seems that's not happening and I'm not sure why we have the issue on CI, @ntabris if you have any inputs they'll be much appreciated. 

cc: @jrbourbeau

ncclementi:
So After chatting with Gabe, he pointed out that the issue is on ffspec and s3fs, I tried to the best of my understanding explain the issues in https://github.com/fsspec/filesystem_spec/issues/974 and https://github.com/fsspec/s3fs/issues/627 but I might have missed some parts. 

I'm not entirely sure how to solve the whole problem, but this is getting more involved than what we expected to be able to have the cluster dumps on s3. I think we should put this PR on hold as there are other things of higher priority. 

cc: @jrbourbeau

gjoseph92:
The easy workaround would probably be to add `"auto_mkdir": False` to `s3_storage_options`: https://github.com/fsspec/filesystem_spec/blob/06c400335cb7d382cd57d94f93a4b30157dad9b2/fsspec/core.py#L308-L310

conftest.py
```diff
@@ -89,22 +90,46 @@ def small_cluster(request):
...
 S3_REGION = "us-east-2"
 S3_BUCKET = "s3://coiled-runtime-ci"
 
 
 @pytest.fixture(scope="session")
 def s3_storage_options():
-    return {"config_kwargs": {"region_name": S3_REGION}}
+    # add "auto_mkdir": False until this is fixed
+    # https://github.com/fsspec/filesystem_spec/issues/974
+    return {"config_kwargs": {"region_name": S3_REGION, "auto_mkdir": False}}
```
ncclementi, line 132: @gjoseph92 Maybe I'm doing something wrong, I tried this as you suggested but I'm getting a TypeError https://github.com/coiled/coiled-runtime/runs/6712573736?check_suite_focus=true#step:6:167

jrbourbeau: I believe you want to add `"auto_mkdir": False` at the same level as `config_kwargs`. Something like (note I've not tested the code suggestion below):

```suggestion
    return {"config_kwargs": {"region_name": S3_REGION}, "auto_mkdir": False}
```

ncclementi:
It looks like the change suggested by @gjoseph92 did not work, we are still getting a permission error
see https://github.com/coiled/coiled-runtime/runs/6714230545?check_suite_focus=true#step:6:177 but seems to come from a different place by looking at the traceback. 

At this point, I'm inclined to punt this. 

cc: @jrbourbeau

jrbourbeau:
@ncclementi when you get a chance, can you check S3 to see if cluster dumps are being uploaded as expected?

conftest.py
```diff
@@ -89,22 +90,51 @@ def small_cluster(request):
...
         client.wait_for_workers(10)
         client.restart()
+
         yield client
 
+        cluster_dump = strtobool(os.environ.get("CLUSTER_DUMP", "false"))
+
+        if cluster_dump and request.node.rep_call.failed:
+            dump_path = os.path.join(s3_cluster_dump_url, small_cluster.name)
+            print(f"Cluster state dump can be found at: {dump_path}")
```
ncclementi, line 120: This print is not showing even when running pytest with -s as an option.

ncclementi:
> @ncclementi when you get a chance, can you check S3 to see if cluster dumps are being uploaded as expected?

@jrbourbeau Yes, it worked as expected!  Although for some reason we don't see the print with the link to the cluster dump.

tests/stability/test_array.py
```diff
@@ -11,3 +11,7 @@ def test_rechunk_in_memory(small_client):
 def test_rechunk_out_of_memory(small_client):
     x = da.random.random((100000, 100000))
     x.rechunk((50000, 20)).rechunk((20, 50000)).sum().compute()
+
+
+def test_dump(small_client):
```
ncclementi, line 16: remove this when before merging

ncclementi:
This seems to be sort of working. Problems I noticed:
I only see 4 cluster dumps this is probably only the ubuntu 3.8, 3.9, 3.10  and macos 3.9 on the folder

Screen Shot 2022-06-09 at 5 18 18 PM

but the windows case is having some problem with how the directory is generated,

Screen Shot 2022-06-09 at 5 18 40 PM
 
The print is now working, but broken for windows, 
```python
---------------------------- Captured log teardown ----------------------------
ERROR    coiled-runtime:conftest.py:122 Cluster state dump can be found at: s3://coiled-runtime-ci/test-scratch/cluster_dumps\test_array-07139b63
```

I believe this is because we are relying on `os.path.join`

jrbourbeau:
Thanks @ncclementi -- I just pushed a commit to fix the windows issue. That would mean we would get 5 cluster dumps in the expected location, which is the number of dumps I would expect to see

ncclementi:
If one of the benchmark tests failed is because I accidentally killed a cluster. my bad

ncclementi:
Alright, things work. I'll remove the test that causes the failure and after that, we can merge!

Screen Shot 2022-06-07 at 6 32 50 PM

ncclementi:
@jrbourbeau I'm trying to use request to get the test name, I wasn't able to run this locally, I'm trying something which is a bit of a shot in the dark, if it doesn't work, we can revert it and leave it for a future PR. Is getting late, so I'm going to log off for today.

ncclementi:
Alrighty!! Things are working! I set things up to have two different tests that would fail on purpose, that were named `test_dump` and `test_other_dump` and we got a cluster dump for each case and separate error messages. 

Screen Shot 2022-06-10 at 11 05 07 AM

and we can see all the cluster dumps in S3 as:
cluster_dump


We just need to remove the tests added to test this workflow, and this is ready! 
cc: @jrbourbeau

tests/stability/test_array.py
```diff
@@ -11,3 +11,11 @@ def test_rechunk_in_memory(small_client):
...
```
ncclementi, line 16: remove this

jrbourbeau (approved):
Thanks for all your work on this @ncclementi

PR #126, opened 2022-05-20
Use spot instances in stability tests

jrbourbeau:
This PR updates our stability tests to automatically use spot instances by default. This will make sure we're resilient to workers randomly leaving the cluster. 

Closes https://github.com/coiled/coiled-runtime/issues/122

EDIT: This also has a small cost-savings benefit, but my main motivation for this change is making sure our stability tests are robust when workers go offline

### DIFF
--- a/tests/stability/conftest.py
+++ b/tests/stability/conftest.py
@@ -1,0 +1,14 @@
+import pytest
+
+
+def pytest_collection_modifyitems(config, items):
+    # Ensure stability tests use spot instances by default
+    marker = pytest.mark.backend_options(spot=True)
+    for item in items:
+        # Add a module-level `spot=True` backend option marker if one doesn't already exist
+        module = item.parent
+        if not any(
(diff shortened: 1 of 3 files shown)


jrbourbeau:
@ncclementi do you have a moment to look at this?

tests/stability/conftest.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+
+def pytest_collection_modifyitems(config, items):
+    # Ensure stability tests use spot instances by default
```
ncclementi, line 5: Were you able to check this works as expected? I wonder how can we check that stability is always using spot instances.

jrbourbeau: Right now it's not easy to get that information (I learned this from @ntabris) so I just added some print statements locally to confirm that `backend_options` are getting piped correctly

ncclementi (approved):

PR #127, opened 2022-05-20
Run upstream build daily instead of weekly

jrbourbeau:
Given the relatively low cost estimates from @ntabris over in https://github.com/coiled/coiled-runtime/issues/122#issuecomment-1132024820, I'll propose we run our upstream CI builds once a day instead of once a week

@hayesgb


hayesgb:
LGTM

PR #128, opened 2022-05-23
Add staging environment

marcosmoyano:
Also push coiled-runtime senvs to staging

@jrbourbeau based on what we talked about today this should be enough. I created the account/user pair in staging and added the API token for that user as a secret. Please let me know if I'm missing something

### DIFF
--- a/.github/workflows/software-environments.yml
+++ b/.github/workflows/software-environments.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Install Coiled
         run: python -m pip install coiled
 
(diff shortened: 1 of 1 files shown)


jrbourbeau (commented):
Thanks @marcosmoyano! I pushed a small commit to temporarily run this workflow on PRs so we can make sure everything runs as expected

jrbourbeau (approved):
Looks like everything is working as expected -- thanks @marcosmoyano!

PR #131, opened 2022-05-24
Add `xgboost` and `dask-ml`

hayesgb:
In this PR we add both `xgboost` and `dask-ml` as dependencies, and `pytest-timeout` as a development dependency.  To build this software environment remotely, we follow the instructions for building the experimental `coiled-runtime` locally and install it into a conda environment.  We also create an environment file, either manually or with `conda lock` and build the senv with
```
import coiled
coiled.create_software_environment(name="my-senv", conda="environment.yml")
```

Then we follow the instructions outlined in the `README.md` to set the `COILED_SOFTWARE_NAME`.

We also observe that, when testing locally, `dask.config.set` does not override [setting the coiled software environment.](https://github.com/coiled/coiled-runtime/blob/1fbfc124a6f497855b767511b668a7c59010a9ca/conftest.py#L63), and we get the default software image instead of the manually created image from above.  Adding the `software` parameter in the `Cluster` object fixes this issue.


README.md
```diff
@@ -37,7 +37,7 @@ conda install -c conda-forge conda-build
 conda build recipe -c conda-forge --output-folder dist/conda --no-anaconda-upload
 
 # Install the built `coiled-runtime` metapackage
-conda install -c conda-forge -c ./dist/conda/ coiled-runtime
+conda install -c ./dist/conda/ coiled-runtime
```
jrbourbeau, line 40: I think we want to keep `-c conda-forge` to make sure we can install dependencies like `numpy`, `pandas`, etc. Though you probably need to make `conda-forge` lower priority than `coiled` as `coiled-runtime` now exists on `conda-forge`

jrbourbeau:
Okay this is working now, but currently blocked (as @ncclementi mentioned offline) on `xgboost>=1.6` not being available on `conda-forge` (xref https://github.com/conda-forge/xgboost-feedstock/issues/89, https://github.com/conda-forge/xgboost-feedstock/pull/88)

ncclementi:
I don't think the xgboost test run? I'm a bit confused about where the test is, trying to take a look at it now. It seems the test was added under `/integrations` and we are not running that directory. I think I'll move it to runtime for now to try to run it.

ci/scripts/install_coiled_runtime.sh
```diff
@@ -11,8 +11,13 @@ then
   mamba env update --file latest.yaml
 else
   mamba install -c conda-forge coiled-runtime=$COILED_RUNTIME_VERSION
+  # TODO: Remove before merging
```
ncclementi, line 14: The latest soft env runs, the 003 and 004 runtimes don't. I thought by adding this mamba install here but it seems that doesn't do it.

jrbourbeau: We'll want to just skip `test_xgboost_distributed_training` on the 0.0.3 and 0.0.4 builds. We should be able to move the `mamba install` below and things should pass

ncclementi: I see, when you mean skip would be something like this? 
```python
import pytest
from packaging.version import Version
import coiled_runtime

@pytest.mark.skipif(
    Version(coiled_runtime.__version__) < Version("0.1.0"),
    reason="no xgboost or dask-ml packages in runtime 0.0.3 and 0.0.4",
)
def test_xgboost_distributed_training(small_client):
```

I noticed that `coiled-runtime` is in the package list but I can't import it. I thought of using this from the context but not sure what's the best practice here. 
https://github.com/coiled/coiled-runtime/blob/5f71950fed48d4de8e200e2af94610349b1e1cea/conftest.py#L66-L80

tests/runtime/test_xgboost.py
```diff
@@ -0,0 +1,44 @@
+import dask.dataframe as dd
+import pytest
+
+# `coiled-runtime=0.0.3,0.0.4` don't contain `xgboost`
+xgb = pytest.importorskip("xgboost")
+
+
+def test_xgboost_distributed_training(small_client):
+    # `coiled-runtime=0.0.3,0.0.4` don't contain `dask_ml`
+    dask_ml = pytest.importorskip("dask_ml")
```
ncclementi, line 10: @jrbourbeau Is there a reason behind this `importorskip` being inside the tests instead of outside?

jrbourbeau: This isn't strictly needed now, but given this module is for `xgboost`-related test, I can see us adding more tests in the future that use `xgboost` but not `dask-ml`. In that case we'll still want to skip this test, but run other ones that don't rely on `dask-ml`. So putting this `importorskip` inside the test is a bit forward looking

ncclementi: Oh I see, that makes sense! Thanks for the explanation :)

PR #154, opened 2022-06-01
Push docker images of `coiled-runtime`

### LINKED ISSUE #142: Build and push docker images of runtime to public docker hub
Currently the runtime/default Coiled software environments are built by Coiled from the conda env file. The result is that Coiled builds a docker image which is stored in our public ECR registry.

We (platform) would like it if your CI instead built a docker image and pushed this to docker hub, then created the Coiled software environment from that (using `--container` arg instead of `--conda`).

If the default software environments are pulled from docker hub, then we don't have expense of egress on our public ECR registry. Rate limiting *shouldn't* be a problem because each node pulling image would have different IP address.

(Less important but this also makes local dev easier because we could much more quickly create software env from public images in docker hub. Some of us are already building our own images and pushing to docker hub to do this, but it seems like this might as well be done automatically for runtime using CI.)

### LINKED ISSUE #147: Publish nightly Coiled software environments
In https://github.com/coiled/coiled-runtime/pull/128 we automated creating Coiled software environment for released versions of `coiled-runtime` on staging.coiledhq.com. As a follow-up we should also update our software environment CI to create a nightly Coiled software environment that uses the `coiled-runtime` nightly build. This will increase the chance that we catch regressions/bugs/etc early.

ncclementi:
- [x] Closes #142 
- [x] Closes #147


ntabris:
I suspect v3 of the github action thing works differently than v1 used here:

https://github.com/coiled/docker/blob/main/.github/workflows/ci-deploy.yml#L20

jrbourbeau:
> Coiled currently overwrites the entrypoint on images, we could make this configurable but often this is the right thing to do and it's what we always do at present

What is the entrypoint overwritten with currently? 

> I think Coiled lets you specify base image and then conda install additional packages. That's also maybe a normal thing to do. Will that work or would this have to be changed to mamba install?

I'm okay not having these docker images not being used to as a base for having separate `conda` packages installed. At least to start. FWIW I would love to see coiled software environments switch to using mamba instead of conda for environment solves

ntabris:
> What is the entrypoint overwritten with currently?

To not use an entrypoint (so `""`). I get the sense there's a mix of how people use `entrypoint`, often it's something that makes sense interactively but not on cluster (e.g., start jupyter server), but sometimes I guess it's stuff like activating a conda env. Maybe it's important to expose this. Regardless, it makes things a little simpler at the moment for me/platform if these images don't require `entrypoint`.

> I'm okay not having these docker images not being used to as a base for having separate conda packages installed

I guess I'm okay with this too, but it seems *better* if it's easy to use this as base for adding additional conda packages, no? Or is that not something you think people should be doing?

.github/workflows/runtime-dockerhub.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      
+      - name: Build and push ${{ matrix.image }} image
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: coiled/${{ matrix.image }}
```
jrbourbeau, line 29:
```suggestion
          tags: coiled/${{ matrix.image }}:latest
```

.github/workflows/runtime-dockerhub.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      
+      - name: Build and push ${{ matrix.image }} image
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: coiled/${{ matrix.image }}:latest
```
jrbourbeau, line 29:
```suggestion
          tags: coiledci/${{ matrix.image }}:latest
```

jrbourbeau (commented):
Thanks for your work on this @ncclementi. It looks like we're able to build and push images up at this point. As few next steps (some of which you and I already discussed offline):

- [x] Update docker hub credentials to be able to push to the `coiled` account
- [ ] Add `build-args` for passing in the appropriate conda environment file into `Dockerfile` at build time
- [ ] Consolidate GHA workflows so that we build / push docker images and create Coiled software environments all in a single workflow. I'd advocate for moving the docker steps over into the existing `.github/workflows/software-environments.yml` workflow
- [ ] Confirm everything is working as expected by pushing up new Coiled software environments to staging and then try to run the Coiled quickstart

.github/workflows/runtime-dockerhub.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+      
+      - name: Build and push ${{ matrix.image }} image
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: coiledci/${{ matrix.image }}:latest
```
jrbourbeau, line 29:
```suggestion
          tags: coiled/${{ matrix.image }}:latest
```

.github/workflows/runtime-dockerhub.yml
```diff
@@ -0,0 +1,29 @@
...
+      - environments/**
+  pull_request: # only to test while developing
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image: [runtime_0-0-3_py39]
```
jrbourbeau, line 16: Currently this is pushing to `coiled/runtime_0-0-3_py39:latest` on docker hub. However, I think `coiled/coiled-runtime-py39:0.0.3` uses a more conventional name. @ncclementi thoughts on using this naming convention instead?

jrbourbeau, line 16: Currently this is pushing to `coiled/runtime_0-0-3_py39:latest` on docker hub. However, I think `coiled/coiled-runtime-py39:0.0.3` uses a more conventional name. @ncclementi thoughts on using this naming convention instead?

ncclementi: yes, using 0.0.3 seems a better choice. Now that things are working I will get to move things to the software yaml and use build-args. Thank you @jrbourbeau for troubleshooting the credential issue.

jrbourbeau: After looking at [Dask's docker images](https://hub.docker.com/r/daskdev/dask/tags), I'd like to revise my previous suggestion and propose we go with `coiled/coiled-runtime:0.0.3-py3.9` instead. This way, we only have a single docker repository for all our runtime images and all the additional information on needs (i.e. runtime + Python version) are specified in the tag. What do you think @ncclementi?

ncclementi:
@jrbourbeau It seems like the logic to build the soft envs with build-args is correct (test pending) but I'm having permission problems again when trying to push. If you can take a look at what could be happening,  I know that last week you did something that worked, I wonder if the token expired or something.

.github/workflows/runtime-dockerhub.yml
ncclementi: That seems reasonable to me. Happy to modify the code accordingly.

ncclementi: I think we will need to give the bot access to the right repo in dockerhub, @jrbourbeau  can you take care of that before I push to see if it works?

jrbourbeau: Yep, on it now. I just created a new `coiled/coiled-runtime` repo in the `coiled` docker hub org (see https://hub.docker.com/repository/docker/coiled/coiled-runtime). The bot should have permissions to push to that repo too

jrbourbeau:
Looks like you're updates were successful 👍

ncclementi:
Alright the push seems to work, I was able to create a software environment from the container, but I didn't run anything in it. I'm uncommenting the code that I commented for testing purposes. 

I only have one concern before declaring this as done, which is what would happen when this gets merged to main. I assume it won't ruin the software environment job as `/environments` wasn't modified, is that correct? 
I'm not sure how overwriting images on dockerhub works.

.github/workflows/software-environments.yml
```diff
@@ -42,4 +42,21 @@ jobs:
           export PYTHON_VERSION_FORMATTED=$(echo "${{ matrix.python-version }}" | sed 's/\.//g' )
           export COILED_SOFTWARE_ENV_NAME=coiled-runtime-$RUNTIME_VERSION_FORMATTED-py$PYTHON_VERSION_FORMATTED
 
-          coiled env create --name coiled/$COILED_SOFTWARE_ENV_NAME --conda environments/$COILED_SOFTWARE_ENV_NAME.yml
\ No newline at end of file
+          #coiled env create --name coiled/$COILED_SOFTWARE_ENV_NAME --conda environments/$COILED_SOFTWARE_ENV_NAME.yml
```
ncclementi, line 45: whoops, I forgot to uncomment this line. Since tests are running I'll let the CI finish and then push the change.

.github/workflows/software-environments.yml
```diff
@@ -42,4 +42,21 @@ jobs:
           export PYTHON_VERSION_FORMATTED=$(echo "${{ matrix.python-version }}" | sed 's/\.//g' )
           export COILED_SOFTWARE_ENV_NAME=coiled-runtime-$RUNTIME_VERSION_FORMATTED-py$PYTHON_VERSION_FORMATTED
 
-          coiled env create --name coiled/$COILED_SOFTWARE_ENV_NAME --conda environments/$COILED_SOFTWARE_ENV_NAME.yml
\ No newline at end of file
+          coiled env create --name coiled/$COILED_SOFTWARE_ENV_NAME --conda environments/$COILED_SOFTWARE_ENV_NAME.yml
```
jrbourbeau, line 45: Can we move this step to after we've pushed the corresponding docker image and then update the Coiled software environment creation step to use `--container` instead of `--conda`? I believe that was the ultimate motivating factor behind https://github.com/coiled/coiled-runtime/issues/142

ncclementi: Testing this now on as PR, commenting out the `test.yaml` workflow. Will uncomment tests after seeing if software build from container works.

jrbourbeau:
> For prod things are working

Sorry, I should have mentioned this earlier but I believe these software environments are being used as the default envs for all Coiled clusters. In the future we should probably not touch prod until we've tested on staging that everything works.

ncclementi:
> Sorry, I should have mentioned this earlier but I believe these software environments are being used as the default envs for all Coiled clusters. In the future we should probably not touch prod until we've tested on staging that everything works.

Oh shoot, I did not know that, sorry about this. What should be the course of action here? Things on staging are not necessarly working, the bot seems to not have access.

jrbourbeau:
I'm running tests over in https://github.com/coiled/coiled-runtime/pull/162 which should pull in the default software environments for half of the CI jobs. If those pass, then we should be good. Otherwise, we'll just want to rebuild the software environments using `--conda` again

jrbourbeau:
Alright, prod software environments have been rebuilt to use `--conda` again and we're rerunning CI on `main` to confirm that everything is back to normal (xref https://github.com/coiled/coiled-runtime/issues/163#issuecomment-1149075452)

Also, @marcosmoyano updated permissions so the bot should be able to push to staging now (xref https://github.com/coiled/coiled-runtime/pull/162#issuecomment-1149077424)

ncclementi:
For the context of this PR then, should we comment the prod server out then?

jrbourbeau:
> For the context of this PR then, should we comment the prod server out then?

Yeah, that's should do it

ncclementi:
Ok, so things are green for staging, @marcos the soft envs built from container should be available on staging now.

 I'll uncomment the test.yml so we run the whole suite, now those will use the conda build ones.

ncclementi:
I think we should not merge this until we confirm things on staging work and then get deployed on prod.

The new `software-environment.yml` will only run on main if the `environments/` gets modified, so technically there is no risk in merging it, but it might be a good idea to wait until we know things work, and maybe trigger it manually.

ntabris:
I've verified that the `mamba`-based images do work in staging.

In particular, I did this

```
coiled.create_software_environment(name="runtime-mamba-container", container="coiled/coiled-runtime:0.0.3-py3.8")
```

and then was able to spin up cluster in staging using "runtime-mamba-container".

But (take note!) you'll need the not-yet-released `coiled` client in order to build software environment that defaults to not overriding the entrypoint.

What version of `coiled` is used in CI? (Latest? The one from runtime?)

ncclementi:
CI uses what is installed with the runtime, since this is not pinned it should be "latest".

ncclementi:
> What magic is in the entrypoint that's currently used in prod? Would it be possible to just add a couple of lines to our `Dockerfile` here to make things work?

@ntabris would you mind commenting on this to see if we can make this work regarding the coiled-runtime version?

ntabris:
> I'm also wondering how backwards compatible things are. Will things also just work with, for example, coiled=0.2.4?

You need to *create* software environment with newer coiled, but version of coiled *using* the software environment doesn't matter.

(The logic is a little convoluted, but the reason was to make the least disruption to existing customers. Basically if you have existing software environments or if you haven't updated your client and make software environments, then nothing will change.)

jrbourbeau:
> You need to create software environment with newer coiled, but version of coiled using the software environment doesn't matter.

Ah, grand. We can definitely create software envs with the latest `coiled` release, but since these are being used as the default software environments for all clusters, I wanted to make sure we had some level of backwards compatibility. It sounds like we do 👍

ncclementi:
I modify things to include in this PR a work to close  #147 . Things are working as expected in the CI, I thought since the work was related we might as well review it together.

.github/workflows/software-environments-nightly.yml
```diff
@@ -0,0 +1,77 @@
+name: Create software environments
+
+on:
+  pull_request: #remove after testing
```
ncclementi, line 4: Remove this line

ntabris:
FYI I took a look at `dask-engineering/coiled-runtime-nightly-py390-0-4a2206080023` senv recently build on prod and it looks good (it's container-based senv and will use entrypoint in image.)

ncclementi:
@ntabris Yes I was testing with that account over here https://github.com/coiled/coiled-runtime/pull/165 to check if things worked, but the logic to use only one workflow was a bit messy so I decided to move things here and make it cleaner 

The CI, that just run here, should have created nightly software environments built on staging. In case you want to use them.

.github/workflows/software-environments-nightly.yml
ncclementi:
```suggestion
```

.github/workflows/software-environments-nightly.yml
```diff
@@ -0,0 +1,77 @@
...
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+      
+      - name: Build and push runtime nightly ${{ env.RUNTIME_NIGHTLY_VERSION }} python ${{ matrix.python-version }}
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: coiled/coiled-runtime:nightly-${{ env.RUNTIME_NIGHTLY_VERSION }}-py${{ matrix.python-version }}
+          build-args: ENV_FILENAME=${{ env.ENV_FILENAME }}
+
+      - name: Create Runtime ${{ env.RUNTIME_NIGHTLY_VERSION}} Python ${{ matrix.python-version }} ${{ matrix.server }} software environment
```
ncclementi, line 72:
```suggestion
      - name: Create Runtime ${{ env.RUNTIME_NIGHTLY_VERSION}}, Python ${{ matrix.python-version }}, Server ${{ matrix.server }} software environment
```

ncclementi:
For reference, I've just created a cluster in prod with `coiled=0.2.3` using a software environment created using a container, run the quickstart and things worked just fine. 
@jrbourbeau Then I would say it is safe to merge this from the compatibility point of view. 

cc: @ntabris for visibility

ncclementi:
I remove the line to run this on PR so once CI finishes I'd say this is ready.

environments/coiled-runtime-nightly-py39.yml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - coiled/label/dev
+  - dask/label/dev
+  - conda-forge
+dependencies:
+    - python=3.8
```
jrbourbeau, line 6:
```suggestion
    - python=3.9
```

.github/workflows/software-environments-nightly.yml
```diff
@@ -0,0 +1,76 @@
...
+          export PYTHON_VERSION_FORMATTED=$(echo "${{ matrix.python-version }}" | sed 's/\.//g' )
+          export COILED_SOFTWARE_ENV_NAME=coiled-runtime-$RUNTIME_NIGHTLY_VERSION_FORMATTED-py$PYTHON_VERSION_FORMATTED
+          export ENV_FILENAME=coiled-runtime-nightly-py$PYTHON_VERSION_FORMATTED
+
+          # Put variables into $GITHUB_ENV so it can be used in subsequent workflow steps
+          echo COILED_SOFTWARE_ENV_NAME=$COILED_SOFTWARE_ENV_NAME >> $GITHUB_ENV
+          echo ENV_FILENAME=$ENV_FILENAME >> $GITHUB_ENV
+          echo RUNTIME_NIGHTLY_VERSION=$RUNTIME_NIGHTLY_VERSION >> $GITHUB_ENV 
+      
+      - name: Find and Replace nightly
```
jrbourbeau, line 51: Nitpick: This name seems a bit more clear about the intent of this step 

```suggestion
      - name: Pin latest nightly coiled-runtime
```

.github/workflows/software-environments-nightly.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Checkout source
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v1
+
+      - name: Install Coiled
+        run: python -m pip install coiled
+
+      - name: Create Runtime software environment environment variables
```
jrbourbeau, line 31:
```suggestion
      - name: Create environment variables
```

.github/workflows/software-environments-nightly.yml
```diff
@@ -0,0 +1,76 @@
...
+jobs:
+  build:
+    name: Python ${{ matrix.python-version }}, nightly build, ${{ matrix.server }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10"]
+        server: ["https://staging.coiledhq.com"]
+        include: #leaving this logic as we build nightly soft env for prod.
```
jrbourbeau, line 17: I don't quite understand the meaning behind this comment -- could you elaborate a bit more?

.github/workflows/software-environments-nightly.yml
```diff
@@ -0,0 +1,76 @@
+name: Create software environments
```
jrbourbeau, line 1: We want this workflow to have a unqiue name. It currently has the same name as the `.github/workflows/software-environments-stable.yml` workflow 

```suggestion
name: Create nightly Coiled software environments
```

.github/workflows/software-environments-stable.yml
```diff
@@ -33,14 +33,33 @@ jobs:
       - name: Install Coiled
         run: python -m pip install coiled
 
-      - name: Create Runtime ${{ matrix.runtime-version }} Python ${{ matrix.python-version }} ${{ matrix.server }} software environment
-        env:
-          DASK_COILED__SERVER: ${{ matrix.server }}
-          DASK_COILED__TOKEN: ${{ secrets[matrix.token-name] }}
+      - name: Create Runtime software environment environment variables
```
jrbourbeau, line 36: Similar to comment in the other workflow

```suggestion
      - name: Create environment variables
```

.github/workflows/software-environments-nightly.yml
jrbourbeau: While we're at it, can we change the name in the other workflow to be `Create Coiled software environments` instead of `Create software environments`?

jrbourbeau (commented):
Thanks for all you work here @ncclementi. I left a few minor comments, but otherwise I think this is good to merge 


> For reference, I've just created a cluster in prod with coiled=0.2.3 using a software environment created using a container, run the quickstart and things worked just fine.

Thanks for doing that @ncclementi -- agreed that the changes here should be backwards compatible then

.github/workflows/software-environments-nightly.yml
ncclementi: Yes, I think it's not clear. But we talked about the possibility of having nightlies soft-envs in prod and this comment meant to say, leave this logic in case we want to add in the future the other server.

jrbourbeau (approved):

jrbourbeau:
@ncclementi I tried to create a cluster on prod using one of the new software environments 

```python
cluster = coiled.Cluster(n_workers=10, software="coiled/coiled-runtime-0-0-3-py39")
```

and the cluster never came up (see https://cloud.coiled.io/jrbourbeau/clusters/34443/details for more info about the cluster). Looking through one of the worker's logs, this appears to be a relevant error message: 

```
Jun 14 18:41:29 ip-10-4-6-135 cloud-init[1211]: ERROR: for worker  Cannot start service worker: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "python": executable file not found in $PATH: unknown
```

Also cc @ntabris for visibility

ncclementi:
I can reproduce this, but I didn't have this problem before when testing things. In fact I've just tried with the software env I created in dask-engineering to test this and everything worked fine. 
```
cluster = Cluster(name = "test_default", software="dask-engineering/coiled-runtime-0-0-3-py39")
```
I'm not sure what is happening here.

jrbourbeau:
When I look at that software environment, it appears to use the older-style `--conda` specification 

```
❯ coiled env inspect dask-engineering/coiled-runtime-0-0-3-py39
container:
None


conda:
{'channels': ['conda-forge'], 'dependencies': ['coiled-runtime=0.0.3', 'python=3.9']}


pip:
None


post_build:
None

```

ncclementi:
Yes, you are right, my bad (copied the wrong line). 
I meant to say this soft-env  "dask-engineering/coiled-runtime-nightly-py390-0-4a2206080023" which was a nightly build and that one worked. SO something went wrong on the build of the coiled account ones? 

```
cluster = Cluster(name = "test_default", software="dask-engineering/coiled-runtime-nightly-py390-0-4a2206080023")
```

PR #159, opened 2022-06-02
Remove `worker_memory` and use t3.large on `Cluster` creation

ncclementi:
- [x] Closes https://github.com/coiled/coiled-runtime/issues/157


jrbourbeau (approved):
Thanks @ncclementi

PR #164, opened 2022-06-08
Re-enable test_deadlock.py::test_repeated_merge_spill

fjetter:
Should be fixed by now and if not I really would like to know

### DIFF
--- a/tests/stability/test_deadlock.py
+++ b/tests/stability/test_deadlock.py
@@ -5,6 +5,7 @@ import dask
 from distributed import Client, wait
 
 
+@pytest.mark.latest_runtime
 def test_repeated_merge_spill():
     with coiled.v2.Cluster(
         name=f"test_deadlock-{uuid.uuid4().hex}",


tests/stability/test_deadlock.py
```diff
@@ -2,11 +2,9 @@
 
 import coiled.v2
 import dask
-import pytest
 from distributed import Client, wait
 
 
-@pytest.mark.skip(reason="https://github.com/dask/distributed/issues/6110")
 def test_repeated_merge_spill():
```
ncclementi, line 8: Can you mark it to run only on the latest version of dask? The tests are going to fail for the coiled runtime version 0.0.3 which has a hard pin. 
I think you can decorate it with `@pytest.mark.latest_runtime` and that should work.

fjetter: Fine with me. FWIW, for these things I think we should have a `skipif(version<version_number)` such that we run the tests for all versions above this, not just the latest

jrbourbeau:
Thanks @fjetter -- It looks like this is a duplicate of https://github.com/coiled/coiled-runtime/pull/99 so let's just go with that one instead

PR #165, opened 2022-06-08
Nightly Coiled software environments  [DNM]

### LINKED ISSUE #147: Publish nightly Coiled software environments

ncclementi:
- [ ] Closes #147
- [ ] Blocked on #154


ncclementi:
For some reason, the include / exclude logic using on-cron doesn't quite work. As it still includes py37 and runtime 0.0.3 when on-cron is true. 

I tried setting the include as:
```
- on-cron: false 
  python-version: "3.7"
  runtime-version: "0.0.3"
```
but that doesn't work, it creates the job anyways for py37 and runtime 0.0.3. Outside that the logic seems to work. 

I'm inclining to go for the two separate yaml files, one for stable releases and other for nightlies

ncclementi:
I would deprecate this in favor of additions to https://github.com/coiled/coiled-runtime/pull/154

PR #169, opened 2022-06-09
Use cp of nyc taxi data in pour ci bucket to avoid cross region

ncclementi:
To avoid cross-region data accessing costs, I copied over data needed to the S3 ci-runtime bucket. This PR has the changes to use the copied data instead.

### DIFF
--- a/tests/runtime/test_coiled.py
+++ b/tests/runtime/test_coiled.py
@@ -4,14 +4,13 @@ import pandas as pd
 
 def test_quickstart_csv(small_client):
     ddf = dd.read_csv(
-        "s3://nyc-tlc/csv_backup/yellow_tripdata_2019-*.csv",
+        "s3://coiled-runtime-ci/nyc-tlc/yellow_tripdata_2019_csv/yellow_tripdata_2019-*.csv",
         dtype={
             "payment_type": "UInt8",
             "VendorID": "UInt8",
             "passenger_count": "UInt8",
             "RatecodeID": "UInt8",
         },
-        storage_options={"anon": True},
         blocksize="16 MiB",
     ).persist()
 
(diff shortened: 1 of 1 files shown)


ncclementi:
@jrbourbeau  I think this is ready, the test failing is unrelated, and we've seen this problem before, that we should explore over https://github.com/coiled/coiled-runtime/issues/168

jrbourbeau (commented):
Thanks @ncclementi -- I think the changes here are fine. That said, I'm wondering why we don't spin up a cluster in the region where `s3://nyc-tlc/csv_backup/yellow_tripdata_2019-*.csv` lives? That would have the added benefit of checking that the dataset used in the quickstart is available and code snippet works.

jrbourbeau:
Is the quickstart with the original nyc dataset run automatically somewhere else? For example, a Coiled integration test? cc @ncclementi @shughes-uk

ncclementi:
> Thanks @ncclementi -- I think the changes here are fine. That said, I'm wondering why we don't spin up a cluster in the region where `s3://nyc-tlc/csv_backup/yellow_tripdata_2019-*.csv` lives? That would have the added benefit of checking that the dataset used in the quickstart is available and code snippet works.

Well in that case we wouldn't be able to use the fixture and we will have to create a separate cluster for that. We went back and forth multiple times and we decided to do this since it's easier to keep track of things by having the data on buckets of the same region. We did this for the parquet tests.

jrbourbeau:
Ah, I see. FWIW in this particular case I think we could still use the fixture if we add a `backend_options` marker, but probably not worth worrying about for now. Let's go with this 👍

shughes-uk:
We use taxi cab in our tests, but start our clusters in the same region

jrbourbeau:
Awesome, good to know that's covered -- thanks @shughes-uk

PR #172, opened 2022-06-10
Add cluster spinup benchmark

jrbourbeau:
Let's monitor how long it takes to spin up a Coiled cluster with default parameters. This will let us, among other things, see what impact adding new packages to the runtime has on cluster spin up time 

Closes https://github.com/coiled/coiled-runtime/issues/111

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -126,8 +126,13 @@ def s3_scratch(s3):
 
 
 @pytest.fixture(scope="function")
-def s3_url(s3, s3_scratch, request):
-    url = f"{s3_scratch}/{request.node.originalname}-{uuid.uuid4().hex}"
+def s3_url(s3, s3_scratch, test_name):
+    url = f"{s3_scratch}/{test_name}-{uuid.uuid4().hex}"
     s3.mkdirs(url, exist_ok=False)
     yield url
     s3.rm(url, recursive=True)
+
+
+@pytest.fixture
+def test_name(request):
+    return request.node.originalname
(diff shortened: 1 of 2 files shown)


conftest.py
```diff
@@ -126,8 +126,13 @@ def s3_scratch(s3):
...
+def s3_url(s3, s3_scratch, test_name):
+    url = f"{s3_scratch}/{test_name}-{uuid.uuid4().hex}"
     s3.mkdirs(url, exist_ok=False)
     yield url
     s3.rm(url, recursive=True)
+
+
+@pytest.fixture
+def test_name(request):
+    return request.node.originalname
```
ncclementi, line 138: I believe this is not the test name but the test file/module. Should we rename the fixture to match this?

jrbourbeau: Good point. What this is depends on the context in which the fixture is used. If used directly in a test, it will be the test name but, as you mentioned, if called from a module scoped fixture I think it will be module name instead. I've gone ahead and reverted this change as it's not needed to add the spinup time benchmark

ncclementi:
@jrbourbeau I was chatting earlier with @shughes-uk and it seems Coiled collects this information already, I wonder if we still want this here, or if we want to revert these changes. I guess it doesn't hurt, maybe we could eventually compare with the Coiled database data.

jrbourbeau:
Nice, I hadn't realized Coiled collected cluster spinup times. I think having this test is still useful, in particular once we have automated benchmark checks running on PRs.

shughes-uk:
I made a popsql query that plots the average spinup time of any clusters created by the `oss-benchmark-bot` that use a software environment with `default` in the name

PR #183, opened 2022-06-17
Make cluster_dump available for other clusters

### LINKED ISSUE #184: Make cluster_dumps available beyond the scope of small_client fixture
See discussion https://github.com/coiled/coiled-runtime/issues/177#issuecomment-1156716714

ncclementi:
- [x] Closes #184

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -174,7 +174,7 @@ def upload_cluster_dump(request, s3_cluster_dump_url, s3_storage_options):
         # and outside fixtures.
         try:
             yield
-        except Exception:
+        except:  # noqa: E722
             failed = True
             raise
         finally:
--- a/conftest.py
+++ b/conftest.py
@@ -151,15 +151,6 @@ def s3_cluster_dump_url(s3, s3_scratch):
     return dump_url
 
 
-# #Use this fixture to use when testing locally
-# from dask.distributed import LocalCluster
-# @pytest.fixture
-# def local_client(upload_cluster_dump):
-#     with LocalCluster() as cluster:
-#         with Client(cluster) as client:
-#             with upload_cluster_dump(client, cluster):
-#                 yield client
-
 # this code was taken from pytest docs
 # https://docs.pytest.org/en/latest/example/simple.html#making-test-result-information-available-in-fixtures
 @pytest.hookimpl(tryfirst=True, hookwrapper=True)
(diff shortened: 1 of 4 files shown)


ncclementi:
@ian-r-rose it looks like something different is happening when launching a cluster via coiled, that the plugin doesn't work as it worked locally. 

At this point I'm not sure what's going on, I would have expect "failed" to be true in this case and not try to compute `request.node.rep_call.failed` but it didin't. Any extra ideas on what could be happening? 
https://github.com/coiled/coiled-runtime/runs/6993735017?check_suite_focus=true#step:6:150

conftest.py
```diff
@@ -170,3 +149,40 @@ def s3_cluster_dump_url(s3, s3_scratch):
...
+        # the code below is a workaround to make cluster dumps work with clients in fixtures
+        # and outside fixtures.
+        try:
+            yield
+        except:  # noqa: E722
+            failed = True
+            raise
+        finally:
+            cluster_dump = strtobool(os.environ.get("CLUSTER_DUMP", "false"))
+            logger.error(f"Print failed var for testing purposes, failed=: {failed}")
```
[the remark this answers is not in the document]
ncclementi: The new logic I added works when the test passes, see https://github.com/coiled/coiled-runtime/runs/7013741536?check_suite_focus=true

I'm testing a failure now and see if this works.

conftest.py
```diff
@@ -170,3 +149,46 @@ def s3_cluster_dump_url(s3, s3_scratch):
...
+        try:
+            yield
+        except:  # noqa: E722
+            failed = True
+            raise
+        finally:
+            try:
+                # we need this for tests that are not using the client fixture
+                # for those cases request.node.rep_call.failed can't be access.
+                failed_req = request.node.rep_call.failed
```
ian-r-rose, line 184: Can this just be
```suggestion
                failed = request.node.rep_call.failed
```
?

Do we need the extra flag?

conftest.py
```diff
@@ -170,3 +149,46 @@ def s3_cluster_dump_url(s3, s3_scratch):
...
+@pytest.fixture
+def upload_cluster_dump(request, s3_cluster_dump_url, s3_storage_options):
+    @contextlib.contextmanager
+    def _upload_cluster_dump(client, cluster):
+        failed = False
+        # the code below is a workaround to make cluster dumps work with clients in fixtures
+        # and outside fixtures.
+        try:
+            yield
+        except:  # noqa: E722
```
ian-r-rose, line 177:
```suggestion
        except Exception:
```

conftest.py
```diff
@@ -170,3 +149,46 @@ def s3_cluster_dump_url(s3, s3_scratch):
...
+            failed = True
+            raise
+        finally:
+            try:
+                # we need this for tests that are not using the client fixture
+                # for those cases request.node.rep_call.failed can't be access.
+                failed_req = request.node.rep_call.failed
+            except AttributeError:
+                failed_req = False
```
ian-r-rose, line 186:
```suggestion
                failed = False
```

ian-r-rose (commented):
Thanks @ncclementi, this looks good to me, just some minor suggestions

PR #193, opened 2022-06-24
Fix cluster dump loggic when raise occurs

ncclementi:
Currently, when we go through the path in line 177 we raise, causing the last piece of code not to run. This PR fixes the problem

https://github.com/coiled/coiled-runtime/blob/45bc8e9efa852fcf84c5a78289691157f48fc15e/conftest.py#L168-L194


ncclementi:
The failure is a known flaky deadlock, being investigated. Merging this as it fixes the cluster dump issue.

PR #197, opened 2022-06-30
Add runn tests for 0.0.4 

ncclementi:
Currently, we are not the tests for 0.0.4, this PR includes these cases. This will make CI much longer, I wonder to what extent we need to keep running the tests in 0.0.3. 

Opinions? 

cc: @hayesgb and @ian-r-rose

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -114,7 +114,7 @@ jobs:
       matrix:
         os: ["ubuntu-latest"]
         python-version: ["3.9"]
-        runtime-version: ["latest", "0.0.3"]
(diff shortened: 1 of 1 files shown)


ian-r-rose:
As long as they are in the test matrix, shouldn't they be parallelized? If that's the case, I wouldn't worry too much about the extra time.

ncclementi:
You are right @ian-r-rose , I forgot that we were parallelizing.

The failure we are seeing is related to https://github.com/coiled/coiled-runtime/issues/166 if you are ok with it would you mind merging this one in?

ian-r-rose:
No objections to merging here.

ian-r-rose (approved):

PR #208, opened 2022-07-12
Collect and visualize benchmark results

### LINKED ISSUE #186: Benchmarking tool
As part of https://github.com/coiled/coiled-runtime/issues/148, we should decide on what benchmarking tool we want to use. Currently we're just having `pytest` print out test run times, but it was always intended for us to switch to using a more proper benchmarking framework at some point. [`asv`](https://asv.readthedocs.io) is the primary (only?) benchmarking tool I'm aware of that's used in our space (e.g. NumPy, pandas, scikit-image, Xarray all use `asv` to monitor performance). Because of this I'm inclined to default to using `asv` unless there is a compelling reason not to do so. 

Are folks aware of other benchmarking tools we should consider?

### LINKED ISSUE #187: Determine where historical benchmarking data should live
As part of https://github.com/coiled/coiled-runtime/issues/148, we'll need to store historical benchmark results somewhere. We should determine where those results should live. For example:

- In this repository
- In a dedicated `coiled-runtime-benchmarks` (or some other name) repository
- In S3
- ...

Related to https://github.com/coiled/coiled-runtime/issues/186

### LINKED ISSUE #188: Bring benchmarking framework online
As part of https://github.com/coiled/coiled-runtime/issues/148, after decisions have been made on https://github.com/coiled/coiled-runtime/issues/186 and https://github.com/coiled/coiled-runtime/issues/187, we should make the relevant code changes to get our benchmarking framework up and running

### LINKED ISSUE #189: Viewing benchmark results
As part of https://github.com/coiled/coiled-runtime/issues/148, it would be great to have a way to publicly display historical benchmark results. Here's an example of a webpage displaying pandas benchmark results https://pandas.pydata.org/speed/pandas

ian-r-rose:
Fixes #186, fixes #187, fixes #188, fixes #189.

This is a proof-of-concept based on the discussion in #186 using the "custom fixtures" solution, for the reasons therein.

The stack involves:

1. Metrics are collected using pytest fixtures. Currently this just does wall clock time, but I've built in fields for others and I don't see many barriers to implementing them.
2. Historical benchmarking data is stored in sqlite databases in s3. This could also be CSV, JSON, Parquet, but I found that working with sqlalchemy+sqlite handles multiple connections nicely when running with pytest-xdist, and using alembic allows for not-too-horrible migrations should we want to change the schema. Eventually, this could be replaced with some notion of repeated "jobs" in coiled itself.
3. Historical benchmarking data is currently parameterized per-platform, per-coiled-runtime, per-python-version. This is subject to change.
4. Historical data are visualized using static sites on github pages. I've used panel to generate them, but other tools are possible.
Sample: https://coiled.github.io/coiled-runtime/benchmark-ubuntu-latest-0.0.4-py3.9.html


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,214 @@
...
+    name: Benchmarks - ${{ matrix.os }}, Python ${{ matrix.python-version }}, Runtime ${{ matrix.runtime-version }}
+    needs: software
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+        python-version: ["3.9"]
+        runtime-version: ["0.0.4"]
```
ncclementi, line 88: It looks like we are not running this on the latest software environment built above? is the idea to include it later on? just making sure we don't build a soft env that we won't use : )

ncclementi (commented):
I know this is WIP but I left some comments.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,214 @@
...
+          bash ci/scripts/run_tests.sh tests/benchmarks/test_parquet.py --benchmark
+
+      - name: Upload benchmark db
+        if: >
+          always() &&
+          (steps.benchmark_tests.outcome == 'success' || steps.benchmark_tests.outcome == 'failure')
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.RUNTIME_CI_BOT_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.RUNTIME_CI_BOT_AWS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: us-east-2  # this is needed for boto for some reason
```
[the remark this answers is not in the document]
ian-r-rose: It might be, I was thinking along the same lines! Not sure what's going on there, but I also wonder whether it's related to the region thing that you ran into with the s3 pytest fixtures here

gjoseph92 (commented):
Super cool!

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,214 @@
...
+
+      - name: Set up environment
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          miniforge-variant: Mambaforge
+          use-mamba: true
+          python-version: ${{ matrix.python-version }}
+          environment-file: ci/environment.yml
+
+      - name: Build Coiled Software Environment
```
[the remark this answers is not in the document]
ian-r-rose: Yes, I'd be very happy to avoid this step! I don't have the strongest grasp on the nightly images and what their lifecycle is, so any suggestions for how to structure that are welcome

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,214 @@
...
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.RUNTIME_CI_BOT_AWS_SECRET_ACCESS_KEY }}
+          COILED_RUNTIME_VERSION: ${{ matrix.runtime-version }}
+          DB_NAME: benchmark-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}.db
+        run: |
+          bash ci/scripts/run_tests.sh tests/benchmarks/test_parquet.py --benchmark
+
+      - name: Upload benchmark db
+        if: >
+          always() &&
+          (steps.benchmark_tests.outcome == 'success' || steps.benchmark_tests.outcome == 'failure')
```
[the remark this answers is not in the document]
ian-r-rose: I think there is `canceled`, but I'm okay not pushing changes in that case

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,214 @@
...
+    branches:
+      - main
+      - custom-dashboarding
+  workflow_dispatch:
+
+concurrency:
+  # Fairly strict concurrency rule to avoid stepping on benchmark db.
+  # Could eventually replace with a real db in coiled, RDS, or litestream
+  group: benchmarks
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
ian-r-rose: I agree, I can add a note to remove before merging

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,214 @@
...
+on:
+  push:
+    branches:
+      - main
+      - custom-dashboarding
+  workflow_dispatch:
+
+concurrency:
+  # Fairly strict concurrency rule to avoid stepping on benchmark db.
+  # Could eventually replace with a real db in coiled, RDS, or litestream
```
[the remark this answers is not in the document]
ian-r-rose: I've also been looking for an excuse to use it :)

That's a really annoying limitation.

dashboard.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    Crawl the tests directory and try to grab code for each test on a best-effort
+    basis. This relies on the tests being importable from this script, so the
+    environment should be similar enough that that is possible.
+    """
+    source: dict[str, str] = {}
+    files = glob.glob("tests/**/test_*.py", recursive=True)
+    for f in files:
+        try:
+            # Fragile!
```
ncclementi, line 30: Does this mean that it sometimes works and some doesn't?

dashboard.py
```diff
@@ -0,0 +1,165 @@
...
+    originalname: str
+        The name of the test without any fixture or other modifications.
+
+    df: pandas.DataFrame
+        A dataframe with the test data in it.
+
+    spec: ChartSpec
+        Data for how to render the timeseries
+    """
+    df = df[~df[spec.field].isna() & ~df.start.isna()]
```
ncclementi, line 59: what's the purpose of this line?

tests/benchmarks/test_coiled.py
```diff
@@ -1,8 +1,14 @@
 import uuid
 
+import pytest
 from coiled import Cluster
 
 
+@pytest.fixture(autouse=True)
+def parquet_benchmark_fixture(benchmark_time):
```
ncclementi, line 8: Why do we need this here? not sure I follow

ncclementi (commented):
Added some comments, mostly for my understanding

ci/environment-dashboard.yml
```diff
@@ -0,0 +1,12 @@
+channels:
+  - conda-forge
+dependencies:
+  - pytest
+  - python
```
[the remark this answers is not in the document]
ian-r-rose: Sure, I can add some constraints here.

dashboard.py
ian-r-rose: It works today, but it might fail if our test suite includes other dependencies, has weird side effects, or otherwise makes it difficult to import. A nicer approach might be to parse it with `ast` in some way, but I didn't want to go too far down a rabbit hole, and `inspect` is fairly easy to use.

I'm trying to be fairly forgiving here (with a wide try/except), so I think the worst case scenario is that the code is missing for some (or all) functions.

dashboard.py
ian-r-rose: It's just dropping nulls for the relevant columns we are plotting. I should probably use `dropna` here.

tests/benchmarks/test_coiled.py
ian-r-rose: It automatically applies the `benchmark_time` fixture to all the tests in the module. It would be equivalent to just include it in the test itself. I suppose I was just looking ahead to when this module has multiple tests in it.

conftest.py
```diff
@@ -48,7 +62,24 @@ def pytest_collection_modifyitems(config, items):
...
 
 
-def get_software():
+def get_coiled_runtime_version():
+    if strtobool(os.environ.get("TEST_UPSTREAM", "false")):
+        return "upstream"
+    try:
+        return os.environ["COILED_RUNTIME_VERSION"]
+    except KeyError:
+        # Determine software environment from local `coiled-runtime` version (in installed)
```
jrbourbeau, line 71:
```suggestion
        # Determine software environment from local `coiled-runtime` version (if installed)
```

conftest.py
```diff
@@ -71,10 +102,137 @@ def get_software():
...
+    if not benchmark_db_session:
+        yield
+    else:
+        node = request.node
+        run = TestRun(
+            session_id=testrun_uid,
+            name=node.name,
+            originalname=node.originalname,
+            path=str(node.path.relative_to(TEST_DIR)),
+            dask_version=dask.__version__,
```
jrbourbeau, line 179: Should we also include `distributed_version`?

setup.cfg
```diff
@@ -1,5 +1,13 @@
 [flake8]
 max-line-length = 120
+exclude = alembic/versions/*
+ignore =
+    E20,   # Extra space in brackets
+    W503,  # line break before binary operator
+    W504,  # line break after binary operator
```
jrbourbeau, line 7: Why add these?

tests/benchmarks/test_coiled.py
jrbourbeau, line 8: As we want to collect timing data for all tests, should we not make this a top-level `autouse` fixture in `conftest.py`?

tests/benchmarks/test_parquet.py
```diff
@@ -25,13 +30,14 @@ def parquet_cluster():
         yield cluster
 
 
-@pytest.fixture(scope="module")
```
jrbourbeau, line 28: Nice catch. It looks like we were reusing clients between tests in this module

benchmark_schema.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Memory data
+    average_memory = Column(Float, nullable=True)
+    peak_memory = Column(Float, nullable=True)
+
+    # Timing data
+    compute_time = Column(Float, nullable=True)
+    disk_spill_time = Column(Float, nullable=True)
+    serializing_time = Column(Float, nullable=True)
+    transfer_time = Column(Float, nullable=True)
```
jrbourbeau, line 45: I could be missing something, but are we collecting this information?

.github/workflows/tests.yml
```diff
@@ -121,6 +121,10 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Maybe mark for benchmarking
+        if: always() # && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
```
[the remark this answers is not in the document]
ian-r-rose: That was my intention, yes. Do these conditions seem reasonable to you?

conftest.py
ian-r-rose: I was presuming that they would always be identical (otherwise we have bigger problems).

Do you think there is a use-case for tracking them separately? If so, now would be the time to add it.

setup.cfg
ian-r-rose: Disagreement between flake8 and `black` on some fortmatting stuff. Dask core also has these.

tests/benchmarks/test_coiled.py
ian-r-rose: Yes, we can do that -- I've been a little skittish around global `autouse` in case there was some test down the line that we *don't* want to collect benchmarking information on. But I can't think of it right now, so it's a bit theoretical.

benchmark_schema.py
ian-r-rose: No, we are not. These, as well as URLs for cluster dumps and performance reports, are not currently being collected, but I think we'll want to collect them soon, and I wanted to save us a database migration.

conftest.py
jrbourbeau: Ah, that's a good point. I can imagine a future where we decouple `dask` and `distributed` release cycles, though that could be far in the future. 

Alternatively, doesn't `coiled_runtime_version` capture both the `dask` and `distributed` versions?

benchmark_schema.py
jrbourbeau: How expensive are migrations? If they're cheap, I'd prefer to have the current database schema reflect information that's actually being gathered. Otherwise, could you add a comment here mentioning this information isn't yet collected?

conftest.py
ian-r-rose: > Alternatively, doesn't coiled_runtime_version capture both the dask and distributed versions?

Sort of -- this is used more as a unit of categorizing different runs for comparisons. That is to say, this is either a version number, or `"latest"`, or `"upstream"`. So it's a little loose, but important for determining which timeseries a particular run goes into.

tests/benchmarks/test_coiled.py
ian-r-rose: This is done

.github/workflows/tests.yml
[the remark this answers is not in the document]
jrbourbeau: I think so? Looking at the [docs for `github.ref`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) I see

> This is only set if a branch or tag is available for the event type

So I guess there are some cases where it might not work as expected (in particular I'm wondering about the cron job we have running on `main`). Regardless, I think this we should give this a try and can always update if needed

benchmark_schema.py
ian-r-rose: They are more expensive from a nuisance perspective than a disk storage or time perspective. They involve a bit of work, potentially from someone who is unfamiliar with them, and can be easy to get wrong.

I'd prefer to add a comment, since I think I'd like to add these pretty soon, if that's okay

jrbourbeau: Sounds good, I'll defer to your judgement here. Thanks for the additional context

.github/workflows/tests.yml
```diff
@@ -328,4 +364,86 @@ jobs:
...
+
+      - name: Generate dashboards
+        run: |
+          python dashboard.py
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@4.1.7
+        with:
+          branch: gh-pages
+          folder: static
```
jrbourbeau, line 449: Can we put links to these pages somewhere? For example, a badge in the README

.github/workflows/tests.yml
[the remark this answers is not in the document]
ian-r-rose: Good point. The [scheduling](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) functionality only works on the default branch, so it seems like `github.ref` should be set there. But I suppose we will find out.

.github/workflows/tests.yml
ian-r-rose: Yep! I was avoiding doing that until we settled on how the pages are grouped/categorized. But I think now is a good time to do it.

conftest.py
gjoseph92: Personally, I like capturing it even if we don't expect them to differ at all. When something is wrong, it's nice to be able to confirm that they did in fact match up.

.github/workflows/tests.yml
ian-r-rose: This is going to be a lot of badges, any objection to that?

jrbourbeau: How difficult would it be to have a single landing page with links to all the individual python / runtime version / ... combination pages?

ian-r-rose: Not too hard. Here is a look at all the combos as badges:

image

conftest.py
ian-r-rose: Okay, with n=2, I've added the distributed version

ian-r-rose:
Okay, this should be good to go

jrbourbeau (approved):
Thanks for all your work on this @ian-r-rose -- looking forward to building up a bunch of test suite runs : )

PR #216, opened 2022-07-22
Generate dashboard more insistently

ian-r-rose:
The last step of this was skipped in the scheduled run yesterday, seemingly because of an upstream job failure (even though its actual dependency was successful). See [this SO](https://stackoverflow.com/questions/65735099/github-workflow-job-not-executed-even-when-job-needs-successful) discussion.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -419,6 +419,7 @@ jobs:
 
   static-site:
     needs: process-results
+    # Always generate the site, as this can be skipped even if an indirect dependency fails (like a test run)
     if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
     name: Build static dashboards
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -419,6 +419,7 @@ jobs:
 
   static-site:
     needs: process-results
+    if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
```
ncclementi, line 422: should we add a comment and why do we do this? 
Thinking something like this, not sure if the syntax of the comment is correct here. 
```suggestion
    # run always even when dependant workflow failed. 
    if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
```

ian-r-rose: Sure, good idea.

ian-r-rose, line 422:
```suggestion
    # Always generate the site, as this can be skipped even if an indirect dependency fails (like a test run)
    if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
```

ncclementi (approved):

PR #219, opened 2022-07-29
Fix tooltip rendering

### LINKED ISSUE #218: Tooltip always show duration even in memory plots [dashboard]
Showing durations on tooltip when in memory tab: 

Screen Shot 2022-07-28 at 7 59 29 PM

I'm pretty sure this comes from how the tooltip is created, see here https://github.com/coiled/coiled-runtime/blob/fb3a04f91581c2582c520b30006f63cd0055f156/dashboard.py#L80 . I think that line should be something like 

```python
altair.Tooltip(f"{spec.field}:Q", title=spec.label), 
```
I tried this change and the only caveat is that instead of saying durations now it'll say Wall Clock, for the Wall clock part.  

with the proposed change : 

Screen Shot 2022-07-28 at 8 14 14 PM
Screen Shot 2022-07-28 at 8 15 32 PM

cc: @ian-r-rose

ncclementi:
- [ ] Closes #218

Example of rendering in the issue.

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -77,7 +77,7 @@ def make_timeseries(originalname, df, spec) -> altair.Chart | None:
                 altair.Tooltip("call_outcome:N", title="Test Outcome"),
                 altair.Tooltip("coiled_runtime_version:N", title="Coiled Runtime"),
                 altair.Tooltip("dask_version:N", title="Dask"),
-                altair.Tooltip("duration:Q", title="Duration"),
+                altair.Tooltip(f"{spec.field}:Q", title=spec.label),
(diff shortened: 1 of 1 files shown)


ian-r-rose (approved):
Thanks @ncclementi! We may want to put more data into the tooltips later, but this is a nice fix for now

ncclementi:
@ian-r-rose I see we have the KeyError provblem, we've seen flaky in CI but I just saw this one too that I haven't seen. Do you know what could this be? https://github.com/coiled/coiled-runtime/runs/7570980076?check_suite_focus=true#step:7:188

ncclementi:
Merging as the CI failures are unrelated and reported.

PR #226, opened 2022-08-04
Add regression detection [DNM yet]

### LINKED ISSUE #190: Performance regression visibility
Today each CI run prints the time it took each test to run. Technically, this contains the information we need to identify a performance regression, but in practice such a regression is likely to go unnoticed as we don't regularly check the times printed in CI. As part of https://github.com/coiled/coiled-runtime/issues/148, it would be great to have CI automatically fail if there was a significant performance regression detected in our benchmarks.

ncclementi:
- [x] Closes #190

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -337,7 +337,7 @@ jobs:
   process-results:
     needs: [runtime, benchmarks, stability]
     name: Combine separate benchmark results
-    if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+    if: always() && github.repository == 'coiled/coiled-runtime'
     runs-on: ubuntu-latest
     concurrency:
       # Fairly strict concurrency rule to avoid stepping on benchmark db.
@@ -371,6 +371,7 @@ jobs:
           bash ci/scripts/combine-dbs.sh
 
       - name: Upload benchmark db
+        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.RUNTIME_CI_BOT_AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.RUNTIME_CI_BOT_AWS_SECRET_ACCESS_KEY }}
@@ -388,7 +389,7 @@ jobs:
   regressions:
     needs: process-results
     # Always check for regressions, as this can be skipped even if an indirect dependency fails (like a test run)
-    if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+    if: always()
     name: Detect regressions
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -370,14 +370,14 @@ jobs:
           ls -lhR benchmarks
           bash ci/scripts/combine-dbs.sh
 
-      # - name: Upload benchmark db
-      #   env:
-      #     AWS_ACCESS_KEY_ID: ${{ secrets.RUNTIME_CI_BOT_AWS_ACCESS_KEY_ID }}
-      #     AWS_SECRET_ACCESS_KEY: ${{ secrets.RUNTIME_CI_BOT_AWS_SECRET_ACCESS_KEY }}
-      #     AWS_DEFAULT_REGION: us-east-2  # this is needed for boto for some reason
-      #     DB_NAME: benchmark.db
-      #   run: |
-      #     aws s3 cp $DB_NAME s3://coiled-runtime-ci/benchmarks/
+      - name: Upload benchmark db
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.RUNTIME_CI_BOT_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.RUNTIME_CI_BOT_AWS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: us-east-2  # this is needed for boto for some reason
+          DB_NAME: benchmark.db
+        run: |
+          aws s3 cp $DB_NAME s3://coiled-runtime-ci/benchmarks/
 
       - name: Upload benchmark results as artifact
         uses: actions/upload-artifact@v3
@@ -413,7 +413,7 @@ jobs:
           python detect_regressions.py
       
       - name: Create regressions summary
-        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+        if: always()
         run: |
           echo "$(<regressions_summary.md)" >> $GITHUB_STEP_SUMMARY
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -405,6 +405,8 @@ jobs:
           environment-file: ci/environment-regressions.yml
 
       - name: Run detect regressions
+        env:
+          ANSIBLE_FORCE_COLOR: '1'
         run: |
           python detect_regressions.py
       
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -383,7 +383,7 @@ jobs:
 
   regressions:
     needs: process-results
-    # Always generate the site, as this can be skipped even if an indirect dependency fails (like a test run)
+    # Always check for regressions, as this can be skipped even if an indirect dependency fails (like a test run)
     if: always() && github.event_name == 'pull_request' && github.repository == 'coiled/coiled-runtime' #always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
     name: Detect regressions
     runs-on: ubuntu-latest
@@ -402,7 +402,7 @@ jobs:
           miniforge-variant: Mambaforge
           use-mamba: true
           python-version: "3.9"
-          environment-file: ci/environment-regressions.yml
+          environment-file: ci/environment-dashboard.yml
 
       - name: Run detect regressions
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -156,7 +156,7 @@ jobs:
 
       - name: Upload benchmark results
         uses: actions/upload-artifact@v3
-        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+        if: always() && github.event_name == 'pull_request' && github.repository == 'coiled/coiled-runtime' #always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
         with:
           name: runtime-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}
           path: runtime-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}.db
@@ -213,7 +213,7 @@ jobs:
 
       - name: Upload benchmark results
         uses: actions/upload-artifact@v3
-        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+        if: always() && github.event_name == 'pull_request' && github.repository == 'coiled/coiled-runtime' #always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
         with:
           name: benchmark-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}
           path: benchmark-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}.db
@@ -292,7 +292,7 @@ jobs:
 
       - name: Upload benchmark results
         uses: actions/upload-artifact@v3
-        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+        if: always() && github.event_name == 'pull_request' && github.repository == 'coiled/coiled-runtime' #always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
         with:
           name: stability-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}
           path: stability-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}.db
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -123,7 +123,7 @@ jobs:
           fetch-depth: 0
 
       - name: Maybe mark for benchmarking
-        if: always() && github.event_name == 'pull_request' && github.repository == 'coiled/coiled-runtime' #always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
         run: echo BENCHMARK=true >> $GITHUB_ENV
 
       - name: Set up environment
@@ -158,7 +158,7 @@ jobs:
 
       - name: Upload benchmark results
         uses: actions/upload-artifact@v3
-        if: always() && github.event_name == 'pull_request' && github.repository == 'coiled/coiled-runtime' #always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
         with:
           name: runtime-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}
           path: runtime-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}.db
@@ -181,7 +181,7 @@ jobs:
           fetch-depth: 0
 
       - name: Maybe mark for benchmarking
-        if: always() && github.event_name == 'pull_request' && github.repository == 'coiled/coiled-runtime' #always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
         run: echo BENCHMARK=true >> $GITHUB_ENV
 
       - name: Set up environment
@@ -216,7 +216,7 @@ jobs:
 
       - name: Upload benchmark results
         uses: actions/upload-artifact@v3
-        if: always() && github.event_name == 'pull_request' && github.repository == 'coiled/coiled-runtime' #always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
         with:
           name: benchmark-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}
           path: benchmark-${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}.db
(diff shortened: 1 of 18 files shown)


.github/workflows/tests.yml
```diff
@@ -402,21 +366,87 @@ jobs:
...
-        run: |
-          aws s3 cp $DB_NAME s3://coiled-runtime-ci/benchmarks/
+      # - name: Upload benchmark db
+      #   env:
+      #     AWS_ACCESS_KEY_ID: ${{ secrets.RUNTIME_CI_BOT_AWS_ACCESS_KEY_ID }}
+      #     AWS_SECRET_ACCESS_KEY: ${{ secrets.RUNTIME_CI_BOT_AWS_SECRET_ACCESS_KEY }}
+      #     AWS_DEFAULT_REGION: us-east-2  # this is needed for boto for some reason
+      #     DB_NAME: benchmark.db
+      #   run: |
+      #     aws s3 cp $DB_NAME s3://coiled-runtime-ci/benchmarks/
```
ian-r-rose, line 376: Thank you for being careful here :)

ncclementi:
Raising an exception with a summary of what's failing seems to be working, see https://github.com/coiled/coiled-runtime/runs/7681316108?check_suite_focus=true. 

But I've just realized that I didn't use the latest run to compare because the individual databases only get updated on main, so I modify that to test it. 

Separate comment: Currently the threshold is (mean + 1 std) where mean is computed with the latest 10 runs excluding the one being evaluated for regression. That seems to be tight, although in some cases we will be missing regression problems due to averaging, in cases like test_download_throughput[s3fs] in https://coiled.github.io/coiled-runtime/coiled-upstream-py3.9.html

tests/test_detect_regressions.py
```diff
@@ -0,0 +1,72 @@
...
+
+        test_names = list(by_test.groups.keys())
+        for name in test_names:
+            df_test = by_test.get_group(name)
+
+            df_passed = df_test[df_test.call_outcome == "passed"]
+
+            # stats of latest 10 runs exclude last point
+
+            stats_dict[f"({runtime, name})"] = {
```
ncclementi, line 40: I created a dictionary here because initially, I thought I might want to have all this info and dump it into a dataframe or something and we could do a table or something else for the dashboard. But I'm not sure it is the cleanest approach. Thoughts?

ncclementi:
I'm having problems with getting the last value of the timeseries being the one of the run that just happened. It seems the database is not getting updated because I'm using the database artifact after the single DBs were merged and I keep getting the incorrect last value. 

- I put a sleep on benchamark/test_coiled.py::test_default_cluster_spinup_time and that should have triggered the exception and it didn't. 
- If I look at one of the failed examples, I see
```python-traceback
E               runtime= 'coiled-0.0.3-py3.7', name= 'test_shuffle_parquet', duration_last = 226.18878746032715, dur_threshold= 223.97877921614503 
```
and if I look at the last uploaded value of that test I see that is the value it's picking up. 
Screen Shot 2022-08-04 at 6 55 21 PM

I'm not sure what I'm doing wrong here, and why the is not locally updating after the combine step. @ian-r-rose do you have any idea what could be happening here?

detect_regressions.py
```diff
@@ -0,0 +1,121 @@
...
+                        f"peak_mem_last = "
+                        f"{(stats['peak_memory_last'], stats['peak_memory_last-1'], stats['peak_memory_last-2'])}, "
+                        f"{peak_mem_threshold= } \n"
+                    )
+
+                    regressions.append(reg)
+
+    # convert dict to dataframe to write to xml to use later ask ian if this is what he had in mind.
+    # df_stats = pandas.DataFrame.from_dict(stats_dict, orient="index")
+    # df_stats.to_xml("stats.xml")
```
ian-r-rose, line 110: This is what I had in mind: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary

Looks like it's unstructured markdown, not XML

ncclementi: So it seems that we can do `df.to_markdown()`. Now after seeing the example in the I assume that what you want in the summary is a summary of the regressions as a table, not all the stats for every test. 

Having a summary of the regressions as a nice markdown I think is possible. I'm thinking something like every row is a test_name, and as columns we have `[runtime_ver, regression_type, mean, last_val, last_val-1, last_val-2, thershold]` . 

Would something like this work for a summary?

ncclementi:
@ian-r-rose I know you mentioned that you preferred having this as a script instead of a test. I noticed that by running this as a script and not a test, we don't get the red on the CI run. 

If the reason behind having a script is that we want to be able to run this again maybe when creating the static page and doing a nice table there with all the stats we could just save the dict with all the data, upload it as an artifact and use it during the statics step.  Thoughts?

detect_regressions.py
ian-r-rose: Sure, I don't have a strong idea of the best presentation, so if you come up with something that you think is useful, that would be a great starting point.

ian-r-rose:
> @ian-r-rose I know you mentioned that you preferred having this as a script instead of a test. I noticed that by running this as a script and not a test, we don't get the red on the CI run.

I'm not sure I follow -- I see a red X on your regressions test  [here](https://github.com/coiled/coiled-runtime/actions/runs/2806316021). It should be a failure if the last command of the script has a non-zero exit code (which your `assert` accomplishes)

> If the reason behind having a script is that we want to be able to run this again maybe when creating the static page and doing a nice table there with all the stats we could just save the dict with all the data, upload it as an artifact and use it during the statics step. Thoughts?

I'm not against an artifact, but it does seem a little over-complicated to me, since it's not expensive to re-compute these values. Maybe if we start wanting to do a monte-carlo Bayesian changepoint analysis or something...

ncclementi:
> I'm not sure I follow -- I see a red X on your regressions test [here](https://github.com/coiled/coiled-runtime/actions/runs/2806316021). It should be a failure if the last command of the script has a non-zero exit code (which your `assert` accomplishes)

Yes, I meant to say the traceback wasn't red like [here](https://github.com/coiled/coiled-runtime/runs/7682057527?check_suite_focus=true) for example. But I guess it's ok.

> I'm not against an artifact, but it does seem a little over-complicated to me, since it's not expensive to re-compute these values. Maybe if we start wanting to do a monte-carlo Bayesian changepoint analysis or something...

👍

ian-r-rose:
> Yes, I meant to say the traceback wasn't red like [here](https://github.com/coiled/coiled-runtime/runs/7682057527?check_suite_focus=true) for example. But I guess it's ok.

Oh, interesting. I'm not sure how to accomplish that! I'll poke around google

ian-r-rose:
Looks like GitHub actions supports ANSI colors: https://github.com/ian-r-rose/gha-test/runs/7700189190?check_suite_focus=true

You could try to use something like rich to format nice text output if you wanted to (with the `force_terminal` option).

.github/workflows/tests.yml
```diff
@@ -402,21 +366,91 @@ jobs:
...
+          miniforge-variant: Mambaforge
+          use-mamba: true
+          python-version: "3.9"
+          environment-file: ci/environment-regressions.yml
+
+      - name: Run detect regressions
+        run: |
+          python detect_regressions.py
+      
+      - name: Create regressions summary
```
ncclementi, line 411: It seems that this step doesn't run when the previous one failed I might have to use create and artifact and create a separate workflow, not sure yet.

ian-r-rose: You can add an `if: always()` or similar condition to get around that.

ncclementi:
I was able to get some extra color on the regression report, locally I see all the regressions in red, but in CI I don't. 
https://github.com/coiled/coiled-runtime/runs/7734995707?check_suite_focus=true

I was able to get the markdown table https://github.com/coiled/coiled-runtime/actions/runs/2820970035/attempts/1#summary-7734995707

ncclementi:
After a conversation with @ian-r-rose : 
- Split the script into two separate functions so we can reuse the `detect_regressions()` as a separate script.
- Add a check to only compute stats when having more than 6 values. I'd like this number to be bigger but since we don't have enough entries I'm using 6 for now. 
- Add a check to only compute regressions stats if the tests are not obsolete (there is a run in the latest 7 days)

I think this is ready for a thorough review, so far I haven't reverted changes that allow things to run on this PR. That would be the latest step. 

cc: @jrbourbeau in case you want to take a look too.

.github/workflows/tests.yml
```diff
@@ -402,21 +366,92 @@ jobs:
...
 
       - name: Upload benchmark results as artifact
         uses: actions/upload-artifact@v3
         with:
           name: benchmark.db
           path: benchmark.db
 
+  regressions:
+    needs: process-results
+    # Always generate the site, as this can be skipped even if an indirect dependency fails (like a test run)
```
ian-r-rose, line 386:
```suggestion
    # Always check for regressions, as this can be skipped even if an indirect dependency fails (like a test run)
```

ci/environment-regressions.yml
```diff
@@ -0,0 +1,8 @@
+channels:
```
ian-r-rose, line 1: Maybe this is worth combining with the static site env? Especially if we want to start importing some of this functionality when generating summaries.

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+
+import pandas
+import sqlalchemy
+
+
+def detect_regressions(database_file):
+
+    engine = sqlalchemy.create_engine(f"sqlite:///{database_file}")
+
+    df = pandas.read_sql("select * from test_run where platform = 'linux'", engine)
```
ian-r-rose, line 12: Since you also filter by `"passed"` below, we might as well do it here:
```suggestion
    df = pandas.read_sql("select * from test_run where platform = 'linux' and call_outcome = 'passed'", engine)
```

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+        runtime=(
+            "coiled-"
+            + df.coiled_runtime_version
+            + "-py"
+            + df.python_version.str.split(".", n=2).str[:2].str.join(".")
+        ),
+        category=df.path.str.split("/", n=1).str[0],
+    )
+
+    # by_name_runtime = df.groupby(["name", "runtime"])
```
ian-r-rose, line 25: Stray comment

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+
+        test_names = list(by_test.groups.keys())
+        for name in test_names:
+            df_test = by_test.get_group(name)
+
+            df_passed = df_test[df_test.call_outcome == "passed"]
+            # check for empty dataframe and having enough data to do some stats
+            # currently we don't have a lot of data but eventually use 10 instead of 6.
+
+            if not df_passed.empty and (len(df_passed) >= 6):
```
ian-r-rose, line 44: If it has length greater than six, it's also not empty, so I think the first check is redundant.

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+
+                # check the test is not obsolete.
+                date_last_run = datetime.datetime.strptime(
+                    df_passed.start.values[-1].split()[0], "%Y-%m-%d"
+                )
+                date_threshold = datetime.datetime.today() - datetime.timedelta(7)
+
+                if date_last_run < date_threshold:
+                    # the latest run was 7+ days ago, test is obsolete
+                    pass
```
ian-r-rose, line 54: We could probably work this into the SQL query as well, though it would take a bit more work. In the meantime, I think it would be a bit cleaner to use the pandas util:
```python
if pd.to_timestamp(df_passed.start.iloc[-1]) < pd.Timestamp.now() - pd.Timedelta(days=7):
    pass
```

(I'm not sure why pandas is having a tough time with the datetime column in the db)

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+                            "peak_memory",
+                            stats["peak_memory_mean"],
+                            stats["peak_memory_last"],
+                            stats["peak_memory_last-1"],
+                            stats["peak_memory_last-2"],
+                            peak_mem_threshold,
+                        ]
+            else:
+                # df_pass is empty or we don't have enough stats
+                pass
```
ian-r-rose, line 160: It's equivalent to just remove these lines

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+                date_last_run = datetime.datetime.strptime(
+                    df_passed.start.values[-1].split()[0], "%Y-%m-%d"
+                )
+                date_threshold = datetime.datetime.today() - datetime.timedelta(7)
+
+                if date_last_run < date_threshold:
+                    # the latest run was 7+ days ago, test is obsolete
+                    pass
+                else:
+                    # get category for report
```
ian-r-rose, line 56: There is a lot of duplication in this block: basically identical code for duration, peak memory, and average memory. Since we plan upon adding more metrics than just these, I think it's worth doing the legwork to make this agnostic to which field we are working with. So instead we could have something like

```python
for metric in ["duration", "peak_memory", "average_memory"]:
    ...  # check for regressions
```

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+                    peak_mem_threshold = (
+                        stats["peak_memory_mean"] + stats["peak_memory_mean"]
+                    )
+
+                    # Only raise if the last three show regression
+                    if (
+                        stats["duration_last"] >= dur_threshold
+                        and stats["duration_last-1"] >= dur_threshold
+                        and stats["duration_last-2"] >= dur_threshold
+                    ):
```
ian-r-rose, line 91: A bit more legible (to me at least):

```python
if (df_passed.duration.iloc[-3:] >= dur_threshold).all():
   ...
```

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+                # df_pass is empty or we don't have enough stats
+                pass
+
+    return reg_df, regressions
+
+
+def regressions_report(reg_df, regressions):
+
+    # write reg_df to markdown for GHA summary
+    reg_df.to_markdown("regressions_summary.md")
```
ian-r-rose, line 168: It should also be possible to write directly to `$GITHUB_STEP_SUMMARY` here and avoid the intermediate file, though it's fine as is.

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+
+    if regressions:
+        raise Exception(
+            f"\x1b[31m Regressions detected {len(regressions)}: \n{''.join(regressions)} \x1b[0m"
+        )
+    else:
+        assert not regressions
```
ian-r-rose, line 175: This should never happen, since you've already checked `if regressions` above.

detect_regressions.py
```diff
@@ -0,0 +1,184 @@
...
+                date_threshold = datetime.datetime.today() - datetime.timedelta(7)
+
+                if date_last_run < date_threshold:
+                    # the latest run was 7+ days ago, test is obsolete
+                    pass
+                else:
+                    # get category for report
+                    category = df_passed.category.unique()[0]
+                    # stats of latest 10 runs exclude last point
+                    stats_dict[f"({runtime, name})"] = {
```
ian-r-rose, line 59: I'm not sure if I see a ton of value to creating a stats dict which holds most of the same information as the dataframe. Why not just operate on the dataframe directly: slicing, computing mean, std, etc should all be pretty cheap

ian-r-rose (commented):
Thanks @ncclementi!

detect_regressions.py
ncclementi: I'm not sure I follow how we can write directly to the summary without doing a `to_markdown` step.

ian-r-rose: I'm not proposing skipping the `to_markdown` step, just that we could probably directly write to `$GITHUB_STEP_SUMMARY` here. Something like

```python
with open(os.environ.get("GITHUB_STEP_SUMMARY", "regressions_summary.md") mode="a") as f:
    reg_df.to_markdown(f)
```
then drop the additional step.

Doesn't really matter too much, though

ncclementi: Oh I see, yes I think I like it in the action now because it makes it more clear what's happening, at least to me. But we could re-evaluate in the future

detect_regressions.py
```diff
@@ -0,0 +1,115 @@
...
+                                df_test[metric].iloc[-3] * units_norm,
+                            )
+                            reg = (
+                                f"{runtime = }, {name = }, {category = }, "
+                                f"last_three_{metric} {u} = "
+                                f"{last_three}, "
+                                f"{metric}_threshold {u} = {metric_threshold * units_norm} \n"
+                            )
+
+                            regressions.append(reg)
```
ian-r-rose, line 81: These summary strings could also be a column in `reg_df`, might make it slightly easier to pass the dataframe around as we incorporate this into the static site.

ian-r-rose (approved):
Thanks @ncclementi! I'll take another look tomorrow when I'm fresher, but this looks like it's ready to start pestering us

ncclementi:
Just noticed that this passed but not really working, it's returning an empty df and regressions string, but it shouldn't be the case. Trying to fix this locally.

detect_regressions.py
```diff
@@ -52,7 +60,7 @@ def detect_regressions(database_file):
                         category = df_test.category.unique()[0]
                         metric_threshold = (
                             df_test[metric][-13:-3].mean()
-                            + 2 * df_test[metric][-13:-3].std()
+                            + 1 * df_test[metric][-13:-3].std()
```
ian-r-rose, line 63: I'd still advocate for 2 standard deviations: a ~40% false positive rate would get annoying pretty quickly

ncclementi: Yes I agree, I was just trying to get regressions on purpose to see things work, I'll revert this back soon.

ncclementi:
Things work as expected see https://github.com/coiled/coiled-runtime/runs/7810509623?check_suite_focus=true and summary https://github.com/coiled/coiled-runtime/actions/runs/2847981410#summary-7810509623

I'll move to 2 standard deviations as 1 is to strict, and I'll start removing things that were only to test on this PR.

.github/workflows/tests.yml
ncclementi: @ian-r-rose is it safe to uncomment this now? Unless there are any other things you want to check?

ian-r-rose: I think we're ready

.github/workflows/tests.yml
```diff
@@ -406,21 +370,92 @@ jobs:
...
+          use-mamba: true
+          python-version: "3.9"
+          environment-file: ci/environment-dashboard.yml
+
+      - name: Run detect regressions
+        run: |
+          python detect_regressions.py
+      
+      - name: Create regressions summary
+        if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
```
ian-r-rose, line 416: I don't think we need the `ref` or `repository` check here

ncclementi: do we want to run this on PRs too? I thought since we are only updating the database on ref = main, we should keep this. Or will this be cover since the check it's also at the step level.

ian-r-rose: Isn't this job gated by the same check above? I just mean that it's redundant.

That being said, I actually think that running this on PRs would be good practice. We mostly just don't want to upload the new db on PRs

.github/workflows/tests.yml
```diff
@@ -334,42 +334,6 @@ jobs:
...
-            github.issues.create({
-                owner: variables.owner,
-                repo: variables.name,
-                body: issue_body,
-                title: title,
-            })
-
   process-results:
     needs: [runtime, benchmarks, stability]
     name: Combine separate benchmark results
```
ian-r-rose, line 339: So, if we wanted to run the regression check on PRs, I think we'd have to remove the `github.ref` check on `process-results`, but then add it to the "Upload benchmark db" step. Then `regressions` could just `always()` run.

ian-r-rose:
This looks great! Let's watch CI to make sure things behave as we expect, then merge!

ncclementi:
@ian-r-rose thanks for pushing that fix, it looks like we don't have regressions as of this PR. : )

ian-r-rose:
Let's take it for a spin!

PR #229, opened 2022-08-05
Integration tests for spilling

### LINKED ISSUE #136: Integration tests: spill/unspill
### Persist a lot of data on disk

What happens when we load 10x more data than we have RAM?  

- Does Coiled have enough storage by default?  How do we modify this?
- How are we doing against theoretical performance?  Is Dask's spill-to-disk efficient?

### Pseudocode

```python
x = da.random.random(...).persist()  # load the data 
wait(x)
x.sum().compute()  # force us to read the data again
```

[EDIT by @crusaderky ]
In addition to the trivial use case above, we would also like to have 
https://github.com/dask/distributed/blob/f7f650154fea29978906c65dd0225415da56ed11/distributed/tests/test_active_memory_manager.py#L1079-L1085

scaled up to production size. This will stress the important use case of spilled tasks that are taken out of the spill file and back into memory not to be computed, but to the transferred to another worker.
This stress test should find a sizing that is spilling/unspilling heavily but is still completing successfully.
Related:
- #135
- #140

hendrikmakait:
Closes #136

### DIFF
--- a/tests/stability/test_spill.py
+++ b/tests/stability/test_spill.py
@@ -15,7 +15,12 @@ def test_spilling(keep_around):
         worker_disk_size=55,
         worker_vm_types=["t3.medium"],
         wait_for_workers=True,
-        environ={"DASK_DISTRIBUTED__SCHEDULER__ALLOWED_FAILURES": "0"},
+        environ={
+            # Note: We set allowed-failures to ensure that no tasks are not retried
+            #  upon ungraceful shutdown behavior during adaptive scaling
+            #  but we receive a KilledWorker() instead.
+            "DASK_DISTRIBUTED__SCHEDULER__ALLOWED_FAILURES": "0"
+        },
     ) as cluster:
         with Client(cluster) as client:
             arr = da.random.random((200, 2**27)).persist()  # 200 GiB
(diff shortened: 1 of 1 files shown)


tests/stability/test_spill.py
```diff
@@ -0,0 +1,48 @@
...
+    with Cluster(
+        name=f"test_spill-{uuid.uuid4().hex}",
+        n_workers=5,
+        worker_disk_size=45,
+        worker_vm_types=["t3.medium"],
+        wait_for_workers=True,
+        environ={
+            "DASK_DISTRIBUTED__SCHEDULER__ALLOWED_FAILURES": "0",
+            "DASK_DISTRIBUTED__WORKER__CONNECTIONS__INCOMING": "1",
+            "DASK_DISTRIBUTED__WORKER__CONNECTIONS__OUTGOING": "1",
```
hendrikmakait, line 40: We need to reduce the number of incoming and outgoing connections to avoid getting the worker `oom-killed`. With a default chunk size of 100 MiB from the `dask.array`, having 4 incoming and 4 outgoing connections at the same time means that we might use up to 800 MiB for communications. With the (previous) default of a `t3.medium` machine, this appears to be ~25 % of the actual memory that we can use. Together with the misconfigured `Worker.memory_limit` (coiled/feedback#185), we reliably trigger the `oom-killer` when running the workload without the adjusted config. See https://github.com/dask/distributed/issues/6208 for an existing upstream issue.

fjetter (approved):

ian-r-rose:
Thanks @hendrikmakait!

tests/stability/test_spill.py
ian-r-rose: @hendrikmakait is this safe to remove now that coiled/feedback#185 is resolved?

hendrikmakait: No, there is an ongoing effort in dask/distributed#6208 to resolve the underlying issue. coiled/feedback#185 made this even worse.

ian-r-rose: Thanks. I think we may need to skip this test for a while until this is a bit more stable (cf #280 )

PR #232, opened 2022-08-09
Don't require AWS env vars to be set

gjoseph92:
There are other ways to get AWS credentials than environment vars, which boto will handle fine. This is annoying when running tests locally.


ian-r-rose (approved):
Thanks @gjoseph92

jrbourbeau (approved):
Thanks @gjoseph92! Let's merge after CI finishes

PR #234, opened 2022-08-09
Make CI more robust to transient conda setup failures

jrbourbeau:
We've been observing some CI failures in `main` due to conda download timeouts. This PR increases related timeouts and retries (similar to what we're doing in `distributed`) to hopefully be more robust to these types of failures


ian-r-rose (approved):
Thanks @jrbourbeau

PR #237, opened 2022-08-10
Remove strict priority

ian-r-rose:
Attempt at fixing #236.  Strict priority seemingly doesn't work for direct URLs to nightly conda builds


jrbourbeau (approved):
Nice -- thanks for the quick fix @ian-r-rose. I should have tested the upstream build over in https://github.com/coiled/coiled-runtime/pull/234.

PR #239, opened 2022-08-10
More specific time benchmarking

### LINKED ISSUE #238: `benchmark_time` includes cluster startup time
The `benchmark_time` fixture simply times how long it takes the entire test to run:
https://github.com/coiled/coiled-runtime/blob/54c49ccac9546188acc5bc1099ab1b31e3ccae1a/conftest.py#L206-L212

However, that doesn't distinguish between cluster startup time (and any test initialization time) and actual runtime of the core workload.
1. Cluster startup is somewhat availability-dependent and we should expect to be a little variable
2. Improvements/regressions to cluster startup time might look like improvements/regressions in dask performance, and visa versa

Both are valuable metrics to track, but they shouldn't be combined into one.

cc @ian-r-rose @jrbourbeau

ian-r-rose:
Fixes #238 

Create a `benchmark_time` contextmanager fixture for more granular benchmarking of tests when desired. The old behavior is available through `auto_benchmark_time`. The fixture is no longer autoused throughout the whole test suite to allow for custom usage, though I have applied it to `small_client` and `parquet_client` so those should continue working.

### DIFF
--- a/tests/benchmarks/test_parquet.py
+++ b/tests/benchmarks/test_parquet.py
@@ -26,12 +26,12 @@ def parquet_cluster():
 
 
 @pytest.fixture
-def parquet_client(parquet_cluster, sample_memory):
+def parquet_client(parquet_cluster, sample_memory, benchmark_time):
     with distributed.Client(parquet_cluster) as client:
         parquet_cluster.scale(N_WORKERS)
         client.wait_for_workers(N_WORKERS)
         client.restart()
-        with sample_memory(client):
+        with sample_memory(client), benchmark_time:
             yield client
 
 
--- a/conftest.py
+++ b/conftest.py
@@ -201,17 +201,27 @@ def test_run_benchmark(benchmark_db_session, request, testrun_uid):
         benchmark_db_session.commit()
 
 
-@pytest.fixture(scope="function", autouse=True)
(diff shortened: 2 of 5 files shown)


tests/benchmarks/test_parquet.py
```diff
@@ -26,12 +26,12 @@ def parquet_cluster():
 
 
 @pytest.fixture
-def parquet_client(parquet_cluster, sample_memory):
+def parquet_client(parquet_cluster, sample_memory, benchmark_time):
     with distributed.Client(parquet_cluster) as client:
         parquet_cluster.scale(N_WORKERS)
         client.wait_for_workers(N_WORKERS)
         client.restart()
```
gjoseph92, line 33: Yeah, this is a great example of things we don't want to include in the benchmark measurement!

gjoseph92 (approved):
Thanks @ian-r-rose! This is great. And as a bonus, I expect to see all of the benchmarks get a bit faster now (that they're not measuring `wait_for_workers` and `restart` 😁).

ian-r-rose:
Yes, waiting on a restart is the big one, it’s often fast, but sometimes
takes quite a long time. I expect some of the weird outliers we’ve seen to
happen less often (eg in the download throughput benchmark)

PR #240, opened 2022-08-11
Try removing explicit region to see what breaks.

ian-r-rose:


ian-r-rose:
Everything seems in order, I couldn't find any problems with removing the region

ian-r-rose:
Failure to build a software environment -- haven't seen one of those in a while, but wouldn't happen with in a `package_sync` world (we'd have fun new errors).

ncclementi (approved):

ncclementi:
> Failure to build a software environment -- haven't seen one of those in a while, but wouldn't happen with a `package_sync` world (we'd have fun new errors).

I'm not sure I follow this, I bet this was supposed to go on a different PR CI?

ian-r-rose:
> I'm not sure I follow this, I bet this was supposed to go on a different PR CI?

There was a previous failure, I re-ran it.

PR #242, opened 2022-08-12
Add `ci-failure` label to upstream report

jrbourbeau:
This should make sure our `ci-failure` label is automatically applied when an upstream test failure issue is opened

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -432,19 +432,14 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            const title = "⚠️ CI failed ⚠️"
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -432,19 +432,14 @@ jobs:
...
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            const title = "⚠️ CI failed ⚠️"
             const workflow_url = `https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}`
             const issue_body = `[Workflow Run URL](${workflow_url})`
-            const variables = {
-                owner: context.repo.owner,
-                name: context.repo.repo,
-                creator: "github-actions[bot]"
-            }
```
jrbourbeau, line 442: This is an unrelated simplification. It looks like we can just inline these values directly below.

ian-r-rose (approved):
Thanks @jrbourbeau

PR #243, opened 2022-08-12
Benchmarks for some common workloads

gjoseph92:
Closes https://github.com/coiled/coiled-runtime/issues/191

This adds all benchmarks from https://github.com/dask/distributed/issues/6560#issuecomment-1165215417.

They have been slightly rewritten, so the tests pick their data size automatically as a factor of total cluster memory. This is intended to make https://github.com/coiled/coiled-runtime/issues/241 significantly easier: by just parametrizing over different clusters, we can test both [strong and weak scaling](https://en.wikipedia.org/wiki/Scalability#Weak_versus_strong_scaling) of the same tests.

This also adds a couple of utility functions, `scaled_array_shape` and `timeseries_of_size`, to help writing dynamically-scalable tests like this. I found that utilities like this make translating real-world examples into benchmark tests significantly more pleasant: you just look at `real_world_data_size / real_world_cluster_size`, multiply that factor by `cluster_memory(client)`, and pass the target data size into one of the helper functions.


tests/workloads/test_dataframe.py
```diff
@@ -0,0 +1,60 @@
...
+    partition_size = sizeof(p)
+    total_size = partition_size * df.npartitions
+    print(
+        f"~{len(p) * df.npartitions:,} rows x {len(df.columns)} columns, "
+        f"{format_bytes(total_size)} total, "
+        f"{df.npartitions:,} {format_bytes(partition_size)} partitions"
+    )
+
+
+def test_dataframe_align(small_client):
```
ian-r-rose, line 18: If I understand your intention here, these tests are aiming at growing to incorporate weak scaling. I'm thinking through how they will show up on the static site as it exists today: the current strategy is to take parameterized tests and show them on a single chart as multiple lines. I think that will work well for these, but I'm wondering if it would be clearer to the reader if they were named something like `test_dataframe_align_weak_scaling` etc. I suppose it's a bit of letting charts wag the dog, but doesn't offend me too much.

tests/workloads/test_dataframe.py
```diff
@@ -0,0 +1,60 @@
...
+    memory = cluster_memory(small_client)  # 76.66 GiB
+
+    df = timeseries_of_size(
+        memory // 4,
+        start="2020-01-01",
+        freq="1200ms",
+        partition_freq="24h",
+        dtypes={i: float for i in range(100)},
+    )
+    print_dataframe_info(df)
```
ian-r-rose, line 55: How expensive do these local computes get? Is it ever a non-negligible amount of the test runtime?

tests/utils_test.py
```diff
@@ -0,0 +1,185 @@
...
+    final = tuple(s for s in resolved_shape if s is not None)
+    assert len(final) == len(resolved_shape), resolved_shape
+
+    actual_nbytes = np.prod(final) * dtype.itemsize
+    error = (actual_nbytes - target_nbytes) / actual_nbytes
+    assert abs(error) < max_error, (error, actual_nbytes, target_nbytes, final)
+    return final
+
+
+def wait(thing, client, timeout):
```
ian-r-rose, line 82: I would love it if `distributed.wait` did this.

tests/workloads/test_array.py
```diff
@@ -0,0 +1,145 @@
...
+        padded = da.pad(arr, pad_width=[(0, 1), (0, 0), (0, 0)], mode="wrap")
+        old_chunks = padded.chunks
+        new_chunks = list(old_chunks)
+        new_chunks[0] = 5001
+        rechunked = da.rechunk(padded, chunks=new_chunks)
+        return rechunked
+
+    up = pad_rechunk(u)
+    vp = pad_rechunk(v)
+    result = dx[..., None] * up - dy[..., None] * vp
```
ian-r-rose, line 127: I recoil at this definition of vorticity

ian-r-rose (commented):
You previously expressed interest in restructuring some of the test layout -- do we still want to rip that bandaid off?

tests/workloads/test_dataframe.py
gjoseph92: I think it's negligible (haven't actually timed it, but never notice it locally). And since the partition size is fixed, it should always remain negligible.

Nonetheless, I'm a little iffy on using the automatic timing from `small_client`; I might kinda prefer using `benchmark_time` manually. But as we start adding more things to benchmark, the Cartesian explosion of `small_client_time`, `small_client_time_memory`, etc. gets massive. Maybe we just have a `small_client` and `small_client_no_benchmarks` fixture, and you have to add every fixture manually in the latter case? That's annoying though, especially when adding a new thing to measure later.

tests/utils_test.py
gjoseph92: `raise: bool = False` would probably be a reasonable argument to have there.

https://github.com/dask/distributed/issues/6885

tests/workloads/test_array.py
gjoseph92: Go complain to @TomNicholas :)

No, it's entirely possible I misunderstood the example in https://github.com/dask/distributed/issues/6571 and it wasn't meant to be vorticity. Ideas for a better name for it?

tests/workloads/test_dataframe.py
gjoseph92: > the current strategy is to take parameterized tests and show them on a single chart as multiple lines

Ah, I didn't realize that's how they'd show up. I was thinking we'd eventually want to test both strong and weak scaling, so annoyingly, I imagine we'd want a separate chart for the strong vs weak case (but still probably multiple lines for each cluster size within them?). I wonder if we can make some sort of special `scaling_type` fixture that `pytest.mark`s the different cases, and we could use those marks in the dashboard script to separate them?

ian-r-rose: From an implementation standpoint, I would love to not have to rely on pytest magic (or even importing pytest) to generate any data visualizations (just the data in sqlite). If you wanted to test both strong and weak scaling *within the same test*, I might prefer to just parameterize with a magic `id` string that we can identify in the test name. (i.e., if we see `weak-scaling` and `strong-scaling` in the parameterized test name, make two charts).

tests/workloads/test_dataframe.py
ian-r-rose: Sure, I'm happy with using the `benchmark_time` fixture manually. I liked having the "auto" fixture for ease of use to get off the ground quickly, but I see no problem with more custom stuff per test. I also wouldn't worry too much about re-using the `small_client` fixture. It wouldn't be hard to create a module-scoped cluster fixture that does your own thing for these workflows.

Is your worry only about the "wall clock" timing measuring some extra local work + client/cluster communication? If so, I might argue that that measure will always be a bit noisy, and it would be worth it to push on some [task group timing statistics](https://gist.github.com/ian-r-rose/b826ef75e3cdc6e83e3b31878eaf1305) as a more cluster-centric view on timing data. Those two measures would be pretty complementary.

tests/workloads/test_array.py
ian-r-rose: I was mostly joking :)

I'd expect it to be something like `v * dx - u * dy`, where `dx` and `dy` are actually `1/dx` and `1/dy`, but it doesn't actually matter for what's being measured here.

TomNicholas: [This is where it's coming from](https://mitgcm.readthedocs.io/en/latest/algorithm/algorithm.html?highlight=realtive%20%20vorticity#relative-vorticity), but for the dask graph we are trying to test I agree it doesn't matter.

tests/utils_test.py
```diff
@@ -0,0 +1,185 @@
+import dask
```
ian-r-rose, line 1: This needs a `from __future__ import annotations`

tests/workloads/test_dataframe.py
gjoseph92: > Is your worry only about the "wall clock" timing measuring some extra local work + client/cluster communication?

Yeah, it is, which I agree isn't a big deal and should fall within the normal noise of runtime anyway.

tests/workloads/test_dataframe.py
gjoseph92: I wasn't saying that the dashboard script would use pytest magic. Just that we could use such pytest magic to parameterize the tests over strong vs weak scaling mode, then pull out that parameter and write it into the database, vaguely like:

```diff
diff --git a/conftest.py b/conftest.py
index 78fb427..6667f25 100644
--- a/conftest.py
+++ b/conftest.py
@@ -184,6 +184,7 @@ def test_run_benchmark(benchmark_db_session, request, testrun_uid):
             python_version=".".join(map(str, sys.version_info)),
             platform=sys.platform,
             ci_run_url=WORKFLOW_URL,
+            scaling=request.keywords.get("scaling", None),
         )
         yield run
 ```
Then the dashboard script could just group by the `scaling` column to render different scaling modes in different charts.

**Comment by gjoseph92** [comment]
> You previously expressed interest in restructuring some of the test layout -- do we still want to rip that bandaid off?

We probably do, but let's do it in a separate PR. I've managed to organize these in a way I'm satisfied with (they're separate from all the other benchmarks right now), but it might be good to eventually mix some of the other ones into these?

@ian-r-rose do you think this is ready to merge? The failed tests look like running out of AWS vCPU quota and things like that, not actual failures (so it's hard to know).

**Comment by ian-r-rose** [comment]
> @ian-r-rose do you think this is ready to merge? The failed tests look like running out of AWS vCPU quota and things like that, not actual failures (so it's hard to know).

Yes, I think this is ready, except they are not being run right now (unless I'm missing something?). The test workflow currently runs the different directories separately in their own jobs, so nobody is running the `workflows` directory. I'd like to discuss changing that (probably in the same refactor as we were discussing above), but for now we should probably just give `workflows` their own job.

**Comment by gjoseph92** [comment]
Okay, I've gotten everything working except for `tests/benchmarks/test_array.py::test_climatcic_mean`.

When I run this test on its own against coiled-runtime 0.0.4, it does fine. The dashboard looks as it should, tasks go quickly. There's a little bit of spillage but not much.

However, it's always failing in the full CI job. With a lot of pain and watching GitHub CI logs and clicking at random at clusters on the coiled dashboard, I managed to find the cluster that was running the test and watch it. The dashboard looked a bit worse, more data spilled to disk. Workers kept running out of memory and restarting. So progress was extremely slow, and kept rolling back every time a worker died.

Screen Shot 2022-08-16 at 7 53 06 PM

**I think this demonstrates a real failure case. This workload *should* work in the allotted time, but it doesn't in a real-world case (when other things have run before it).**

What do we want to do here? Merge it skipped? **How do we want to handle integration tests for things that are currently broken?** It's important not to forget about them, and also valuable to see when exactly they get fixed. If we skip it, I can try to remember to manually test it and un-skip it when something gets merged upstream that resolves it, but that's easy to mess up, and seeing the clear history gives us both traceability and a good story.

-----

Theories for why it's failing:
* On [distributed==2022.6.0](https://github.com/coiled/coiled-runtime/runs/7869994497?check_suite_focus=true#step:5:41), `MALLOC_TRIM_THRESHOLD_` [hasn't been set yet](https://distributed.dask.org/en/stable/changelog.html#id9) by default. That might make the difference. Note though that the test passes even without it being set, if it's run on a fresh cluster. So that's clearly not the only problem. Plus, we're `client.restart()`-ing the workers before every test, so the _workers_ should be in the same brand-new state regardless of whether the test is run on its own, or after others. However, `client.restart()` doesn't do that much to the scheduler, so maybe that's where the problem is.
* We've know that every subsequent time you submit a workload to the scheduler, it runs slower and slower, and scheduler memory grows and grows: https://github.com/dask/distributed/issues/4987#issuecomment-877473670. (There's no reason to think things have changed since that research last year.)

    As the scheduler gets sluggish, it will be slower to both tell workers about data-consuimg downstream tasks to run (instead of the data-producing root tasks they've already been told to run), _and_ it will be slower to allow them to delete keys that are completed and aren't needed anymore. Note that just because a worker runs a downstream task (like writing a chunk to zarr) doesn't mean the worker gets to immediately release the upstream data—it must be explicitly told by the scheduler to do so. If the scheduler is slow, the worker will go load even more data into memory while keeping around the chunks that have already been written to zarr and should have been released.

    Thus we see the double-whammy of root task overproduction: as soon as the delicate balance of scheduler latency is thrown off, workers will simultaneously produce memory faster than they should, and release memory slower than they should:
    - https://github.com/dask/distributed/issues/5223
    - https://github.com/dask/distributed/issues/5114

    Basically, I think this will only be fixed by https://github.com/dask/distributed/pull/6614, or by understanding and fixing whatever's causing the scheduler to slow down (which is further out) https://github.com/dask/distributed/issues/4987.

**Comment by ntabris** [comment]
> With a lot of pain and watching GitHub CI logs and clicking at random at clusters on the coiled dashboard, I managed to find the cluster that was running the test and watch it

@gjoseph92 any chance you have the cluster id or a link to the details page?

**Comment by gjoseph92** [comment]
@ntabris sadly no, I closed that tab too long ago. And AFAICT there are no logging statements in the tests that allow you to see which cluster a test was running on, and the cluster names in tests are random.

**Comment by ntabris** [comment]
> AFAICT there are no logging statements in the tests that allow you to see which cluster a test was running on, and the cluster names in tests are random

Hm, that might be good to change.

Regardless I found it based on the public IP for the scheduler (shown in your screenshot!).

https://cloud.coiled.io/dask-engineering/clusters/51386/details

I might poke around a little at the logs tomorrow (and anyone else is of course welcome to as well).

**Comment by ntabris** [comment]
Oh, the first worker I checked shows over 2000 lines of solid:

```
distributed.worker_memory - WARNING - Unmanaged memory use is high. This may indicate a memory leak or the memory may not be released to the OS; see https://distributed.dask.org/en/latest/worker-memory.html#memory-not-released-back-to-the-os for more information. -- Unmanaged memory: 5.09 GiB -- Worker memory limit: 7.11 GiB
```

It's around 10/s for about 5 minutes.

**Comment by ian-r-rose** [comment]
Thanks for writing this up @gjoseph92. I'm trying this out right now, and was able to reproduce some of this locally (well, not locally, but you know what I mean). If I run `test_climatcic_mean` by itself, it does okay, but if I run the whole module things start to go badly in the same way you describe.

I'm still looking into this, but I wanted to note that I looked at the scheduler system tab while things were going poorly, and didn't see anything overly concerning:
image

**Comment by ian-r-rose** [comment]
> **I think this demonstrates a real failure case. This workload _should_ work in the allotted time, but it doesn't in a real-world case (when other things have run before it).**
> 
> What do we want to do here? Merge it skipped? **How do we want to handle integration tests for things that are currently broken?** It's important not to forget about them, and also valuable to see when exactly they get fixed. If we skip it, I can try to remember to manually test it and un-skip it when something gets merged upstream that resolves it, but that's easy to mess up, and seeing the clear history gives us both traceability and a good story.

I like the idea of having a good history of memory usage, duration, etc, even for failing tests. I would note, however, that the measurements for tests that don't finish aren't really indicative of anything real. The duration would just show the timeout until things were fixed. The memory metrics would be a bit more meaningful, but I'd still hesitate to interpret them. So I might still advocate for skipping the problematic tests until we can figure out what's going on.

File: tests/benchmarks/test_array.py [hunk 6]
```diff
@@ -0,0 +1,145 @@
...
+    )
+    print_size_info(memory, target_nbytes, data)
+    # 383.20 GiB - 3924 100.00 MiB chunks
+
+    result = da.sum(data, axis=1)
+
+    wait(result, small_client, 10 * 60)
+
+
+def test_climatcic_mean(small_client):
```
ian-r-rose, line 63:
```suggestion
def test_climatic_mean(small_client):
```

gjoseph92: 🤦

gjoseph92:
That is interesting that the scheduler looks okay. For now, I've just skipped it. We can talk about what to do in general on https://github.com/coiled/coiled-runtime/issues/252, and for the specific test on https://github.com/coiled/coiled-runtime/issues/253.

tests/workloads/test_dataframe.py
ian-r-rose: This sounds like a good approach

ian-r-rose:
This is all green -- anything else you'd like to do here before we merge @gjoseph92?

gjoseph92:
Nope, I think this is good! Thanks for the review.

PR #245, opened 2022-08-12
Fix static site gen

### LINKED ISSUE #244: ⚠️ CI failed ⚠️
[Workflow Run URL](https://github.com/coiled/coiled-runtime/actions/runs/2849573466)

ian-r-rose:
Handle BaseExceptions in dashboard generation, include xgboost and dask-ml in dashboard env

Fixes #244. No matter how careful you think you are being, somebody comes along and inherits from `BaseException`.

### DIFF
--- a/ci/environment-dashboard.yml
+++ b/ci/environment-dashboard.yml
@@ -9,6 +9,8 @@ dependencies:
   - panel=0.13
   - coiled
   - dask
+  - dask-ml
   - distributed
+  - xgboost
   - pandas
   - tabulate
--- a/dashboard.py
+++ b/dashboard.py
@@ -35,7 +35,7 @@ def get_test_source():
                     if not callable(fn):
                         continue
                     source[f[len("tests/") :] + "::" + test] = inspect.getsource(fn)
-        except Exception:
+        except BaseException:  # Some pytest exceptions inherit directly from BaseException
             pass
     return source
 


ci/environment-dashboard.yml
```diff
@@ -9,6 +9,8 @@ dependencies:
   - panel=0.13
   - coiled
   - dask
+  - dask-ml
   - distributed
+  - xgboost
```
jrbourbeau, line 14: Just checking, are these still needed with the `BaseException` change below? I'm curious why these are needed

ian-r-rose: The thing that was failing was the part that grabs the code snippet for tests. That was intended to be very fault tolerant, though I missed just how fault tolerant it needed to be. The change to `BaseException` allows it to just not get the code snippet if an import of the module fails. But if we include the packages, the import should succeed, and we get the snippet.

However, now that I think of it, I think you are right that we could make this work without `xgboost` in that environment, thanks to the defensiveness of `importorskip`. I'll take a look at that.

Also, this code introspection is obviously kind of a hack, if you have ideas for how to improve it, I'd be curious to hear them.

ian-r-rose: > However, now that I think of it, I think you are right that we could make this work without xgboost in that environment, thanks to the defensiveness of importorskip. I'll take a look at that.

Never mind, this was off-base. I think this PR is a decent way forward, and we can just be super fault-tolerant here.

PR #250, opened 2022-08-16
Ensure we have enough data to semi-reliably identify regressions

### LINKED ISSUE #249: Ignore decreases in memory usage / wall time in regression detection
Currently our regression detection will treat increases and decreases in metrics equally. However, decreases in memory usage and wall time shouldn't trigger a regression detection. See https://github.com/coiled/coiled-runtime/pull/247#issuecomment-1216830194 for some additional context. 

I spoke with @ian-r-rose offline and he said he has a quick fix for this

ian-r-rose:
Fixes #249

### DIFF
--- a/detect_regressions.py
+++ b/detect_regressions.py
@@ -54,9 +54,8 @@ def detect_regressions(database_file):
                 pass
             else:
                 for metric in ["duration", "average_memory", "peak_memory"]:
-                    # check that we have enough data to do some stats
-                    # currently we don't have a lot of data but eventually use 10 instead of 6.
-                    if len(df_test[metric]):
+                    # check that we have enough data to do some stats (last three plus previous ten)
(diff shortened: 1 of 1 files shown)


jrbourbeau (commented):
Thanks @ian-r-rose! It's not totally clear to me how this relates to https://github.com/coiled/coiled-runtime/issues/249. This appears to be about having enough previous runs to compare to. Is decreased memory usage / wall time still treated as a regression, or am I missing something?

ian-r-rose:
>  Is decreased memory usage / wall time still treated as a regression, or am I missing something?

I believe that it is not treated as a regression (that is to say, it's correct right now). The issue here was just that there was not enough history in the timeseries to actually make a decent measure of a regression.

jrbourbeau (approved):
Let's give it a try then 👍

PR #259, opened 2022-08-18
Add 0.1.0 release to test build matrix

jrbourbeau:
This PR:

1. Adds new Coiled software environments which use the `coiled-runtime=0.1.0` release
2. Adds `0.1.0` and removes `0.0.3` from our CI builds

### DIFF
--- a/.github/workflows/software-environments-stable.yml
+++ b/.github/workflows/software-environments-stable.yml
@@ -15,20 +15,13 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.8", "3.9", "3.10"]
-        runtime-version: ["0.0.3", "0.0.4"]
+        runtime-version: ["0.0.4", "0.1.0"]
         server: ["https://cloud.coiled.io", "https://staging.coiledhq.com"]
         include:
           - server: "https://cloud.coiled.io"
             token-name: COILED_BENCHMARK_BOT_TOKEN
           - server: "https://staging.coiledhq.com"
             token-name: COILED_STAGING_BENCHMARK_BOT_TOKEN
-          - python-version: "3.7"
-            runtime-version: "0.0.3"
-        exclude:
-          # `coiled-runtime=0.0.3` pins to a version of Dask
-          # that doesn't support Python 3.10
-          - python-version: "3.10"
-            runtime-version: "0.0.3"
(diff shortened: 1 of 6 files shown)


.github/workflows/software-environments-stable.yml
```diff
@@ -15,20 +15,13 @@ jobs:
...
             token-name: COILED_BENCHMARK_BOT_TOKEN
           - server: "https://staging.coiledhq.com"
             token-name: COILED_STAGING_BENCHMARK_BOT_TOKEN
-          - python-version: "3.7"
-            runtime-version: "0.0.3"
-        exclude:
-          # `coiled-runtime=0.0.3` pins to a version of Dask
-          # that doesn't support Python 3.10
-          - python-version: "3.10"
-            runtime-version: "0.0.3"
```
ian-r-rose, line 31: Huzzah

ian-r-rose (approved):

.github/workflows/software-environments-stable.yml
jrbourbeau: 🎉

jrbourbeau: This should also mean no more flaky "Scheduler didn't phone home" failures

jrbourbeau:
So we need the software environments before we can add `0.1.0` to our test suite matrix...isolating those changes over in https://github.com/coiled/coiled-runtime/pull/260. We should merge that PR and then after the software environments are build, then rerun things here

jrbourbeau:
The failing tests here are unrelated to the changes in this PR and have been seen in `main` (xref https://github.com/coiled/coiled-runtime/issues/265)

jrbourbeau:
@ian-r-rose, just checking, is there anything I need to do re: benchmarking infrastructure when it comes to adding / removing runtime versions? If not, I'll go ahead and press the green button here

ian-r-rose:
> @ian-r-rose, just checking, is there anything I need to do re: benchmarking infrastructure when it comes to adding / removing runtime versions? If not, I'll go ahead and press the green button here

I don't think there is any action required right now.

We'll probably want to make some updates to stop generating panels for out-of-date tests or matrix builds, but I don't think we need to block on that

PR #282, opened 2022-08-26
Add performance reports

fjetter:
This automatically adds performance reports that are uploaded to Coiled

Naming scheme

`cluster-name`-`test_name`

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -17,6 +17,7 @@ import distributed
 import filelock
 import pytest
 import s3fs
+from coiled import performance_report
 from distributed import Client
 from distributed.diagnostics.memory_sampler import MemorySampler
 from toolz import merge
@@ -316,6 +317,7 @@ def small_cluster(request):
 
 @pytest.fixture
 def small_client(
+    request,
     small_cluster,
     upload_cluster_dump,
     benchmark_all,
@@ -325,7 +327,9 @@ def small_client(
         client.wait_for_workers(10)
         client.restart()
 
-        with upload_cluster_dump(client, small_cluster):
+        with upload_cluster_dump(client, small_cluster), performance_report(
+            f"{small_cluster.name}-{request.node.name}"
+        ):
             with benchmark_all(client):
                 yield client
 
--- a/conftest.py
+++ b/conftest.py
@@ -12,12 +12,12 @@ import time
 import uuid
 from distutils.util import strtobool
 
+import coiled
 import dask
 import distributed
 import filelock
 import pytest
 import s3fs
-from coiled import performance_report
 from distributed import Client
 from distributed.diagnostics.memory_sampler import MemorySampler
 from toolz import merge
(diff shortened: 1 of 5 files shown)


ian-r-rose:
I don't think this will work well due to constraints on the platform side (unless something has changed)

https://github.com/coiled/coiled-runtime/issues/281#issuecomment-1228604493

ncclementi:
I agree with Ian. Let's keep in mind that there is a limit on the history to keep performance reports, which is 5-10 per account, I just confirmed this with Samantha and nothing has changed, so I don't see how this will be useful.

fjetter:
We should be good to go now

ian-r-rose:
I think we should also dump the URL to the performance report in the benchmarking db, which will make it easier to link to from our charts and regression scripts.

There is already a slot in the schema, we just need to populate it.

conftest.py
```diff
@@ -401,3 +410,24 @@ def _upload_cluster_dump(client, cluster):
...
+@pytest.fixture
+def upload_performance_report(test_run_benchmark):
+    @contextlib.contextmanager
+    def _upload_performance_report(name):
+        with coiled.performance_report(name) as obj:
+            yield
+        # TODO: This will not scale well if we have many reports. Instead, the
+        # performance report ctx manager should somehow yield an object that has
+        # the url
+        if obj is None:
```
fjetter, line 424: I opened https://gitlab.com/coiled/cloud/-/merge_requests/4985 to make this more accessible.

dashboard.py
```diff
@@ -98,6 +104,9 @@ def make_timeseries(originalname, df, spec) -> altair.Chart | None:
                 altair.Tooltip("dask_version:N", title="Dask"),
                 altair.Tooltip(f"{spec.field}:Q", title=spec.label),
                 altair.Tooltip("ci_run_url:N", title="CI Run URL"),
+                altair.Tooltip(
+                    "performance_report_url:N", title="Performance Report URL"
+                ),
```
fjetter, line 109: I'm not sure if the tooltip will help all that much because we cannot click on the tooltips. Another delivery medium might be better suited.
In a perfect world, clicking would open a dedicated page for that test with additional info, etc.

crusaderky: https://altair-viz.github.io/gallery/scatter_href.html

crusaderky (approved):

ian-r-rose:
There are a few other cluster fixtures which should also get this treatment (not everything is covered by `small_client`)

conftest.py
ian-r-rose: I agree that this should be exposed via the `coiled` API.

I wonder if this would be easier today if we capture stdout from the context manager cleanup, which includes the URL

fjetter: if the upstream PR is accepted I'd rather keep it as is until this is deployed. I doubt that we'll accumulate a significant amount of reports until this version is rolled out and I'd rather not mess with stdout parsing

dashboard.py
fjetter: href already point to the CI job

conftest.py
fjetter: FWI upstream was merged

fjetter:
> There are a few other cluster fixtures which should also get this treatment (not everything is covered by small_client)

done

jrbourbeau (commented):
Thanks @fjetter! 

Unfortunately it looks like we're hitting a limit `coiled` has on performance report size now. cc @ntabris for visibility. Would folks be okay, for example, increasing that limit?

FWIW `test_double_diff` has other problems (xref https://github.com/coiled/coiled-runtime/issues/316, https://github.com/coiled/coiled-runtime/issues/265) so if that's the only problematic report, I'd be okay temporarily skipping that one for now.

ntabris:
> Unfortunately it looks like we're hitting a limit `coiled` has on performance report size now. cc @ntabris for visibility. Would folks be okay, for example, increasing that limit?

I think the current limit is 10MB. Want to suggest a bigger number? Do these grow linearly with number of tasks / duration of run / something like that?

fjetter:
> Do these grow linearly with number of tasks / duration of run / something like that?

Yes, probably it grows with both tasks and duration. afaik, we never looked into making the reports smaller. What I would suggest is to drop the limit entirely and compress the file before uploading. IIUC we do not intend to serve the generated HTML page anyhow but would require users to download it instead

fjetter:
I rewrote the upload such that we only log an error if the upload is rejected due to size. This is not optimal but we should be able to work with this for all other test.

we could also consider simply archiving the performance reports in github

hendrikmakait:
We may want to increase the `task-stream-length` from 100,000 tasks by default. Otherwise it will get truncated for some jobs such as `test_double_diff`.

PR #283, opened 2022-08-26
Imporve detect regression in PRs [WIP]

ncclementi:
See https://github.com/coiled/coiled-runtime/issues/270#issuecomment-1227798553 

I think comparing just the PR run to a threshold of mean + 2std might be strict but let's see what happens. 

cc: @ian-r-rose for visibility

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -388,6 +388,11 @@ jobs:
 
       - name: Run detect regressions
         run: |
+          if [[ ${{ github.event_name }} = 'pull_request' ]]
+          then
+            export IS_PR='true'
+          fi
+          echo "IS_PR=$IS_PR"
           python detect_regressions.py
       
       - name: Create regressions summary
--- a/detect_regressions.py
+++ b/detect_regressions.py
@@ -1,10 +1,12 @@
+import os
 import pathlib
+from distutils.util import strtobool
 
 import pandas as pd
 import sqlalchemy
 
 
-def detect_regressions(database_file):
+def detect_regressions(database_file, is_pr=False):
 
     engine = sqlalchemy.create_engine(f"sqlite:///{database_file}")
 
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -388,6 +388,11 @@ jobs:
 
       - name: Run detect regressions
         run: |
+          if [[ ${{ github.event_name }} = 'pull_request' ]]
+          then
+            export IS_PR='true'
+          fi
+          echo "IS_PR=$IS_PR"
```
ian-r-rose, line 395: I think we need to put this into `$GITHUB_ENV` to have it persist across steps (see, e.g., the old "maybe mark for benchmarking")

detect_regressions.py
```diff
@@ -57,12 +59,19 @@ def detect_regressions(database_file):
                     # check that we have enough data to do some stats (last three plus previous ten)
                     if len(df_test.loc[df_test[metric].notna()]) > 13:
```
ian-r-rose, line 60: I suppose this `13` could also be `10 + nlast`, but I wouldn't sweat it.

ian-r-rose (commented):
This looks like a good approach to me!

.github/workflows/tests.yml
ncclementi: We would if we need it for in other steps but we don't use it anywhere else, that's why I didn't put it into $GITHUB_ENV. Unless I'm missing something here.

ncclementi:
It looks like this condition makes might be too strict. We have 19 regressions detected https://github.com/coiled/coiled-runtime/runs/8074223218?check_suite_focus=true .  If I use 2.5 std then it goes to 9 regressions detected. Not sure what exactly we want to do here.

ian-r-rose:
I'd say [this](https://github.com/coiled/coiled-runtime/runs/8082970353?check_suite_focus=true) is a pretty clear example of some false positives. I wonder if we should have a *minimum* threshold delta. That is to say, instead of `mean + 2stdev`, we could do `mean + min(2stdev, 1 GiB)` for memory, `mean + min(2stdev, 1 s)` for duration, or similar.

.github/workflows/tests.yml
ian-r-rose: You're absolutely right, my mistake

ian-r-rose:
> It looks like this condition makes might be too strict. We have 19 regressions detected https://github.com/coiled/coiled-runtime/runs/8074223218?check_suite_focus=true . If I use 2.5 std then it goes to 9 regressions detected. Not sure what exactly we want to do here.

It would make sense to me to have a higher threshold for PR checks. Thinking about it from a Bayesian perspective, I think the basic question is: "how much would it take for me to update my priors that a regression was not introduced?".  Short of an actual changepoint analysis, I think I would want something like the following:

1. Consistent results for multiple runs in a row to detect small regressions (this is what we do on `main`)
2. On a PR branch, we won't have good statistics, so we want to be much more certain it's not natural noise. So I think 3 or 4 standard deviations would be okay.

ncclementi:
> I'd say [this](https://github.com/coiled/coiled-runtime/runs/8082970353?check_suite_focus=true) is a pretty clear example of some false positives. I wonder if we should have a _minimum_ threshold delta. That is to say, instead of `mean + 2stdev`, we could do `mean + min(2stdev, 1 GiB)` for memory, `mean + min(2stdev, 1 s)` for duration, or similar.

I like this idea. I'll take look at it

ian-r-rose (approved):
This looks good to me, let's see how CI likes it!

ncclementi:
I've realized that we want `max(2stdev, unit_of_messearument)` instead of min, I've tried it locally and this seems like a nice approach. Plus I pushed the `n_std` for PRs to be 3.

PR #286, opened 2022-08-29
Factor out cluster and client

fjetter:
I'm currently rerunning a couple of our benchmarks and am trying to inspect the cluster mid-flight / after the benchmark. Within pytest, this is a bit awkward. I can insert breakpoints, etc. but the experience is not nice.

I could move / copy the entire test code to a jupyter notebook and gain much more flexibility. To stay as close to the original it would be nice if we exported the standard functions as ctxmanagers or similar.

This is an example of how we could do this. Thoughts?

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -1,0 +1,41 @@
+from __future__ import annotations
+
+import contextlib
+import uuid
+
+from coiled import Cluster
+from distributed import Client
+
+
+@contextlib.contextmanager
+def small_cluster(
+    backend_options=None,
+    module=None,
+):
+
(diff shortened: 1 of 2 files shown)


utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+import contextlib
+import uuid
+
+from coiled import AWSOptions, Cluster
+from distributed import Client
+
+
+@contextlib.contextmanager
+def small_cluster(
```
fjetter, line 11: naming of module and functions TBD

Issue #292, opened 2022-08-31
Automation of benchmark comparison


fjetter:
## Requirements
- Be able to compare two distinct git handles (commit, branch, etc.) of dask/dask and/or dask/distributed; needs to target forks
- It is fine if it is merely possible to use a commit OR a branch OR a PR etc. instead of generic git refs
- Tests are used from the current commit / PR of the coiled-runtime
- Initial iteration must be extendable to compare N commits
- One or many CI job(s) is spawned that runs *all* available benchmarks and collects a summary
- The CI job collects all data that is also collected by the ordinary benchmark tests
- The CI job produces a summary / report comparing these single point numbers

## Nice to haves
- Varying config options
- Select a subset of the benchmarks to run
- Varying environment 
- Varying test code for different git handles (if a prototype change breaks an API, for example)
- We also want to collect memory sampling, performance reports, etc.

crusaderky:
I went through the codebase. @jrbourbeau @ncclementi @ian-r-rose please correct me if I say something wrong or impossible.

The `Tests` github action does the following:
1. build the `latest` conda package for `coiled-runtime`
2. create coiled software environments for `latest`, `0.1.0`, and `0.0.4`, in cartesian product with different versions of python
3. run the full battery of tests against the coiled environments created at step 2
4. collect and compare results
5. destroy coiled software environments

If that's the case, my plan is fairly straightforward:

1. Create additional software environments
2. Create a new static HTML report to compare whatever was run in the battery of tests

Sample workflow
-----------------------
I want to test https://github.com/dask/distributed/pull/6614.

I create a new branch of coiled-runtime, in my own fork: `crusaderky/coiled-runtime/AB_6614`

In the new branch, I create the following files. Each file is a bunch of ``kwargs`` to `coiled.create_software_environment`:

#### AB_parameters/6614_before.yaml
```yaml
conda:
    - python=3.9
    - coiled-runtime=0.1.0
pip:
    - dask==2022.8.1
    - git+https://github.com/dask/distributed@f07f384eb0b8b6bad7518e8535e42fdba106ed4b
```

#### AB_parameters/6614_inf.yaml
```yaml
conda:
    - python=3.9
    - coiled-runtime=0.1.0
pip:
    - dask==2022.8.1
    - git+https://github.com/dask/distributed@dd81b424971e81616e1a52fa09ce4698a5002d41
```

#### AB_parameters/6614_12.yaml
```yaml
conda:
    - python=3.9
[...]

PR #294, opened 2022-09-01
WIP barchart for comparing different commits rather than a timeseries.

ian-r-rose:
image

Towards fixing #292


crusaderky:
I'm picking this up

PR #235, opened 2022-08-09
Package Sync

ian-r-rose:
Pushing #222 a bit further to see if anything breaks. This

1. Removes a couple of controlling environment variables, which should simplify local development as well as CI workflows. Specifically `TEST_UPSTREAM` has been folded into `COILED_RUNTIME_VERSION`, which can be a version number, "upstream", or "latest", and `COILED_SOFTWARE_VERSION` is no longer necessary in a `package_sync` universe.
2. Removes the setup/cleanup jobs. This also means there are fewer artifacts passed around between jobs, which is nice.
3. "upstream" is now run on pull requests

PR #297, opened 2022-09-01
test for regressions

ian-r-rose:
Opening an empty PR to see the current state of things


ian-r-rose:
Correct -- I'm trying to compare to #235

PR #296, opened 2022-09-01
[WIP] Run test suite on arbitrary coiled environments

crusaderky:
This is one of two PRs to address #292.
This PR adds the ability to run the whole test suite on arbitrary additional coiled software environments.
The other PR will generate a static HTML report (#294 or superseding PR)

TODO documentation

PR #307, opened 2022-09-06
A/B test report

### LINKED ISSUE #292: Automation of benchmark comparison

crusaderky:
- Partially closes #292
- Complements #296
- Supersedes #294

DONE: bar chart by test
TODO: bar chart by runtime

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -7,6 +7,7 @@ import inspect
 import os
 import pathlib
 import sys
+from typing import Literal
 
 import altair
 import pandas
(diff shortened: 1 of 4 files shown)


dashboard.py
```diff
@@ -17,51 +17,73 @@
...
+            spec.field,
+            "name",
+            "dask_version",
+            "distributed_version",
+            "runtime",
+        ]
+    ]
+    # Altair will re-sort alphabetically, unless it's explicitly asked to sort by
+    # something else. Could not find an option to NOT sort.
+    df["order"] = pandas.RangeIndex(0, df.shape[0])
```
crusaderky, line 58: @ian-r-rose any ideas?

ian-r-rose: You may be able to do this with explicitly passing `sort=None`

Issue #306, opened 2022-09-05
`distributed=2022.6.0` used by `v0.1.0` contains regression


hendrikmakait:
In `distributed=2022.6.0`, a regression was introduced (https://github.com/dask/distributed/issues/6624) that causes issues with `map_overlap`. This regression has been fixed in `2022.6.1` (https://github.com/dask/community/issues/258#issuecomment-1165546771). I suggest a new minor release of `coiled-runtime` that bumps the version of `dask` and `distributed`.

XREF: This causes a newly created test for #305 to fail.

ncclementi:
Closing this as we are planning on releasing 0.2.0 soon.

PR #308, opened 2022-09-06
Tests for work stealing

### LINKED ISSUE #305: Stealing sensitive tests
We'd like to have a couple of tests that are related to stealing sensitive workloads.

1. Very simple example, currently causing problems (https://github.com/dask/distributed/issues/6573)

```python
import time
import dask
import distributed

# dask.config.set({"distributed.scheduler.work-stealing": False})
client = distributed.Client(n_workers=4, threads_per_worker=1)
root = dask.delayed(lambda n: "x" * n)(dask.utils.parse_bytes("1MiB"), dask_key_name="root")
results = [dask.delayed(lambda *args: None)(root, i) for i in range(10000)]
dask.compute(results)
```

2. Upscaling situation (example from https://github.com/dask/distributed/issues/4471)

```python
import time
import distributed
import dask.array as da
import dask


client = distributed.Client(n_workers=1, threads_per_worker=1, memory_limit='1G')
print(client.dashboard_link)

# Slow task.
def func1(chunk):
    if sum(chunk.shape) != 0: # Make initialization fast
        time.sleep(5)
    return chunk

def func2(chunk):
    return chunk

data = da.zeros((30, 30, 30), chunks=5)
result = data.map_overlap(func1, depth=1, dtype=data.dtype)
result = result.map_overlap(func2, depth=1, dtype=data.dtype)
future = client.compute(result)

print('started computation')

time.sleep(11)
# print('scaling to 4 workers')
# client.cluster.scale(4)

time.sleep(5)
print('scaling to 20 workers')
client.cluster.scale(20)

_ = future.result()
```

3. An example that shows a very inhomogeneous compute. This should mimic a long tail in a computation graph where work stealing should reduce wall time, e.g. by using an [asymmetric distribution](https://numpy.org/doc/stable/reference/random/generated/numpy.random.lognormal.html) of sleep values

hendrikmakait:
Closes #305

### DIFF
--- a/tests/benchmarks/test_work_stealing.py
+++ b/tests/benchmarks/test_work_stealing.py
@@ -16,6 +16,7 @@ def test_trivial_workload_should_not_cause_work_stealing(small_client):
     results = [delayed(lambda *args: None)(root, i) for i in range(10000)]
     futs = small_client.compute(results)
     wait(futs)
+    _ = [fut.result() for fut in futs]
 
     def count_work_stealing_events(dask_scheduler):
         return len(dask_scheduler.events["stealing"])
@@ -77,6 +78,7 @@ def test_work_stealing_on_inhomogeneous_workload(small_client):
     results = [clog(i) for i in delays]
     futs = small_client.compute(results)
     wait(futs)
+    _ = [fut.result() for fut in futs]
 
 
 def test_work_stealing_on_straggling_worker(test_name_uuid):
@@ -106,3 +108,4 @@ def test_work_stealing_on_straggling_worker(test_name_uuid):
             results = [slowinc(i, delay=1) for i in range(1000)]
             futs = client.compute(results)
             wait(futs)
+            _ = [fut.result() for fut in futs]
--- a/tests/benchmarks/test_work_stealing.py
+++ b/tests/benchmarks/test_work_stealing.py
@@ -7,14 +7,15 @@ import pytest
 from coiled.v2 import Cluster
 from dask import delayed, utils
 from distributed import Client, wait
+from tornado.ioloop import PeriodicCallback
 
 
 @pytest.mark.xfail(reason="https://github.com/dask/distributed/issues/6573")
 def test_trivial_workload_should_not_cause_work_stealing(small_client):
     root = delayed(lambda n: "x" * n)(utils.parse_bytes("1MiB"), dask_key_name="root")
     results = [delayed(lambda *args: None)(root, i) for i in range(10000)]
-    fut = small_client.compute(results)
-    wait(fut)
+    futs = small_client.compute(results)
+    wait(futs)
 
     def count_work_stealing_events(dask_scheduler):
         return len(dask_scheduler.events["stealing"])
@@ -74,5 +75,34 @@ def test_work_stealing_on_inhomogeneous_workload(small_client):
         return n
 
     results = [clog(i) for i in delays]
-    fut = small_client.compute(results)
-    wait(fut)
+    futs = small_client.compute(results)
+    wait(futs)
+
+
+def test_work_stealing_on_straggling_worker(test_name_uuid):
+    with Cluster(
+        name=test_name_uuid,
+        n_workers=10,
+        worker_vm_types=["t3.medium"],
+        wait_for_workers=True,
+    ) as cluster:
+        with Client(cluster) as client:
+
+            def clog():
+                time.sleep(1)
+
+            @delayed
+            def slowinc(i, delay):
+                time.sleep(delay)
+                return i + 1
+
+            def install_clogging_callback(dask_worker):
+                pc = PeriodicCallback(clog, 1500)
+                dask_worker.periodic_callbacks["clog"] = pc
+                pc.start()
+
+            straggler = list(client.scheduler_info()["workers"].keys())[0]
+            client.run(install_clogging_callback, workers=[straggler])
+            results = [slowinc(i, delay=1) for i in range(1000)]
+            futs = client.compute(results)
+            wait(futs)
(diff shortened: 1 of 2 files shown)


tests/benchmarks/test_work_stealing.py
```diff
@@ -0,0 +1,108 @@
...
+import distributed
+import numpy as np
+import pytest
+from coiled.v2 import Cluster
+from dask import delayed, utils
+from distributed import Client, wait
+from tornado.ioloop import PeriodicCallback
+
+
+@pytest.mark.xfail(reason="https://github.com/dask/distributed/issues/6573")
```
hendrikmakait, line 13: We should also include this test in `distributed` once we have a fix for dask/distributed#6573

tests/benchmarks/test_work_stealing.py
```diff
@@ -0,0 +1,108 @@
...
+    wait(futs)
+
+    def count_work_stealing_events(dask_scheduler):
+        return len(dask_scheduler.events["stealing"])
+
+    count = small_client.run_on_scheduler(count_work_stealing_events)
+    assert count == 0
+
+
+@pytest.mark.xfail(
```
hendrikmakait, line 27: XREF: #306

tests/benchmarks/test_work_stealing.py
```diff
@@ -0,0 +1,111 @@
...
+from tornado.ioloop import PeriodicCallback
+
+
+@pytest.mark.xfail(reason="https://github.com/dask/distributed/issues/6573")
+def test_trivial_workload_should_not_cause_work_stealing(small_client):
+    root = delayed(lambda n: "x" * n)(utils.parse_bytes("1MiB"), dask_key_name="root")
+    results = [delayed(lambda *args: None)(root, i) for i in range(10000)]
+    futs = small_client.compute(results)
+    wait(futs)
+    _ = [fut.result() for fut in futs]
```
fjetter, line 19:
```suggestion
    client.gather(futs)
```

will give the same result

tests/benchmarks/test_work_stealing.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def count_work_stealing_events(dask_scheduler):
+        return len(dask_scheduler.events["stealing"])
+
+    count = small_client.run_on_scheduler(count_work_stealing_events)
+    assert count == 0
```
fjetter, line 25: I think this is too strict. **If** we wanted to assert anything here we should rather assert that the data is distributed uniformly but we will not be able to suppress all stealing requests. This is also something we'd need to test in distributed itself.

I actually think we don't need to assert anything here. Wall time measurements will be sufficient and we will not need to xfail this test

tests/benchmarks/test_work_stealing.py
```diff
@@ -0,0 +1,111 @@
...
+
+            time.sleep(11)
+            # print('scaling to 4 workers')
+            # client.cluster.scale(4)
+
+            time.sleep(5)
+            print("scaling to 20 workers")
+            cluster.scale(20)
+
+            _ = future.result()
```
fjetter, line 66: You need to enter the benchmark ctx managers first. If you don't want to, use `auto_*` but in this case we do not want to measure the cluster startup so the following should be correct
```suggestion
            with benchmark_time, benchmark_task_durations, benchmark_memory:
                # Slow task.
                def func1(chunk):
                    if sum(chunk.shape) != 0:  # Make initialization fast
                        time.sleep(5)
                    return chunk

                def func2(chunk):
                    return chunk

                data = da.zeros((30, 30, 30), chunks=5)
                result = data.map_overlap(func1, depth=1, dtype=data.dtype)
                result = result.map_overlap(func2, depth=1, dtype=data.dtype)
                future = client.compute(result)

                print("started computation")

                time.sleep(11)
                # print('scaling to 4 workers')
                # client.cluster.scale(4)

                time.sleep(5)
                print("scaling to 20 workers")
                cluster.scale(20)

                _ = future.result()
```

tests/benchmarks/test_work_stealing.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def test_work_stealing_on_straggling_worker(test_name_uuid):
+    with Cluster(
+        name=test_name_uuid,
+        n_workers=10,
+        worker_vm_types=["t3.medium"],
+        wait_for_workers=True,
+    ) as cluster:
+        with Client(cluster) as client:
```
fjetter, line 91: This is missing the measurements

tests/benchmarks/test_work_stealing.py
fjetter: might also be useful to define a `benchmark_all` fixture like

```python
@pytest.fixture
def benchmark_all(benchmark_memory, benchmark_time, benchmark_task_durations):
    with benchmark_memory, benchmark_time, benchmark_task_durations:
        yield
```

fjetter (changes requested):

tests/benchmarks/test_work_stealing.py
fjetter: if it's too fast to get any meaningful signal we should remove it again. asserting on this level should be out of scope for coiled-runtime

tests/benchmarks/test_work_stealing.py
fjetter:
```suggestion
```

Please remove the xfail. This test now succeeds although it's slow.

This is "fixed by https://github.com/dask/distributed/pull/7022 already and it includes this as a test case (or a version o fit)

fjetter (approved):

hendrikmakait:
The detected regressions should be unrelated given that they should be running on independent clusters and have flaked in the past.

Issue #293, opened 2022-09-01
S3 error setting up `s3_scratch` fixture


gjoseph92:
`PermissionError: The unspecified location constraint is incompatible for the region specific endpoint this request was sent to.`

This happens during `s3.mkdirs`:

https://github.com/coiled/coiled-runtime/blob/5d13ab90ff0006efac70f7586fcac0038508d36f/conftest.py#L313-L320

ntabris:
@ian-r-rose is this the semi-mysterious error you were trying to reproduce?

ian-r-rose:
Yeah, possibly related to #240 . Nothing seemed to cause any problems then.

@gjoseph92 do you have a default region set in your aws config?

gjoseph92:
I do, `us-east-2`

ian-r-rose:
Weird. Try upgrading s3fs and boto?

PR #309, opened 2022-09-06
A/B test reports

### LINKED ISSUE #292: Automation of benchmark comparison

crusaderky:
- Partially closes #292
- Complements #314
- Supersedes #294

### In scope
- Generate dashboard in PRs and upload it as an artifact
- General refactoring of dashboard.py
- Bar chart by test
- Bar chart by runtime (absolute measures)
- Bar chart by runtime / baseline runtime
### Out of scope
- Run additional runtimes on arbitrary git hashes (delivered by #296)
- Display historical variance of baseline as a bracket overlay (read below)
- Reduce noise (read below)

Screenshot from 2022-09-07 22-13-14

Screenshot from 2022-09-07 22-13-34

The plots highlight that the performance metrics are heavily affected by noise.

In some cases, this looks like a high variance between runs (which can be seen in the historical plots); in others, the difference is consistent throughout the history, but is not justified by a difference in runtime so I suspect there's some hidden warm-up cost that differs between the runs.

For example, `latest` and `0.1.0` only differ by the addition of `pynvml`, but the A/B charts comparing the two (*top right plot above*) show major differences all over the place - which are all noise. The memory usage plots look just as jittery. Ideally this chart should be used to take the informed decision to finalize a new release (all bars are negative or almost zero) and is also twitter gold - but at the moment, it's not usable due to the underlying data.

On one side, I can incorporate historical data into these plots to show brackets of variance for the baseline. I will in a successive PR. However, such brackets will just tell us what we can already see in the historical graphs: that most tests are swinging wildly and that a +/- 50% runtime is nothing to be surprised about.
We need to sit down and find the best way to actually reduce variance in runtime and memory usage.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -388,6 +388,11 @@ jobs:
 
       - name: Run detect regressions
         run: |
+          if [[ ${{ github.event_name }} = 'pull_request' ]]
+          then
+            export IS_PR='true'
+          fi
+          echo "IS_PR=$IS_PR"
           python detect_regressions.py
       
       - name: Create regressions summary
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -57,16 +57,19 @@ jobs:
           if [[ "$COMMIT" == *"test-upstream"* || ${{ github.event_name }} == "schedule" ]]
           then
             export TEST_UPSTREAM="true"
+            export AB_BASELINE="coiled-upstream-py3.9 coiled-latest-py3.9 coiled-0.1.0-py3.9"
           else
             export TEST_UPSTREAM="false"
+            export AB_BASELINE="coiled-latest-py3.9 coiled-0.1.0-py3.9"
           fi
 
           # Put TEST_UPSTREAM into $GITHUB_ENV so it can be used in subsequent workflow steps
           echo $TEST_UPSTREAM
           echo TEST_UPSTREAM=$TEST_UPSTREAM >> $GITHUB_ENV
 
-          # Put TEST_UPSTREAM into a file so it can be downloaded in subsequent workflow jobs
+          # Put env variables into files so it can be downloaded in subsequent workflow jobs
           echo $TEST_UPSTREAM > test_upstream.txt
+          echo $AB_BASELINE > ab_baseline.txt
 
       - name: Build Coiled Software Environment
         env:
@@ -104,6 +107,7 @@ jobs:
             latest.yaml
             software_name.txt
             test_upstream.txt
+            ab_baseline.txt
 
   runtime:
     name: Runtime - ${{ matrix.os }}, Python ${{ matrix.python-version }}, Runtime ${{ matrix.runtime-version }}
@@ -457,9 +461,14 @@ jobs:
           python-version: "3.9"
           environment-file: ci/environment-dashboard.yml
 
+      - name: Download software environment assets
+        uses: actions/download-artifact@v3
+        with:
+          name: software-environment-py3.9
+
       - name: Generate dashboards
         run: |
-          python dashboard.py
+          python dashboard.py -d benchmark.db -o static -b <(cat ab_baseline.txt)
 
       - name: Deploy 🚀
         uses: JamesIves/github-pages-deploy-action@4.1.7
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -441,7 +441,7 @@ jobs:
   static-site:
     needs: process-results
     # Always generate the site, as this can be skipped even if an indirect dependency fails (like a test run)
-    if: always() && github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
+    if: always()
     name: Build static dashboards
     runs-on: ubuntu-latest
     steps:
@@ -470,8 +470,15 @@ jobs:
         run: |
           python dashboard.py -d benchmark.db -o static -b <(cat ab_baseline.txt)
 
+      - name: Upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: static-dashboard
+          path: static
+
       - name: Deploy 🚀
         uses: JamesIves/github-pages-deploy-action@4.1.7
+        if:  github.ref == 'refs/heads/main' && github.repository == 'coiled/coiled-runtime'
         with:
           branch: gh-pages
           folder: static
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -57,7 +57,7 @@ jobs:
           if [[ "$COMMIT" == *"test-upstream"* || ${{ github.event_name }} == "schedule" ]]
           then
             export TEST_UPSTREAM="true"
-            export AB_BASELINE="coiled-upstream-py3.9 coiled-latest-py3.9 coiled-0.1.0-py3.9"
+            export AB_BASELINE="coiled-upstream-py3.9 coiled-0.1.0-py3.9"
           else
             export TEST_UPSTREAM="false"
             export AB_BASELINE="coiled-latest-py3.9 coiled-0.1.0-py3.9"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -468,7 +468,7 @@ jobs:
 
       - name: Generate dashboards
         run: |
-          python dashboard.py -d benchmark.db -o static -b <(cat ab_baseline.txt)
+          python dashboard.py -d benchmark.db -o static -b $(cat ab_baseline.txt)
 
       - name: Upload artifact
         uses: actions/upload-artifact@v3
--- a/dashboard.py
+++ b/dashboard.py
@@ -53,7 +53,7 @@ def load_test_source() -> None:
     print(f"Discovered {len(source)} tests")
 
 
-def align_to_baseline(df: pandas.DataFrame, baseline: str) -> pandas.DataFrame:
+def align_to_baseline(df: pandas.DataFrame, baseline: str) -> pandas.DataFrame | None:
     """Add columns
 
     - duration_baseline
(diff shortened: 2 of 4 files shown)


ian-r-rose:
@crusaderky it looks like you were hitting some permissions issues. Is it the same as #293? I've been unable to reproduce, but you might try reverting #240 to see if that helps

dashboard.py
```diff
@@ -17,62 +16,141 @@
...
+    tests being importable from this script.
     """
-    Make a single timeseries altair chart for a given test.
+    for fname in glob.iglob("tests/**/test_*.py", recursive=True):
+        mod = importlib.import_module(fname.replace("/", ".")[: -len(".py")])
+        tests = [a for a in dir(mod) if a.startswith("test_")]
+        for test in tests:
+            if (func := getattr(mod, test, None)) and callable(func):
+                # FIXME missing decorators, namely @pytest.mark.parametrize
+                source[fname[len("tests/") :] + "::" + test] = inspect.getsource(func)
```
ian-r-rose, line 39: I'd rather continue to be defensive here with a `try`/`except`

dashboard.py
```diff
@@ -17,62 +16,141 @@
...
+    if not len(df):
+        # Some tests do not have average_memory or peak_memory measures, only runtime
+        return None
+
+    by_test = len(df["fullname"].unique()) == 1
+    if by_test:
+        df = df.sort_values("runtime", key=runtime_sort_key_pd)
+        # Altair will re-sort alphabetically, unless it's explicitly asked to sort by
+        # something else. Could not find an option to NOT sort.
+        y = altair.Y("runtime", sort=altair.EncodingSortField(field="order"))
```
ian-r-rose, line 82: Does `sort=None` work?

dashboard.py
crusaderky: Why? It makes a lot more sense to me to have the whole test suite crash if someone adds a new import in a test file and doesn't add it to `ci/environment-dashboard.yml`.

crusaderky:
> @crusaderky it looks like you were hitting some permissions issues. Is it the same as #293? I've been unable to reproduce, but you might try reverting #240 to see if that helps

No,  the issue was that
- `git@github.com:coiled/coiled-runtime.git` works for git pulls from all forks and for git pushes to my own fork
- `https://github.com/coiled/coiled-runtime.git` also works for git pushes to other forks, but annoyingly it doesn't work in gitkraken

dashboard.py
crusaderky: it works

dashboard.py
ian-r-rose: Because I view this code-scraping to be a fairly fragile hack, and weird stuff can happen on import, especially when pytest is involved. For example, a `pytest.skip` at the module level raises an exception, and I don't want to force everyone to do non-idiomatic pytest stuff on account of this static report. To me, having a missing code snippet for a test is not a big deal.

ian-r-rose: I'm also unhappy about having to add stuff to `ci/environment-dashboard.yml` in the first place to just be able to use `inspect`. If you have ideas about how to get snippets without relying on importing, I'm certainly open to it. But this seemed preferable to trying to parse `.py` files to me.

.github/workflows/tests.yml
```diff
@@ -457,9 +461,14 @@ jobs:
           python-version: "3.9"
           environment-file: ci/environment-dashboard.yml
 
+      - name: Download software environment assets
+        uses: actions/download-artifact@v3
+        with:
+          name: software-environment-py3.9
```
ian-r-rose, line 467: Where are you using this? Hopefully these assets will be going away soon

crusaderky: In the following change (line 471)

dashboard.py
crusaderky: Partially reverted

crusaderky:
Ready for review (assuming CI passes)

jrbourbeau (commented):
@crusaderky it looks like there is a legitimate error in the static site generation script 

```python
Discovered 36 tests
  File "/home/runner/work/coiled-runtime/coiled-runtime/dashboard.py", line 507, in <module>
    main()
Generating static/coiled-upstream-py3.10.html
Generating static/coiled-upstream-py3.9.html
Generating static/coiled-upstream-py3.8.html
Generating static/coiled-latest-py3.10.html
Generating static/coiled-latest-py3.9.html
Generating static/coiled-latest-py3.8.html
Generating static/coiled-0.1.0-py3.10.html
Generating static/coiled-0.1.0-py3.9.html
Generating static/coiled-0.1.0-py3.8.html
Generating static/coiled-0.0.4-py3.10.html
Generating static/coiled-0.0.4-py3.9.html
Generating static/coiled-0.0.4-py3.8.html
Generating static/coiled-0.0.3-py3.9.html
Generating static/coiled-0.0.3-py3.8.html
Generating static/coiled-0.0.3-py3.7.html
Generating static/AB_by_test.html
Generating static/AB_by_runtime.html
  File "/home/runner/work/coiled-runtime/coiled-runtime/dashboard.py", line 497, in main
    align_to_baseline(latest_run, baseline),
  File "/home/runner/work/coiled-runtime/coiled-runtime/dashboard.py", line 80, in align_to_baseline
    df_baseline.set_index("fullname")
  File "/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/pandas/core/indexing.py", line 961, in __getitem__
    return self._getitem_tuple(key)
  File "/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/pandas/core/indexing.py", line 1147, in _getitem_tuple
    return self._multi_take(tup)
  File "/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/pandas/core/indexing.py", line 1098, in _multi_take
    d = {
  File "/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/pandas/core/indexing.py", line 1099, in <dictcomp>
    axis: self._get_listlike_indexer(key, axis)
  File "/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/pandas/core/indexing.py", line 1330, in _get_listlike_indexer
    keyarr, indexer = ax._get_indexer_strict(key, axis_name)
  File "/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 5796, in _get_indexer_strict
    self._raise_if_missing(keyarr, indexer, axis_name)
  File "/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 5859, in _raise_if_missing
    raise KeyError(f"{not_found} not in index")
KeyError: "['stability/test_shuffle.py::test_shuffle_simple', 'stability/test_array.py::test_rechunk_in_memory', 'stability/test_shuffle.py::test_shuffle_parquet'] not in index"
```

crusaderky:
> @crusaderky it looks like there is a legitimate error in the static site generation script

@jrbourbeau fixed now

crusaderky:
@ncclementi @ian-r-rose I called my latest commit `test-upstream`; the setup job outputs "true" in its `Check upstream` task; however the upstream runtime is not there. Am I missing something?

jrbourbeau:
I think things are working as expected. The `upstream` build should only install nightly versions of `dask` and `distributed` in the `latest` CI builds. For example, if you look at [the linux Python 3.9 `latest` build](https://github.com/coiled/coiled-runtime/runs/8237541073?check_suite_focus=true#step:5:20) on your latest commit (the `test-upstream` one), you'll see the the dev version of `dask` and `distributed` are installed from the [`dask` conda channel](https://anaconda.org/dask). Does that help clarify, or am I missing something?

ncclementi:
@crusaderky  this is a bit of a bad design in the name. What happened is that when you called test-upstream then "latest" will have the nightlies see that the nightlies where installed.  https://github.com/coiled/coiled-runtime/runs/8237542627?check_suite_focus=true#step:5:20

That logic is here https://github.com/coiled/coiled-runtime/blob/1614ab58128e3e3dcfa7d213db2c88e7c55ac6d4/ci/create_latest_runtime_meta.py#L41-L49

fjetter:
Re: noise of the samples
I believe if we have historic variances we can employ some statistical significance analysis on the new results. I believe this is something one typically uses a [t-test](https://en.wikipedia.org/wiki/Student%27s_t-test) for to verify whether the new measurement is in agreement to the old one. That's basically a very simple quantity `t = (old_measurement - new_measurement) / sqrt(historic_variance / n_historic_data_points)`. If that value is large, this is true signal (depending on how we define our [confidence interval](https://en.wikipedia.org/wiki/Student%27s_t-distribution#Table_of_selected_values)) @ncclementi I believe you wrote the regression detection script. You might have more insight here than I do.

We can also look into running some tests multiple times. Particularly the ones that have very small runtimes (<30s) are likely more prone to noise and we can afford running them a couple of times.

Obviously, all of this is a follow up topic. I think this is already a great first step

wence-:
> I believe if we have historic variances we can employ some statistical significance analysis on the new results. I believe this is something one typically uses a [t-test](https://en.wikipedia.org/wiki/Student%27s_t-test) for to verify whether the new measurement is in agreement to the old one. That's basically a very simple quantity `t = (old_measurement - new_measurement) / sqrt(historic_variance / n_historic_data_points)`.

A t-test is reasonable if your null hypothesis is that the data are normally distributed, if you don't have that guarantee (for example a long-tailed distribution due to outliers) then you're better off with a non-parametric test (such as [Kolmogorov-Smirnov](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) (available in scipy.stats.kstest)

jrbourbeau (commented):
From what I can tell this looks like it's ready for a final review. @ian-r-rose do you think you'll have time to take a look at the `dashboard.py` changes? Or maybe you already have

There's been some discussion around merge conflicts with https://github.com/coiled/coiled-runtime/pull/235 but from what I can tell it appears they will be relatively minimal, so I'm okay merging this PR first if needed

fjetter (approved):
I only reviewed the code on a high level. If the plots turn out this way, I'm good with this.

PR #310, opened 2022-09-06
Patch release 0.1.1 - bump dask and distributed to 2022.6.1

ncclementi:
PR to get started patch release - see issue #306

I added instructions for PyPI release too, not sure if there is another step here, please take a look in case I forgot something.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ not a personal fork**.
 
 To issue a new `coiled-runtime` release:
 
-1. Locally update the `coiled-runtime` version and package pinnings specified in `recipe/meta.yaml`.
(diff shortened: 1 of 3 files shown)


ntabris:
@jrbourbeau @ncclementi: out of curiosity, what's the status of this?

jrbourbeau (commented):
This came up in an offline meeting this morning. In summary, we decided to hold off on bumping `distributed` from `2022.6.0` to `2022.6.1` in favor of waiting a bit and jumping to a more recent release (e.g. `2022.9.1` -- assuming there isn't a blocking reason to pick that release). @hendrikmakait, as you brought this up originally in https://github.com/coiled/coiled-runtime/issues/306, please let me know if you have reservations around this plan

PR #312, opened 2022-09-07
Test CI 

jrbourbeau:
Testing to see if the issue in https://github.com/coiled/coiled-runtime/issues/311 has been resolved upstream


jrbourbeau:
Looks like the macOS build passed, so the `libprotobuf` issue identified in https://github.com/coiled/coiled-runtime/issues/311 has been resolved

PR #314, opened 2022-09-08
A/B test jobs

### LINKED ISSUE #292: Automation of benchmark comparison

crusaderky:
- Partially closes #292
- Complements #309

This PR adds the ability to run the whole test suite on arbitrary additional coiled software environments.

### DIFF
--- a/AB_environments/README.md
+++ b/AB_environments/README.md
@@ -1,0 +1,102 @@
+# A/B testing
+
+It's possible to run the Coiled Runtime benchmarks on A/B comparisons,
+highlighting performance differences between different released versions
+of dask, distributed, or any of the dependencies and/or between different
+dask configs.
+
+To run an A/B test:
+
+### 1. Create a new branch
+
+Branch from main, on the coiled repo itself. Preferably, call the branch
+something meaningful, e.g. `AB/jobstealing`.
+You *must* create the branch on the Coiled repo (`coiled/coiled-runtime`); CI
+workflows will not work on a fork (`yourname/coiled-runtime`).
+
+### 2. Create files in AB_environments/
+
+Open the `AB_environments/` directory and rename/create files as needed.
+Each A/B runtime is made of exactly two files:
+- `AB_<name>.conda.yaml` (a conda environment file)
+- `AB_<name>.dask.yaml` (a dask configuration file)
+
+You may create as many A/B runtime configs as you want in a single `coiled-runtime`
+branch.
+
+The conda environment file can contain whatever you want, as long as it can run the
+tests; e.g.
+
+```yaml
+channels:
+  - conda-forge
+dependencies:
+    - python=3.9
+    - coiled-runtime=0.1.0
+    - pip:
+      - dask==2022.8.1
+      - distributed=2022.8.1
+```
+In this example it's using `coiled-runtime` as a base, but it doesn't have to. If you do
+use `coiled-runtime` though, you must install any conflicting packages with pip; in the
+example above, `coiled-runtime-0.1.0` pins `dask=2022.6.0` and `distributed=2022.6.0`,
+so if you want to install a different version you need to use pip to circumvent the pin.
+
+Instead of published packages, you could also used arbitrary git hashes of
+arbitrary forks, e.g.
+
+```yaml
+    - pip:
+      - dask==2022.8.1
+      - git+https://github.com/yourname/distributed@dd81b424971e81616e1a52fa09ce4698a5002d41
+```
+The second file in each pair is a dask config file. If you don't want to change the
+config, you must create an empty file.
+
+e.g.
+```yaml
+distributed:
+  scheduler:
+    work-stealing: False
+```
+
+### 3. Create baseline files
+If you create *any* files in `AB_environments/`, you *must* create the baseline environment:
+
+- `AB_baseline.conda.yaml`
+- `AB_baseline.dask.yaml`
+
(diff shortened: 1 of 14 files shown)


AB_environments/AB_sample.dask.yaml
```diff
@@ -0,0 +1,7 @@
+# Sample dask config file for A/B testing
+# It *must* be called AB_<name>.dask.yaml
+# and *must* be accompanied by AB_<name>.conda.yaml
+# Leave empty if you don't want to override anything.
+distributed:
+  scheduler:
+    worker-saturation: 1.2
```
crusaderky, line 7: Notably, if #235 will go through we'll lose the ability to pass environment variables.

The alternative to forcing the developer to create an empty file every time was to ignore the file if it doesn't exist, but the risk of misspellings producing unexpected results was too high. Explicit is better than implicit.

crusaderky:
Ready for final review.
Run with files in `AB_environments/`: https://github.com/coiled/coiled-runtime/actions/runs/3018905647

crusaderky:
# Question:
In the context of A/B performance tests, do we care about the `runtime` and `stability` test folders?
CC @ncclementi @ian-r-rose

hayesgb:
I can see a solid argument for skipping the `runtime` folder, but seems to me it would preferable to keep the `stability` folder in scope.

ci/scripts/dask_config_to_env.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+from __future__ import annotations
+
+import sys
+from typing import Iterator
+
+import yaml
+
+
+def main(fname: str) -> None:
```
fjetter, line 14: I believe we're now able to ship dask config using the dask.config ctx manager. That might be a better interface than this. is there a reason why we use env vars instead?

fjetter (approved):
full disclosure: I haven't reviewed ab_tests in much detail but rather the broad strokes. 

My comments are mostly nitpicks. If there are issues or missing features we can iterate.

Looking forward to see this in action!

AB_environments/README.md
```diff
@@ -0,0 +1,99 @@
...
+channels:
+  - conda-forge
+dependencies:
+    - python=3.9
+    - coiled-runtime=0.1.0
+    - pip:
+      - dask==2022.8.1
+      - distributed=2022.8.1
+```
+- `AB_environments/AB_baseline.dask.yaml`: (empty file)
```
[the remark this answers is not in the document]
crusaderky: Yes. The alternative to forcing the developer to create an empty file every time was to ignore the file if it doesn't exist, but the risk of misspellings quietly resulting in unexpected results was too high. Explicit is better than implicit.

ci/scripts/dask_config_to_env.py
ntabris: > I believe we're now able to ship dask config using the dask.config ctx manager.

Just to be clear, change to ship dask config to cluster is not yet deployed, presumably will go out next week.

crusaderky: ? This is news to me.
Are you saying that when you call `coiled.create_software_environment` it will pick up the local config?
Also, I needed something that could be passed to `coiled env create`. Ideally one would want to pass it a dask config file directly, but I believe there's no such feature?

AB_environments/README.md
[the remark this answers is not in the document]
ntabris: Why not include empty file with appropriate name in repo? That would mean you'd get no custom config as the default, but also make it easier to not accidentally make a new file w/ wrong name (e.g., you'd either edit existing file or you'd use `cp your-file AB_[tab]` and autocomplete to correct filename).

ci/scripts/dask_config_to_env.py
ntabris: Now I'm confused. The feature is that when you create cluster it ships local config. I don't see any connection between dask config and the software environment... am I missing something?

crusaderky: > Just to be clear, change to ship dask config to cluster is not yet deployed, presumably will go out next week.

This is exactly what we need, please ping me when it's generally available

crusaderky: > Now I'm confused. The feature is that when you create cluster it ships local config. I don't see any connection between dask config and the software environment... am I missing something?

The script is currently calling `coiled env create -e DASK_DISTRIBUTED_...=VALUE`

ntabris: > The script is currently calling `coiled env create -e DASK_DISTRIBUTED_...=VALUE`

I could be wrong but I think that just applies when creating the software environment and not when you make a cluster using that software environment.

crusaderky: It works fine for me?
```python
import coiled
import dask.config
import distributed

!coiled env create --name crusaderky/test_vars --conda AB_environments/AB_baseline.conda.yaml -e DASK_TESTVAR=123
cluster = coiled.Cluster(name="test_vars", n_workers=0, software="crusaderky/test_vars")
client = distributed.Client(cluster)
client.run_on_scheduler(lambda: dask.config.get("testvar"))
```
```
123
```

ntabris: Oh, thanks, good to know!

PR #323, opened 2022-09-12
Fix `benchmark_task_durations` fixture when running without `--benchmark` option

### LINKED ISSUE #322: Using `benchmark_task_durations` fixture without `--benchmark` raises `AttributeError`
Running a test that uses the `benchmark_task_durations` without the `--benchmark` option raises `AttributeError: 'NoneType' object has no attribute 'compute_time'`.

**Reproducer**
```python
from distributed import Client

def test_benchmarking(benchmark_task_durations):
    with Client() as client:
        with benchmark_task_durations(client):
            pass
```

hendrikmakait:
* Closes #322

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -229,13 +229,15 @@ def auto_benchmark_time(benchmark_time):
 def benchmark_memory(test_run_benchmark):
     @contextlib.contextmanager
     def _benchmark_memory(client):
-        sampler = MemorySampler()
-        label = uuid.uuid4().hex[:8]
-        with sampler.sample(label, client=client, measure="process"):
-            yield
-
-        df = sampler.to_pandas()
         if test_run_benchmark:
+            yield
+        else:
+            sampler = MemorySampler()
+            label = uuid.uuid4().hex[:8]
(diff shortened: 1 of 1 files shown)


ian-r-rose:
Thanks! I was just about to raise the same issue :)

It looks like the memory benchmarking fixture has the same problem. Mind doing that here while you're at it?

conftest.py
```diff
@@ -229,13 +229,15 @@ def auto_benchmark_time(benchmark_time):
...
+        if test_run_benchmark:
             yield
+        else:
+            sampler = MemorySampler()
+            label = uuid.uuid4().hex[:8]
+            with sampler.sample(label, client=client, measure="process"):
+                yield
 
-        df = sampler.to_pandas()
-        if test_run_benchmark:
```
hendrikmakait, line 238: Technically, this line prevented `benchmark_memory` from raising, but I've adopted the same pattern we use in the other fixtures to reduce cognitive load. @ian-r-rose

ian-r-rose (approved):
Thanks @hendrikmakait

PR #326, opened 2022-09-13
Add fixture docstrings

### LINKED ISSUE #318: Document benchmark fixtures with docstrings
By only documenting the benchmark fixtures in the `README.md`, we have separated the documentation from the code they describe. This makes the description harder to discover and maintain. I suggest moving the documentation into docstrings and have `README.md` refer to the documentation with `conftest.py`.

hendrikmakait:
* Closes #318

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,31 +116,17 @@ Migrations are automatically applied in the pytest runs, so you needn't run them
 ### Using the benchmark fixtures
 
 We have a number of pytest fixtures defined which can be used to automatically track certain metrics in the benchmark database.
-They are summarized here:
+The most relevant ones are summarized here:
 
-**`benchmark_db_engine`**: The SQLAlchemy engine for the benchmark sqlite database. You can control the database name with the environment variable `DB_NAME`, which defaults to `benchmark.db`. Most tests shouldn't need to include this fixture directly.
+**`benchmark_time`**: Record wall clock time duration.
 
-**`benchmark_db_session`**: The SQLAlchemy session for a given test. Most tests shouldn't need to include this fixture directly.
+**`benchmark_memory`**: Record memory usage.
 
-**`test_run_benchmark`**: The SQLAlchemy ORM object for a given test. By including this fixutre in a test (or another fixture that includes it) you trigger the test being written to the benchmark database. This fixture includes data common to all tests, including python version, test name, and the test outcome.
+**`benchmark_task_durations`**: Record time spent computing, transferring data, spilling to disk, and deserializing data.
 
-**`benchmark_time`**: Include this fixture to measure the wall clock time.
+**`benchmark_all`**: Record all available metrics.
 
-**`benchmark_memory`**: This fixture yields a context manager which takes a distributed `Client` object, and records peak and average memory usage for the cluster within the context:
-```python
-def test_something(benchmark_memory):
-    with Client() as client:
-        with benchmark_memory(client):
-            client.submit(expensive_function)
-```
-
-**`benchmark_task_durations`**: This fixture yields a context manager which takes a distributed `Client` object, and records time spent computing, transferring data, spilling to disk, and deserializing data within the context:
-```python
-def test_something(benchmark_task_durations):
-    with Client() as client:
-        with benchmark_task_durations(client):
-            client.submit(expensive_function)
-```
+For more information on all available fixtures and examples on how to use them, please refer to their documentation.
 
 Writing a new benchmark fixture would generally look like:
 1. Requesting the `test_run_benchmark` fixture, which yields an ORM object.
--- a/conftest.py
+++ b/conftest.py
@@ -133,6 +133,17 @@ else:
 
 @pytest.fixture(scope="session")
 def benchmark_db_engine(pytestconfig, tmp_path_factory):
+    """Session-scoped fixture for the SQLAlchemy engine for the benchmark sqlite database.
+
+    You can control the database name with the environment variable ``DB_NAME``,
+    which defaults to ``benchmark.db``. Most tests shouldn't need to include this
+    fixture directly.
+
+    Yields
+    ------
+    The SQLAlchemy engine if the ``--benchmark`` option is set, None otherwise.
+    """
+
     if not pytestconfig.getoption("--benchmark"):
         yield
     else:
@@ -160,6 +171,15 @@ def benchmark_db_engine(pytestconfig, tmp_path_factory):
 
 @pytest.fixture(scope="function")
 def benchmark_db_session(benchmark_db_engine):
+    """SQLAlchemy session for a given test.
+
+    Most tests shouldn't need to include this fixture directly.
+
+    Yields
+    ------
+    SQLAlchemy session for the benchmark database engine if run as part of a benchmark,
+    None otherwise.
+    """
     if not benchmark_db_engine:
         yield
     else:
@@ -169,6 +189,18 @@ def benchmark_db_session(benchmark_db_engine):
 
 @pytest.fixture(scope="function")
 def test_run_benchmark(benchmark_db_session, request, testrun_uid):
+    """SQLAlchemy ORM object representing a given test run.
+
+    By including this fixture in a test (or another fixture that includes it)
+    you trigger the test being written to the benchmark database. This fixture
+    includes data common to all tests, including python version, test name,
+    and the test outcome.
+
+    Yields
+    ------
+    SQLAlchemy ORM object representing a given test run if run as part of a benchmark,
+    None otherwise.
+    """
     if not benchmark_db_session:
         yield
     else:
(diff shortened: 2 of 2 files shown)


conftest.py
```diff
@@ -133,6 +133,17 @@ def get_coiled_software_name():
 
 @pytest.fixture(scope="session")
 def benchmark_db_engine(pytestconfig, tmp_path_factory):
+    """Session-scoped fixture for the SQLAlchemy engine for the benchmark sqlite database.
+
+    You can control the database name with the environment variable ``DB_NAME``,
+    which defaults to ``benchmark.db``. Most tests shouldn't need to include this
```
ncclementi, line 139: When we say "most tests shouldn't need..." I'm assuming these are the tests using the `small_client` fixture right? If so, should say that? I think the way this sentence is, it is not clear when you need it or when you do not.

conftest.py
```diff
@@ -160,6 +171,15 @@ def benchmark_db_engine(pytestconfig, tmp_path_factory):
 
 @pytest.fixture(scope="function")
 def benchmark_db_session(benchmark_db_engine):
+    """SQLAlchemy session for a given test.
+
+    Most tests shouldn't need to include this fixture directly.
```
ncclementi, line 176: same comment as above,

README.md
```diff
@@ -116,31 +116,17 @@ Migrations are automatically applied in the pytest runs, so you needn't run them
 ### Using the benchmark fixtures
 
 We have a number of pytest fixtures defined which can be used to automatically track certain metrics in the benchmark database.
-They are summarized here:
+The most relevant ones are summarized here:
 
-**`benchmark_db_engine`**: The SQLAlchemy engine for the benchmark sqlite database. You can control the database name with the environment variable `DB_NAME`, which defaults to `benchmark.db`. Most tests shouldn't need to include this fixture directly.
+**`benchmark_time`**: Record wall clock time duration.
```
ncclementi, line 121:
```suggestion
**`benchmark_time`**: Records wall clock time duration.
```

README.md
```diff
@@ -116,31 +116,17 @@ Migrations are automatically applied in the pytest runs, so you needn't run them
...
 
-**`benchmark_db_session`**: The SQLAlchemy session for a given test. Most tests shouldn't need to include this fixture directly.
+**`benchmark_memory`**: Record memory usage.
```
ncclementi, line 123:
```suggestion
**`benchmark_memory`**: Records peak and average memory usage.
```

README.md
```diff
@@ -116,31 +116,17 @@ Migrations are automatically applied in the pytest runs, so you needn't run them
...
 
-**`test_run_benchmark`**: The SQLAlchemy ORM object for a given test. By including this fixutre in a test (or another fixture that includes it) you trigger the test being written to the benchmark database. This fixture includes data common to all tests, including python version, test name, and the test outcome.
+**`benchmark_task_durations`**: Record time spent computing, transferring data, spilling to disk, and deserializing data.
```
ncclementi, line 125:
```suggestion
**`benchmark_task_durations`**: Records time spent computing, transferring data, spilling to disk, and deserializing data.
```

README.md
```diff
@@ -116,31 +116,17 @@ Migrations are automatically applied in the pytest runs, so you needn't run them
...
 
-**`benchmark_time`**: Include this fixture to measure the wall clock time.
+**`benchmark_all`**: Record all available metrics.
```
ncclementi, line 127:
```suggestion
**`benchmark_all`**: Records all available metrics.
```

conftest.py
```diff
@@ -204,6 +236,26 @@ def test_run_benchmark(benchmark_db_session, request, testrun_uid):
 
 @pytest.fixture(scope="function")
 def benchmark_time(test_run_benchmark):
+    """Benchmark the wall clock time of executing some code.
```
ncclementi, line 239:
```suggestion
    """Benchmarks the wall clock time of executing some code.
```

conftest.py
```diff
@@ -221,12 +273,49 @@ def _benchmark_time():
 
 @pytest.fixture(scope="function")
 def auto_benchmark_time(benchmark_time):
+    """Benchmark the wall clock time of an individual test.
```
ncclementi, line 276:
```suggestion
    """Benchmarks the wall clock time of an individual test.
```

conftest.py
```diff
@@ -221,12 +273,49 @@ def _benchmark_time():
...
+
+    This fixture automatically records the wall clock time it takes to
+    execute an individual test if run as part of a benchmark. Depending on the test's structure, this may
+    include setup or teardown code that should not be measured. For more
+    control, use the ``benchmark_time`` context manager fixture.
```
ncclementi, line 281: It seems we use this fixture only once, in `test_default_cluster_spinup_time` I wonder if we should just not use it. It's out of the scope of this PR, but any thoughts?

conftest.py
```diff
@@ -221,12 +273,49 @@ def _benchmark_time():
...
+    --------
+    benchmark_time
+    """
     with benchmark_time:
         yield
 
 
 @pytest.fixture(scope="function")
 def benchmark_memory(test_run_benchmark):
+    """Benchmark the memory usage of executing some code.
```
ncclementi, line 301:
```suggestion
    """Benchmarks the memory usage of executing some code.
```

conftest.py
```diff
@@ -246,6 +335,26 @@ def _benchmark_memory(client):
 
 @pytest.fixture(scope="function")
 def benchmark_task_durations(test_run_benchmark):
+    """Benchmark the time spent computing, transferring data, spilling to disk,
```
ncclementi, line 338:
```suggestion
    """Benchmarks the time spent computing, transferring data, spilling to disk,
```

conftest.py
```diff
@@ -274,13 +383,28 @@ def _measure_durations(client):
...
 @pytest.fixture(scope="function")
 def benchmark_all(benchmark_memory, benchmark_task_durations, benchmark_time):
-    """Return a function that creates a context manager for benchmarking.
-
-    Example:
-    >>> def test_example(benchmark_all):
-    >>>    ...
-    >>>    with benchmark_all(client):
-    >>>       client.compute(my_computation)
+    """Benchmark all available metrics.
```
ncclementi, line 386:
```suggestion
    """Benchmarks all available metrics.
```

ncclementi (commented):
I left few comments, and some small suggestions. But overall this looks  good to me.

README.md
hendrikmakait: Given the argumentation regarding [PEP 257](https://peps.python.org/pep-0257/#one-line-docstrings) in my other comment, I'd prefer sticking with a coherent style.

conftest.py
hendrikmakait: I think it's any test that uses one of the "user-facing" benchmarking fixtures, most notably `benchmark_all` as the catch-all solution. See https://github.com/coiled/coiled-runtime/blob/main/tests/stability/test_spill.py#L11 for example. I'd probably go as far as making this a private fixture and saying that it's for internal use and should not be used in tests.

conftest.py
hendrikmakait: I don't really have any strong feelings here, but I think that you're right that we may want to drop this to avoid people using it. I guess in 99% of the cases we want people to use the `benchmark_all` or `small_client` fixtures`. This is work for a follow-up PR, but I think `small_client` could also use some improved documentation, in particular we may want to highlight that it already uses the `benchmark_all` fixture.

conftest.py
ncclementi: Good to know, I was not aware of this. Thanks for the clarifications.

hendrikmakait: FWIW, I was just made aware by @ntabris that the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) recommends the opposite:
> The docstring should be descriptive-style ("""Fetches rows from a Bigtable.""") rather than imperative-style ("""Fetch rows from a Bigtable.""")

So both styles are being used in different places. Since https://numpydoc.readthedocs.io/en/latest/format.html refers to PEP 257 but not Google, I'd give that one precedence.

conftest.py
```diff
@@ -221,12 +273,49 @@ def _benchmark_time():
...
+    as input. The context manager records peak and average memory usage of
+    executing the ``with`` statement if run as part of a benchmark,
+    or does nothing otherwise.
+
+    Example
+    -------
+    .. code-block:: python
+
+        def test_something(benchmark_memory):
+            with benchmark_memory(client):
```
ian-r-rose, line 315: Here and below: you restore the `client` creation as in the README to make the example more explicit? Otherwise, no objection from me.

ncclementi:
I see there are some regressions reported, but they can't be related to this PR as it is only docs, I'll go ahead and merge this, Thanks @hendrikmakait .

ian-r-rose (approved):
Thanks @hendrikmakait

PR #327, opened 2022-09-13
[DNM] Test id3 no categorical - h2o benchmarks

ncclementi:
Testing h2o benchmark runs after re-generating the data avoiding categorical dtypes for id3. See https://github.com/coiled/h2o-benchmarks/issues/11 for reference.

### DIFF
--- a/tests/benchmarks/h2o/test_h2o_benchmarks.py
+++ b/tests/benchmarks/h2o/test_h2o_benchmarks.py
@@ -10,15 +10,15 @@ import pytest
 @pytest.fixture(
     scope="module",
     params=[
-        "s3://coiled-datasets/h2o-benchmark/N_1e7_K_1e2_single.csv",
+        # "s3://coiled-datasets/h2o-benchmark/N_1e7_K_1e2_single.csv",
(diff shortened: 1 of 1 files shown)


ncclementi:
By looking at the q3 runs, I was expecting for them to be faster with the change of the data but I do not notice this. See https://github.com/coiled/coiled-runtime/actions/runs/3048664324/jobs/4914069085#step:6:195 . The run took around 1 min for the 5GB which is what we are currently seeing in the main. 

I'm looking only at runtime 0.0.4 and 0.1.0. Should this change only matter in newer versions of dask? Or maybe at this point is not big enough data?
cc: @jrbourbeau

ian-r-rose:
I'd expect this to matter most for the 50 GB case. It's also a place where we would probably want to try the shuffle-based groupby/agg (which should only be available in `latest` or `upstream`)

ncclementi:
Yes, probably this is best to test out of this repo. I suspect if I run the 50GB queries we will hit the time limits set on the clusters. I'll get the performance reports over in the h2o-repo.

PR #338, opened 2022-09-16
Less cluster memory

fjetter:
This will run the array workloads that accept cluster memory as an input parameter parametrized for a low-memory and high-memory situation.

The tests themselves sometimes still multiply this value but I didn't want to modify test logic. Further down the road, it will surely be an interesting exercise to increase this value to >1.0 as well but for now we're mostly concerned getting the low mem cases right.

For future benchmark development I believe the pattern of adjusting used data size as a function of cluster memory is a good practice that allows us scaling these tests very easily


fjetter:
Failure on the win stability test https://github.com/coiled/feedback/issues/204

fjetter:
Failure on ubu 3.9 upstream is https://github.com/coiled/coiled-runtime/issues/336

I have no idea how this change could introduce regressions. The one test I touched that is flagged as a regression is `test_dataframe_align` but the regression is detected for coiled-runtime 0.0.4

I am surprised that this even kicks in. Given the parametrization I'm wondering how the script correlates the new changes with the old ones. Any idea @ian-r-rose ?

fjetter:
The visualization on this is a bit unfortunate. Could be fixed with a DB migration but not sure if that's worth it
image

alembic/versions/2f28d2536c90_parametrize_over_cluster_memory.py
```diff
@@ -0,0 +1,37 @@
...
+revision = '2f28d2536c90'
+down_revision = '2764a4f5582b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+    """update test_run
+    set name = name || '[100% cluster memory]'
```
fjetter, line 22: @ian-r-rose I'm not entirely sure what `name` and `orignalname` is. is it correct for me to just modify `name`?

ian-r-rose: `originalname` is the name of the actual function, whereas `name` is the full name with any parametrizations tacked on (this is pytest terminology). So the structure of this migration looks correct to me.

conftest.py
```diff
@@ -435,6 +435,25 @@ def small_client(
             yield client
 
 
+def _cluster_memory(client: distributed.Client) -> int:
+    "Total memory available on the cluster, in bytes"
+    return int(
+        sum(w["memory_limit"] for w in client.scheduler_info()["workers"].values())
+    )
```
ian-r-rose, line 442: Seems unfortunate to delete the utility function such that it's inaccessible to tests who might want to do something different with the cluster memory measurement.

fjetter: right, I'll just rename it to `get_cluster_memory`

fjetter:
"proof" that the migration script works
image

ian-r-rose:
> I am surprised that this even kicks in. Given the parametrization I'm wondering how the script correlates the new changes with the old ones. Any idea @ian-r-rose ?

It looks like you've worked this out, but for posterity, the correlation with older runs is based on matching the `name` (not `originalname`, which does not include parameterizations) and the runtime version.

ian-r-rose:
> "proof" that the migration script works image

Nice

conftest.py
```diff
@@ -435,6 +435,25 @@ def small_client(
...
+def get_cluster_memory(client: distributed.Client) -> int:
+    "Total memory available on the cluster, in bytes"
+    return int(
+        sum(w["memory_limit"] for w in client.scheduler_info()["workers"].values())
+    )
+
+
+@pytest.fixture(
+    params=[
+        pytest.param(0.3, id="30% cluster memory"),
```
fjetter, line 447: 30% is just a guess. I'm looking for stuff that "comfortably fits into memory". I figured that if a test actually generates that much data, we could hold two full copies before starting to spill.

This will be individual for every test though since most tests still multiply the cluster-memory by a certain factor etc.

fjetter:
@ian-r-rose (or anybody else) if you are happy with this, please go ahead and merge since I'm likely already out before CI finishes but I think we should get this in before the weekend

ian-r-rose (approved):

jrbourbeau (commented):
I've only taken a quick look at this -- but I like the idea of quickly being able to say "run this test with small, large, etc. datasets compared to the cluster memory" 👍

tests/benchmarks/test_array.py
```diff
@@ -48,35 +47,34 @@ def test_anom_mean(small_client):
...
     wait(result, small_client, 10 * 60)
 
 
-@pytest.mark.skip(
-    "fails in actual CI; see https://github.com/coiled/coiled-runtime/issues/253"
-)
-def test_climatic_mean(small_client):
+def test_climatic_mean(small_client, cluster_memory):
     # From https://github.com/dask/distributed/issues/2602#issuecomment-535009454
+    if cluster_memory == 1.0:
```
gjoseph92, line 68: Why would it ever be 1.0? The actual cluster memory would have to be 1 byte for that to happen. I assume you want to detect the not-30% case, but because you only have access to the (faked) size in bytes, not the factor, that's tricky to do. (See above for discussion.)

tests/benchmarks/test_array.py
```diff
@@ -22,16 +22,15 @@ def print_size_info(memory: int, target_nbytes: int, *arrs: da.Array) -> None:
...
 
-    memory = cluster_memory(small_client)  # 76.66 GiB
-    target_nbytes = memory // 2
+    target_nbytes = cluster_memory // 2
     data = da.random.random(
         scaled_array_shape(target_nbytes, ("x", "10MiB")),
         chunks=(1, parse_bytes("10MiB") // 8),
     )
-    print_size_info(memory, target_nbytes, data)
+    print_size_info(cluster_memory, target_nbytes, data)
```
gjoseph92, line 33: nit: I don't love implementing this by lying to the test about the amount of cluster memory. It makes this logging of size info wrong, for one. It also means that tests that do want to adapt to cluster memory, but don't make sense to run on a variety of parameterizations (maybe they're meant to test spilling, or they know they won't work on larger datasets [see comment below]), are difficult to write.

Instead, why not have `memory_factor` be the fixture, and multiply within the test when appropriate? A little more code, but also more explicit and readable.

gjoseph92 (commented):
> I like the idea of quickly being able to say "run this test with small, large, etc. datasets compared to the cluster memory"

I like this idea as well, but are we sure it's valuable to do automatically on every case? I appreciate setting something up like this so we can use it conveniently in the future, but does it add a valuable enough signal right now to be worth increasing test runtime this much? I could see merging this, but on main only having a single parametrization of `memory_factor`.

tests/benchmarks/test_array.py
fjetter: I think this originates from an earlier iteration. Will need to fix that

tests/benchmarks/test_array.py
hayesgb: +1

hayesgb (commented):
My only question here is whether it makes sense to use the same data sizes for all of the array tests.

Was there a specific reason the designated multiples were chosen for each test.  I think @gjoseph92 may have better context.  See #410 as alternative.

gjoseph92:
> Was there a specific reason the designated multiples were chosen for each test

Absolutely. These multiples are are based on actual workloads reported by actual users relative to the size of their actual cluster.

As an example, `test_anom_mean` came from https://github.com/dask/distributed/issues/2602#issuecomment-498718651, which was sometimes unrunnable using a 75 GiB dataset on a 200GiB cluster (see https://github.com/dask/distributed/issues/2602#issuecomment-870930134). Increasing cluster size significantly would make it runnable. So this test was tuned to trigger the real-world case of this workload with data ~half the size of the cluster, which presumably used to trigger spilling and then failure. Making the data significantly smaller means we'd expect it to finish even if scheduling was horrendous. Making the data significantly bigger means I'm not sure we could reasonably expect it to be runnable.

So adding a multiple to every test's data size might kinda make sense, but we shouldn't just change how all the data sizes are already set up relative to cluster memory—just apply the multiple at the end.

PR #341, opened 2022-09-16
Move test_default_cluster_spinup_time

crusaderky:
`test_default_cluster_spinup_time` takes 300 to 800s to run. While it's running, `small_cluster` remains active and idle.

Exclude the test from A/B tests, which are about different versions of dask or coiled-runtime (coiled is not pinned in coiled-runtime), to save time and money, by moving it from the benchmarks category to runtime.


ncclementi:
This makes sense to me. But I believe that we will have the existing data on the database now being stale, as this test will move to a different page. Is that right @ian-r-rose ? If so, should we declare bankruptcy or just leave the old data there?

ian-r-rose:
>  While it's running, small_cluster remains active and idle.

I'm not sure what you mean by this. `small_cluster` is run in pytest-xdist with `"loadscope"`, so I'd expect it to be fairly independent of the rest of the test suite.

I don't really have any objection to moving the test, I'm just not sure it makes a difference one way or another.

ian-r-rose:
> But I believe that we will have the existing data on the database now being stale, as this test will move to a different page. Is that right @ian-r-rose ? If so, should we declare bankruptcy or just leave the old data there?

That's right, though I just declared bankruptcy on this test a couple of days ago, so this would be a great time to move it if we wanted to :)

crusaderky:
> > While it's running, small_cluster remains active and idle.
> 
> I'm not sure what you mean by this. `small_cluster` is run in pytest-xdist with `"loadscope"`, so I'd expect it to be fairly independent of the rest of the test suite.

What I mean is that today I ran 2 A/B branches, totalling 81 jobs.
I disabled runtime and stability, only leaving benchmarks, and I manually deleted `test_default_cluster_spinup_time` in my branch.

Had I not deleted it, then Coiled would have paid

- 81 x 300s x (1 scheduler + 4 workers) t3.xlarge, for the test itself
- 81 x 300s x [1 t3.xlarge scheduler + 10 t3-large workers], because `small_cluster` was started by another test and, while `test_default_cluster_spinup_time`, is just sitting there idle

Total 
81 x 300s x 5 x $0.0832/h (t3.xlarge)
81 x 300s x 10 x $0.1664/h (t3.large)
= $14 (17c per run)

This is not counting the times where it takes more than 300s, which are fairly frequent (I saw up to 800s).

ian-r-rose:
> 81 x 300s x [1 t3.xlarge scheduler + 10 t3-large workers], because small_cluster was started by another test and, while test_default_cluster_spinup_time, is just sitting there idle

This is the part I don't understand. These tests should be independent of each other, and `small_cluster` is not used here. What makes you say that this cluster is idling and not shutting down after the test is done?

crusaderky:
Ah, I just saw that 
```python
@pytest.fixture(scope="module")
def small_cluster(request):
```
so the cluster is shut down at the end of every test module.
I was convinced it had scope `package` or `session`, which would keep it idling while unrelated test modules are running.
Ironically, now it's been moved to a test file which contains tests that use the `small_client` fixture, so if you enable `runtime` you will actually spend more money than before.

Should I move it to its own module?

ian-r-rose:
Ah, good point. We can move it to it's own module, sure.

PR #342, opened 2022-09-16
Use t3.xlarge as default schdeduler vm type

### LINKED ISSUE #332: bump up scheduler vm size
Lots of test use cluster with `t3.large` scheduler. This has 2 vCPUs, which in this case means single physical core. I strongly recommend using `t3.xlarge` instead, this will reduce potential contention between dask and system + monitoring agents.

(That said, it will also affect your longitudinal comparisons.)

ian-r-rose:
Fixes #332

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -17,17 +17,12 @@ import distributed
 import filelock
 import pytest
 import s3fs
+import sqlalchemy
+from coiled import Cluster
 from distributed import Client
 from distributed.diagnostics.memory_sampler import MemorySampler
-from toolz import merge
-
-try:
-    from coiled.v2 import Cluster
-except ImportError:
-    from coiled._beta import ClusterBeta as Cluster
-
-import sqlalchemy
 from sqlalchemy.orm import Session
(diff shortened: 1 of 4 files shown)


ncclementi (commented):
I assume we are ok with tets_parquet.py still having an `m5.xlarge` or should we moved these ones to t3.xlarge? 

https://github.com/coiled/coiled-runtime/blob/f9f8d9b7a81927a3ecbfee746c251a1fc88075d8/tests/benchmarks/test_parquet.py#L17-L26

Outside that comment, everything else LGTM.

ian-r-rose:
I'm okay with it. We could switch them, at the cost of losing history, but I don't view it as a problem today.

ncclementi:
M5s seem to have better network bandwidth, probably that's the reason why you chose them in the first place. Let's keep it as is then.

ian-r-rose:
> M5s seem to have better network bandwidth, probably that's the reason why you chose them in the first place. Let's keep it as is then.

Ah, yeah,

PR #351, opened 2022-09-19
Unskip `test_work_stealing_on_scaling_up`

jrbourbeau:
Let's try un`skip`ping `test_work_stealing_on_scaling_up` since a possible fix has been deployed on Coiled (xref https://github.com/coiled/coiled-runtime/issues/336#issuecomment-1249955112) 

Closes https://github.com/coiled/coiled-runtime/issues/336

### DIFF
--- a/tests/benchmarks/test_work_stealing.py
+++ b/tests/benchmarks/test_work_stealing.py
@@ -1,11 +1,13 @@
 import time
 
 import dask.array as da
+import distributed
 import numpy as np
 import pytest
 from coiled import Cluster
 from dask import delayed, utils
 from distributed import Client
+from packaging.version import Version
 from tornado.ioloop import PeriodicCallback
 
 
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Okay, looks like we're still running into issues with `test_work_stealing_on_scaling_up`. We're seeing

```python
  +-------------+----------+-----------+----------+
  | Package     | client   | scheduler | workers  |
  +-------------+----------+-----------+----------+
  | dask        | 2022.9.1 | 2022.6.0  | 2022.6.0 |
  | distributed | 2022.9.1 | 2022.6.0  | 2022.6.0 |
  +-------------+----------+-----------+----------+
```

in CI still. cc @ntabris @shughes-uk for visibility

shughes-uk:
I’ll put this at the top of my list to look at, do you have a link to a
specific action run?

ian-r-rose:
Here is the relevant run: https://github.com/coiled/coiled-runtime/actions/runs/3083687439/jobs/4984924644

shughes-uk:
Package sync seems to be not enabled?

ian-r-rose:
Hah, right you are! Seems like it got caught up in a merge issue. I'll push here to verify

shughes-uk:
> Thanks @ian-r-rose. I totally missed that.
> 
> @shughes-uk what are your thoughts on adding a `package_sync` config value to `coiled`? That way we could easily say "Coiled, please always use package sync"

Seems reasonable, i'll add it to the todo list

PR #324, opened 2022-09-12
Repeat A/B tests

crusaderky:

Issue #279, opened 2022-08-25
Simplify CI test workflow


ian-r-rose:
We currently run the submodules in the `tests/` directory in separate jobs:

* benchmarks
* runtime
* stability

Each of those jobs has its own test matrix, so we actually wind up having a large number of sub-jobs for every workflow run.

I'm not convinced that is serving us well today:
1. Each job is parallelized using `pytest-xdist` with ten workers. That makes a lot of sense, since the vast majority of the work is done by coiled clusters, and the CI runner is mostly sitting around waiting for results. However,  we aren't getting the full value out of `pytest-xdist` because we are running it with `loadscope`, so each test module is run as a single unit. This basically means we don't have enough modules to actually saturate the xdist workers.
2. The different jobs have different test matrices. In particular, `"stability"` has the largest matrix, testing a few combinations that are not tested in others. I *think* this is because, at the time of writing, `"stability"` was the fastest submodule to run, so we could afford the extra matrices (is this right @jrbourbeau?). But since #229 the stability tests are quite slow. @hendrikmakait may have ideas to speed it up a bit, but I still am not convinced it's a good idea to have different test matrices for different submodules.
3. There is a lot of code duplication in the workflow yaml for each submodule, which adds cognitive overhead.

I propose we consolidate the above jobs into a single "test" job, which is itself parameterized over a consistent test matrix. I doubt that would actually change the runtime too much, and might even speed it up, as the xdist workers would be better able to balance work. It would also simplify maintenance going forward.

PR #355, opened 2022-09-20
A/B tests with package sync + repeats

crusaderky:
- Supersedes #324 
- Repair A/B tests after #235 
- Add throttling to prevent AWS DoS protection from kicking in
- Refactor regular (non-A/B) tests
- Introduce `repeat: N` setting, which causes every A/B test runtime to be rerun N times
- Overhaul the A/B performance reports to display statistical data instead of exact
- Introduce `test_null_hypothesis: true` setting, which creates a verbatim clone of AB_baseline

Out of scope:
- Merge tests.yaml with ab_tests.yaml. This will happen in a future PR.


crusaderky:
A/B tests evidence: https://github.com/coiled/coiled-runtime/actions/runs/3091876943

crusaderky:
This is ready for review

.github/workflows/tests.yml
```diff
@@ -22,19 +22,65 @@ defaults:
     shell: bash -l {0}
 
 jobs:
-  runtime:
-    name: Runtime - ${{ matrix.os }}, Python ${{ matrix.python-version }}, Runtime ${{ matrix.runtime-version }}
+  tests:
```
ian-r-rose, line 25: Over in #279 I proposed doing something similar to this, but without a new `category` matrix item. What would you think about just running the tests as a single job, and letting xdist do the rest?

ian-r-rose (commented):
Most of these changes look reasonable to me, though I haven't gone through in detail. My main question is whether we should just further simplify the tests workflow matrix rather than pushing these lumpy `include` blocks around.

.github/workflows/tests.yml
crusaderky: I'd say it's reasonable, and it would save some time. Some non-trivial engineering is needed - mind if I do it in a successive PR?

.github/workflows/tests.yml
```diff
@@ -22,19 +22,65 @@ defaults:
...
     runs-on: ${{ matrix.os }}
     timeout-minutes: 120
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest"]
+        os: [ubuntu-latest]
         python-version: ["3.9"]
-        runtime-version: ["upstream", "latest", "0.0.4", "0.1.0"]
+        category: [runtime, benchmarks, stability]
```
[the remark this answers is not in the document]
crusaderky: Even if you run all tests together, you'll still need a wordy `include` paragraph.

Instead of this:
```yaml
      matrix:
        os: [ubuntu-latest]
        python-version: ["3.9"]
        category: [runtime, benchmarks, stability]
        runtime-version: [upstream, latest, "0.0.4", "0.1.0"]
        include:
          # Run stability tests on Python 3.8
          - category: stability
            python-version: "3.8"
            runtime-version: upstream
            os: ubuntu-latest
       ...
```

it will look like this:
```yaml
      matrix:
        os: [ubuntu-latest]
        python-version: ["3.9"]
        test_dirs: [tests]
        runtime-version: [upstream, latest, "0.0.4", "0.1.0"]
        include:
          # Run stability tests on Python 3.8
          - test_dirs: tests/stability
            python-version: "3.8"
            runtime-version: upstream
            os: ubuntu-latest
         ...
```

In the next PR, I want to merge ab_tests.yaml into tests.yaml. In that PR I'll dynamically generate the whole matrix with `discover_ab_environments.py` (to be renamed); the matrix for non-A/B tests will be generated from parameters in `ci/config.yaml` (now `AB_environments/config.yaml`)

.github/workflows/tests.yml
ian-r-rose: Sure, I don't think anything here makes that refactor harder.

.github/workflows/tests.yml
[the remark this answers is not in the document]
ian-r-rose: Yeah, I was proposing just running the whole test suite on every matrix value. Perhaps overkill, however.

> In the next PR, I want to merge ab_tests.yaml into tests.yaml. In that PR I'll dynamically generate the whole matrix with `discover_ab_environments.py` (to be renamed); the matrix for non-A/B tests will be generated from parameters in `ci/config.yaml` (now `AB_environments/config.yaml`)

I'm a little concerned about the complexity of generating bespoke test matrices, and what it will mean for local testing. I was hoping to make the test matrix way *simpler*.

ian-r-rose (approved):
I haven't gone through in great detail, but this looks good to me from a high level

PR #358, opened 2022-09-21
Workers may not be online when Cluster(wait_for_workers=True) returns

crusaderky:
- Work around https://github.com/coiled/platform/issues/103


ntabris (approved):
Looks good

PR #359, opened 2022-09-21
Move `test_default_cluster_spinup_time` to its own module

crusaderky:
Follow-up from https://github.com/coiled/coiled-runtime/pull/341

### DIFF
--- a/tests/runtime/test_cluster_creation.py
+++ b/tests/runtime/test_cluster_creation.py
@@ -1,0 +1,15 @@
+import uuid
+
+from coiled import Cluster
+
+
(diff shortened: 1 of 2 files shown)


jrbourbeau (approved):
Thanks @crusaderky LGTM

@ian-r-rose there are lots of regression here -- are these "known" regressions or are these somehow new?

ian-r-rose:
> @ian-r-rose there are lots of regression here -- are these "known" regressions or are these somehow new?

These mostly look like ones that are known to have a lot of variability. We should look into the regression calculation and try to decrease the sensitivity.

PR #369, opened 2022-09-23
Generate A/B static dashboard on partial failures

### LINKED ISSUE #353: Generate performance report on failure
Tests and A/B tests are not 100% stable, due to an assortment of random failures.

The performance report artifact is currently generated exclusively if all tests are green. I suspect this is a bug in github actions, as the job is tagged with
```yaml
if: always() && ${{ fromJson(needs.discover_ab_envs.outputs.matrix).runtime }}
```
According to the documentation, the `if` clause is automatically rewritten as `success() && {user clause}` unless the user clause itself already contains `always()`, `success()`, or `failure()`. This is not the case in practice, which indicates that something's wrong is going on with the parser.

crusaderky:
Closes #353 

Tested with
```python
def test_random_fail(small_client):
    import random
    import time

    assert random.random() < 0.5
    time.sleep(5)
```
Output:
image
image
image

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -648,8 +648,12 @@ def main() -> None:
     for runtime in runtimes:
         make_timeseries_html_report(df, output_dir, runtime)
 
-    # Do not use data that is more than a week old in statistical analysis
-    df_recent = df[df["end"] > df["end"].max() - pandas.Timedelta("7d")]
+    # Do not use data that is more than a week old in statistical analysis.
+    # Also exclude failed tests.
+    df_recent = df[
+        (df["end"] > df["end"].max() - pandas.Timedelta("7d"))
+        & (df["call_outcome"] == "passed")
+    ]
 
     make_barchart_html_report(df_recent, output_dir, by_test=True)
     make_barchart_html_report(df_recent, output_dir, by_test=False)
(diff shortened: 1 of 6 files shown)


dashboard.py
```diff
@@ -648,8 +648,12 @@ def main() -> None:
...
         make_timeseries_html_report(df, output_dir, runtime)
 
-    # Do not use data that is more than a week old in statistical analysis
-    df_recent = df[df["end"] > df["end"].max() - pandas.Timedelta("7d")]
+    # Do not use data that is more than a week old in statistical analysis.
+    # Also exclude failed tests.
+    df_recent = df[
+        (df["end"] > df["end"].max() - pandas.Timedelta("7d"))
+        & (df["call_outcome"] == "passed")
+    ]
```
crusaderky, line 656: The purpose of these A/B tests is to compare performance, not failure rates. For those there's already the red light on the CI build.

gjoseph92: Hm, but sometimes changes in performance can cause failures. In those cases, this could give a distorted view of a change as being purely good for performance, when it actually might just increase the _variability_ of performance to the point where some of the bad runs are failures.

At the same time, I agree that we usually don't want results from failed tests mixed in. I guess this makes more sense to merge right now, but it worries me a bit to not have an easy option to see both with and without failures.

gjoseph92 (approved):

ian-r-rose (approved):

PR #370, opened 2022-09-24
Use single job for all test categories

### LINKED ISSUE #279: Simplify CI test workflow

crusaderky:
Use pytest-xdist to parallelize [stability, runtime, benchmarks] instead of github actions.

Closes #279

### DIFF
--- a/ci/scripts/run_tests.sh
+++ b/ci/scripts/run_tests.sh
@@ -16,5 +16,4 @@ then
   EXTRA_OPTIONS="$EXTRA_OPTIONS --benchmark"
 fi
 
-python -m pytest -n 10 --dist loadscope $EXTRA_OPTIONS "$@"
-
+python -m pytest -n 8 --dist loadscope $EXTRA_OPTIONS $@
--- a/.github/workflows/ab_tests.yml
+++ b/.github/workflows/ab_tests.yml
@@ -45,7 +45,7 @@ jobs:
   # and AB_environments/config.yaml set repeat > 0
 
   tests:
-    name: A/B Tests - ${{ matrix.category }} ${{ matrix.runtime-version }} ${{ matrix.os }} py${{ matrix.python-version }}
+    name: A/B Tests - ${{ matrix.runtime-version }} ${{ matrix.os }} py${{ matrix.python-version }}
     needs: discover_ab_envs
     runs-on: ${{ matrix.os }}
     timeout-minutes: 120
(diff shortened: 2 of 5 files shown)


crusaderky:
A/B evidence: https://github.com/coiled/coiled-runtime/actions/runs/3119616495

ci/scripts/run_tests.sh
```diff
@@ -16,5 +16,4 @@ then
   EXTRA_OPTIONS="$EXTRA_OPTIONS --benchmark"
 fi
 
-python -m pytest -n 10 --dist loadscope $EXTRA_OPTIONS "$@"
-
+python -m pytest -n 8 --dist loadscope $EXTRA_OPTIONS $@
```
ian-r-rose, line 19: Any particular theory behind this change? Were we hitting rate limits in AWS?

ian-r-rose (approved):
Thanks @crusaderky, I like this change. Looks like CI times are mostly unchanged, since we are much closer to saturating xdist workers here, and most of the work is done in coiled. Pining @jrbourbeau for thoughts

ci/scripts/run_tests.sh
crusaderky: The previous limit was never reached.
benchmarks -> 6 modules
runtime -> 4 modules
stability -> 4 modules

I didn't want to suddenly stress AWS even more than what we're doing now.

ian-r-rose: Right, though it will be reached with this change. I have no objection to this, just wondering if you ran into troubles with 10 here.

PR #372, opened 2022-09-25
Set backend option to send prometheus metrics

ntabris:
I'd like to collect metrics from at least one CI run.

Maybe we want to do this for every CI run, but we should at least check the cost first.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -414,7 +414,7 @@ def small_cluster(request, dask_env_variables):
         n_workers=10,
         worker_vm_types=["t3.large"],  # 2CPU, 8GiB
         scheduler_vm_types=["t3.xlarge"],
-        backend_options={"send_prometheus_metrics": True, **backend_options},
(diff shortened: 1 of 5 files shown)


jrbourbeau (commented):
Thanks @ntabris! 

Can we set this globally instead? IIRC there's a `coiled.backend-options` (or maybe some other name?) config option we can set. That would make sure not only these, but future clusters always send prometheus metrics.

jrbourbeau:
We reference, we do that today with setting the Coiled account here 

https://github.com/coiled/coiled-runtime/blob/0768c940554d3a62fc89d7c938730bccc48a6c5f/conftest.py#L77

jrbourbeau (approved):
Talking with @ntabris about this PR in person. Collecting some of this information right now would be useful so we're going to merge this in. If we continue to find this to be useful I'd like us to globally set the config option (xref https://github.com/coiled/coiled-runtime/pull/372#pullrequestreview-1122557280)

PR #375, opened 2022-09-27
Track clusters

jrbourbeau:
cc @hayesgb @dchudz


jrbourbeau:
@dchudz all cluster associated with this PR should have their name start with `funtime-`

ntabris:
This appears to be a one-off thing, but FYI clusters are now tagged with the CI run id (which Coiled and AWS both store, so we can in theory query on this for finding clusters and also is useful for tracking cost per CI run).

PR #382, opened 2022-09-28
Add simple test for merges 

ncclementi:
Currently skipping dataframes of sizes 1x and 10x of memory cluster.

### DIFF
--- a/tests/benchmarks/test_join.py
+++ b/tests/benchmarks/test_join.py
@@ -1,0 +1,36 @@
+import dask.dataframe as dd
+import pytest
+
+from ..utils_test import cluster_memory, timeseries_of_size
+
+
+@pytest.mark.parametrize("how", ["outer", "inner"])
+def test_join_big(small_client, how):
+    memory = cluster_memory(small_client)  # 76.66 GiB
+
+    df1_big = timeseries_of_size(
+        memory * 0.9,
+    )
+
+    df2_big = timeseries_of_size(
+        memory * 0.9,
+    )
+
+    dd_result = dd.merge(df1_big, df2_big, how=how).compute()
+    assert dd_result
+
+
+@pytest.mark.parametrize("how", ["outer", "inner"])
+def test_join_big_small(small_client, how):
+    memory = cluster_memory(small_client)  # 76.66 GiB
+
+    df_big = timeseries_of_size(
+        memory * 0.9,
+    )
+
+    df_small = timeseries_of_size(
+        memory * 0.1,
+    )
+
+    dd_result = dd.merge(df_big, df_small, how=how).compute()
+    assert dd_result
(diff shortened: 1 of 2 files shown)


tests/benchmarks/test_join.py
```diff
@@ -0,0 +1,58 @@
...
+    df_big["x2"] = df_big["x"] * 1e9
+    df_big = df_big.astype({"x2": "int"})
+
+    df_small = timeseries_of_size(
+        memory * 0.01,
+    )
+
+    df_small["x2"] = df_small["x"] * 1e9
+    df_small = df_small.astype({"x2": "int"})
+    df_small = df_small.repartition(npartitions=1)
```
ncclementi, line 56: @jrbourbeau Adding this seems to add a lot on the time of the call. 
With repartition, I'm seeing a time of ~345 s see https://github.com/coiled/coiled-runtime/actions/runs/3154172388/jobs/5131528746#step:5:116 while if I don't have repartition I see ~67s https://github.com/coiled/coiled-runtime/actions/runs/3153958846/jobs/5131025373#step:5:109

Not sure what we want to do here.

ncclementi: I think the best chance here will be to make df_small a pandas dataframe to see the behavior we expect instead of repartition to 1 partition. Currently, the small dataframe is 1% fo the cluster which would be around 700 MB Maybe this is too big for 1 partition. Need to explore more how this behaves.

tests/benchmarks/test_join.py
```diff
@@ -0,0 +1,57 @@
...
+
+@pytest.mark.parametrize("mem_mult", mem_mult)  # [0.1, 1, 10]
+def test_join_big(small_client, mem_mult):
+    memory = cluster_memory(small_client)  # 76.66 GiB
+
+    df1_big = timeseries_of_size(
+        memory * mem_mult,
+    )
+    df1_big["x2"] = df1_big["x"] * 1e9
+    df1_big = df1_big.astype({"x2": "int"})
```
jrbourbeau, line 27: This is a subtle enough point that we should add a small comment about what this is doing / why it's needed

jrbourbeau (commented):
Thanks @ncclementi. Overall this looks good --  I left one comment about adding a comment. You know this, but let's review the `.github/workflows/tests.yml` changes before merging

tests/benchmarks/test_join.py
ncclementi: Done : )  see [076ddff](https://github.com/coiled/coiled-runtime/pull/382/commits/076ddff3bbad37e34510d7f1de538ab1cdcee9a2)

jrbourbeau (approved):
Thanks @ncclementi

PR #388, opened 2022-09-29
Add futures benchmarks

mrocklin:


tests/benchmarks/test_futures.py
```diff
@@ -0,0 +1,41 @@
+import numpy as np
+from dask.distributed import as_completed, wait
+from distributed.utils_test import inc, slowdec, slowinc
+
+
+def test_single_future(small_client):
+    """How quickly can we run a simple computation?"""
+    small_client.submit(inc, 1).result()
```
ian-r-rose, line 8: This will capture benchmarking information, but I'm not sure I would trust any of it, as I'd expect it to be 100% background noise.

tests/benchmarks/test_futures.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    xs = small_client.map(np.random.random, [1_000_000] * 100, pure=False)
+    ys = small_client.map(slowinc, xs, delay=0.1)
+    zs = small_client.map(slowdec, ys, delay=0.1)
+
+    futures = as_completed(zs)
+    del xs, ys, zs  # Don't keep references to intermediate results
+
+    for future in futures:  # pass through all futures, forget them immediately
+        continue
```
ian-r-rose, line 41: I'm not sure I understand the purpose of this. Won't `futures` still hang on to references?

tests/benchmarks/test_futures.py
```diff
@@ -0,0 +1,41 @@
...
+
+def test_large_map_first_work(small_client):
+    """
+    Large maps are fine, but it's pleasant to see work start immediately.
+    We have a batch_size keyword that should work here but it's not on by default.
+    Maybe it should be.
+    """
+    futures = small_client.map(inc, range(100_000))
+    for _ in as_completed(futures):
+        return
```
ian-r-rose, line 25: return?

tests/benchmarks/test_futures.py
mrocklin: 🤔 what are the sources of background noise these days?  Is it test setup, or is it Dask/Coiled runtime noise?

tests/benchmarks/test_futures.py
mrocklin: It's `as_completed`, so no.

tests/benchmarks/test_futures.py
mrocklin: Yup.  I just want to know when the first thing has completed.

tests/benchmarks/test_futures.py
mrocklin: I'm tempted now to make a test that is just ...


```python
def test_foo(small_client):
    pass
```

jrbourbeau: We've got a similar test already 

https://github.com/coiled/coiled-runtime/blob/c11ea3a6a41ae4d88eef921c14e8de5ec9f04040/tests/runtime/test_cluster_creation.py#L6-L15

mrocklin: That test makes a new cluster.  I'm curious about noise when we're using the existing cluster from the fixture (I would assume that fixture creation time doesn't get counted in metrics).  

I was mostly joking though.  Mostly.  It sounds like noise is a concern.  We should maybe be intentionally sensitive to noise.

jrbourbeau (approved):
Thanks @mrocklin -- LGTM

PR #389, opened 2022-09-29
Add benchmark for reconnecting to an existing cluster

mrocklin:
cc @ntabris


jrbourbeau (approved):
Seems reasonable -- thanks @mrocklin

PR #390, opened 2022-09-29
Add DataFrame column projections tests 

jrbourbeau:
Today we get automatic column projection for `read_parquet` + `getitem` and `read_parquet` + `groupby.agg(...)`. This PR adds benchmarks for these two cases 

cc @ncclementi @mrocklin

### DIFF
--- a/tests/benchmarks/test_parquet.py
+++ b/tests/benchmarks/test_parquet.py
@@ -106,3 +106,25 @@ def test_download_throughput(parquet_client, kind):
         )
     elif kind == "dask":
         distributed.wait(dd.read_parquet(path, engine="pyarrow").persist())
+
+
+def test_parquet_column_projection(parquet_client):
+    """
+    Selecting columns after a `read_parquet` call should push
+    column selection into the original `read_parquet` call
+    """
+    ddf = dd.read_parquet(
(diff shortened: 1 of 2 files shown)


tests/benchmarks/test_parquet.py
```diff
@@ -106,3 +106,25 @@ def load(path):
...
+
+def test_parquet_groupby_agg_column_projection(parquet_client):
+    """
+    `read_parquet` + grouby-aggregation will push column selection
+    into the original `read_parquet` call
+    """
+    ddf = dd.read_parquet(
+        "s3://coiled-runtime-ci/nyc-tlc/yellow_tripdata_2019_parquet/yellow_tripdata_2019-*.parquet",
+    )
+    ddf.groupby("passenger_count").agg({"tip_amount": "mean"}).compute()
```
mrocklin, line 130: I propose that we add a simple elementwise layer in between these two.  I don't want us writing benchmarks around what works today.

jrbourbeau: We had an offline conversation around this. Take home point is there's definitely some value in defensive benchmarks (i.e. this thing works, let's make sure it continues to work), but we should also be thinking more about testing the limits of `dask`.

jrbourbeau:
I'm going to close this PR as we have H2O queries (e.g. `q1`) that are sensitive to these column projections already.

PR #391, opened 2022-09-29
Remove quickstart examples

jrbourbeau:
These are no longer used as the quickstart on cloud.coiled.io. Let's keep the csv example so we can continue to monitor `read_csv` performance. We have other `read_parquet` benchmarks that capture the parquet version of the old quickstart, so we can just remove that one.


ian-r-rose (approved):

ncclementi (approved):
This LGTM

PR #392, opened 2022-09-29
Benchmark DataFrame filtering

jrbourbeau:
This adds a simple row-filtering benchmark


ncclementi:
There were 2 regressions that worried me, terst_xgboost_distributed_training upstream and 0.1.0 but it went back to normal, or the accepted noise. 
`0.1.0`
Screen Shot 2022-10-03 at 3 41 09 PM

`upstream`
Screen Shot 2022-10-03 at 3 41 49 PM

.github/workflows/tests.yml
```diff
@@ -31,52 +31,52 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ["3.9"]
-        pytest_args: [tests]
+        pytest_args: [tests/benchmarks/test_dataframe.py]
```
ncclementi, line 34: We need to uncomment this and skip ci and then I'd say it's ready to go.

ncclementi (approved):

.github/workflows/tests.yml
crusaderky, line 34:
```suggestion
        pytest_args: [tests]
```

crusaderky (approved):

PR #409, opened 2022-09-30
Param array test ddiff

hayesgb:
This PR parameterizes `benchmarks/test_array.py::test_double_diff` to allow evaluating this operations behavior under both high and intermediate memory pressure.

In conversations with @ntabris, we suspect the high variance associated with this test is largely due to dask's current handling of spill to disk behavior.


hayesgb:
Replacing with PR from `coiled` repo rather than personal.

PR #415, opened 2022-10-03
Parametrize test_basic_sum

crusaderky:
It has been argued that `test_basic_sum` is anything but basic and does not represent a realistic use case, because its data generation phase terminates almost instantly, thus ruining all opportunities for pipelining network I/O and task execution.

This PR adds a couple extra more realistic use cases.
In-depth discussion on the impact of scheduler-side queueing below.


crusaderky:
Evidence of the new migration script:

image

crusaderky:
**Executive summary at the end**

image
image

**Raw output:** https://github.com/coiled/coiled-runtime/actions/runs/3183172088

@fjetter @gjoseph92 @hendrikmakait @hayesgb results are as predicted.

### [fast-thin]
This is the original test, unaltered.
`worker-saturation: 1.0` takes +80% time and -35% peak RAM compared to `inf`.
`worker-saturation: 2.0` takes +73% time and -10% peak RAM compared to `inf`.
1.5 sits neatly in between.

### [slow-thin]
This adds a 0.5s sleep after the data generation, for each block, for a total of (3925 chunks * 0.5s / 20 threads) = 98s theoretical "good" extra runtime.

`inf [slow-thin]` takes +93s compared to `inf [fast-thin]` (meaning there's 5s worth of better network pipelining that wasn't happening before) and, unsurprisingly, -29% peak RAM due to the fact that there's less data available to be copied around at any one given moment in time.

`1.0 [slow-thin]` takes +34% time and -40% peak RAM compared to `inf [slow-thin]`
`2.0 [slow-thin]` takes +25% time and -28% peak RAM compared to `inf [slow-thin]`
1.5 is barely distinguishable from 2.0.
`1.0 [slow-thin]` takes +89s compared to `1.0 [fast-thin]` (9s better network pipelining) and -32% peak RAM.
`2.0 [slow-thin]` takes +76s compared to `2.0 [fast-thin]` (22s better pipelining) and -42% peak RAM.

### [fast-square]
This is the exact same amount of data, chunk size, and algorithm as [fast-thin], just chunked in squares instead of 1 column stripes. It was extremely fast (<15s), as predicted. I had to remove it from the test as the time measures were very volatile (10~20s).

### [slow-square]
This is the exact same amount of data, chunk size, and algorithm as [slow-thin], just chunked in squares instead of 1 column stripes. 
We can observe marginal improvements for `worker-saturation: 1.5` over the other measures, both in terms of CPU and RAM.

# Conclusions / Executive Summary

The team was concerned that enabling queueing would unacceptably deteriorate performance in one use case in our test suite, `test_basic_sum`, making it 80% slower. The test supposedly illustrates a trivial map-reduce algorithm with heavy network I/O caused by inefficient chunking.

However, I am now persuaded that the original ``test_basic_sum`` is anything but basic and is unlikely to be representative of a real workload, due to the fact that its embarassingly parallel phase (e.g. data loading), a.k.a. "map" phase, takes a fraction of second to complete and immediately moves onto the reduce phase.

As soon as the map takes as little as 0.5s extra per 100MiB chunk, then the performance degradation introduced by queueing drops to +34% time compared to queueing disabled, while retaining its major RAM savings.

Given the above, I recommend enabling queueing straight away, without waiting for Speculative Task Assignment (STA).

#### Best value for `worker-saturation`
The test showed modest differences between 1.0 (do not send any more rootish tasks to the worker until it's starving), 1.5 (on 2 threads per worker, have 1 extra rootish task already on the worker ready to be executed at all times), and 2.0 (on 2 threads per worker, have 2 extra rootish tasks ready at all times).

1.0 personally makes me feel very nervous, as it will case round-trip-time (RTT) between scheduler and workers to have a direct impact on end-to-end runtime, and some dask users are very likely going to have much higher RTT than coiled.
Therefore I suggest setting it to 1.5.

crusaderky:
Reverted temporary A/B tests. This is ready for review and merge.

ci/scripts/combine-dbs.sh
```diff
@@ -13,6 +13,8 @@ else
 fi
 DB_NAME=benchmark.tmp.db alembic upgrade head
 
+sqlite3 benchmark.tmp.db < ci/scripts/migrate.sql
```
fjetter, line 16: We're using alembic to manage DB migrations. See https://github.com/coiled/coiled-runtime/blob/main/README.md#changing-the-benchmark-schema for instructions (you can reuse your script, just paste the query in the python skeleton that's auto generated.)
See https://github.com/coiled/coiled-runtime/pull/338/files#diff-5d6d77339ddca1137203f45ebe646d7b9227c52acd07a22bf8af33786ab20692 for an example

fjetter (changes requested):
I'm OK with the test modifications but the DB migration should be done using alembic

ci/scripts/combine-dbs.sh
crusaderky: Changed to use alembic. Testing now

PR #424, opened 2022-10-04
No CI failed report when workflows are canceled

### LINKED ISSUE #406: Don't generate CI failure issue when jobs are cancelled
In https://github.com/coiled/coiled-runtime/issues/403 we had a new "CI Failed" issue opened for GHA jobs that were cancelled (due to multiple commits being pushed to `main`). We shouldn't open a new issue if it's due to jobs being cancelled

ncclementi:
- [x] Closes #406

I'm looking at the [CI run](https://github.com/coiled/coiled-runtime/issues/403) mentioned in the issue and I noticed the reason why this open a CI failed report is not because of the cancelation but because of the regression. Notice that the logic says: 

```yaml
  report:
    name: report
    needs: [tests, regressions]
    if: |
      always()
      && github.event_name != 'pull_request'
      && github.repository == 'coiled/coiled-runtime'
      && (needs.tests.result == 'failure' || needs.regressions.result == 'failure') 
```

and according to the GHA docs `jobs.<job_id>.result` is The result of a job in the reusable workflow. Possible values are success, failure, cancelled, or skipped. So what trigger the CI failure issue opened was `needs.regressions.result == 'failure'`

I think the way to go here is to add a condition to the report section that checks `needs.tests.result != 'cancelled'` (proposed on this PR)

This will avoid the report but we will still go through the process-results step which updates the database. I'm not fully convinced we want this, or if we want to avoid the step of process-results altogether when cancelations happen, any thoughts? 

cc: @ian-r-rose

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -122,7 +122,7 @@ jobs:
   process-results:
     needs: tests
     name: Combine separate benchmark results
-    if: always() && github.repository == 'coiled/coiled-runtime'
+    if: always() && github.repository == 'coiled/coiled-runtime' && needs.tests.result != 'cancelled'
     runs-on: ubuntu-latest
     concurrency:
       # Fairly strict concurrency rule to avoid stepping on benchmark db.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -221,6 +221,7 @@ jobs:
       always()
       && github.event_name != 'pull_request'
       && github.repository == 'coiled/coiled-runtime'
+      && needs.tests.result != 'cancelled'
       && (needs.tests.result == 'failure' || needs.regressions.result == 'failure') 
 
     runs-on: ubuntu-latest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -122,7 +122,7 @@ jobs:
   process-results:
     needs: tests
     name: Combine separate benchmark results
-    if: always() && github.repository == 'coiled/coiled-runtime' && needs.tests.result != 'cancelled'
+    if: always() && github.repository == 'coiled/coiled-runtime'
     runs-on: ubuntu-latest
     concurrency:
       # Fairly strict concurrency rule to avoid stepping on benchmark db.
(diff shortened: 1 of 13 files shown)


ian-r-rose:
Thanks @ncclementi, this looks like a reasonable fix to me. I wonder if we should even bother running the regressions script when the test run is canceled. My read of the conditions is that skipping that step would also fix the issue.

ncclementi:
> Thanks @ncclementi, this looks like a reasonable fix to me. I wonder if we should even bother running the regressions script when the test run is canceled. My read of the conditions is that skipping that step would also fix the issue.

Yes, I'm torn here too because it could happen that some tests are canceled and some are not. In this case, do we want the proper runs to be checked for regressions? 
The way it is now, regressions will show ionn the CI, but it won't open an issue if tests were canceled.

ian-r-rose:
If I'm canceling tests, I probably don't care about regressions. That being said, I also don't think it's *wrong* to run the script.

ian-r-rose (approved):

.github/workflows/tests.yml
```diff
@@ -122,7 +122,7 @@ jobs:
   process-results:
     needs: tests
     name: Combine separate benchmark results
-    if: always() && github.repository == 'coiled/coiled-runtime'
+    if: always() && github.repository == 'coiled/coiled-runtime' && needs.tests.result != 'cancelled'
```
ncclementi, line 125: @ian-r-rose I agree that if tests were canceled we don't care about regressions. I added this, I assume this will mean that regression will be skipped, and the change on line 224 can be removed right?
 

In that case I could remove

.github/workflows/tests.yml
```diff
@@ -221,6 +221,7 @@ jobs:
       always()
       && github.event_name != 'pull_request'
       && github.repository == 'coiled/coiled-runtime'
+      && needs.tests.result != 'cancelled'
```
ncclementi, line 224:
```suggestion

```

.github/workflows/tests.yml
ian-r-rose: I think that's not right, since the regressions condition has an `always()`. I meant that we can add this to the regressions step, rather than the process-results step

ian-r-rose (approved):

Issue #401, opened 2022-09-29
Record cluster name/id/details link for each test


ntabris:
We often want to look up more details about a specific test run. This is hard to find. Since pytest doesn't let us easily output during successful tests, it would be nice to have this as report that's generated/saved at end of test run.

cc @ian-r-rose

PR #430, opened 2022-10-06
Always dump cluster state

crusaderky:
Dump cluster state for all successful runs and dump URI to the database.

### DIFF
--- a/benchmark_schema.py
+++ b/benchmark_schema.py
@@ -26,8 +26,6 @@ class TestRun(Base):
     distributed_version = Column(String, nullable=True)
     python_version = Column(String, nullable=True)
     platform = Column(String, nullable=True)
-    cluster_name = Column(String, nullable=True)
-    cluster_id = Column(Integer, nullable=True)
 
     # CI runner data
     ci_run_url = Column(String, nullable=True)
--- a/benchmark_schema.py
+++ b/benchmark_schema.py
@@ -26,6 +26,8 @@ class TestRun(Base):
     distributed_version = Column(String, nullable=True)
     python_version = Column(String, nullable=True)
     platform = Column(String, nullable=True)
+    cluster_name = Column(String, nullable=True)
+    cluster_id = Column(Integer, nullable=True)
 
     # CI runner data
     ci_run_url = Column(String, nullable=True)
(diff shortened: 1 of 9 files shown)


benchmark_schema.py
```diff
@@ -26,6 +26,8 @@ class TestRun(Base):
     distributed_version = Column(String, nullable=True)
     python_version = Column(String, nullable=True)
     platform = Column(String, nullable=True)
+    cluster_name = Column(String, nullable=True)
+    cluster_id = Column(Integer, nullable=True)
```
ian-r-rose, line 30: Not that @ncclementi is doing the same thing in #401. Also, this will require an alembic migration

ncclementi: PR for reference, https://github.com/coiled/coiled-runtime/pull/433

crusaderky: I've reverted my version

crusaderky:
image

Tested I can download the file by simply typing `aws s3 cp <paste from db> ./`

crusaderky:
This is ready for review and merge

benchmark_schema.py
```diff
@@ -39,12 +39,12 @@ class TestRun(Base):
     average_memory = Column(Float, nullable=True)
     peak_memory = Column(Float, nullable=True)
 
-    # Timing data (Note: this is not yet collected)
+    # Timing data (not yet collected)
```
ian-r-rose, line 42: Hah, this comment is out of date, we are collecting this info now. But I can update separately.

conftest.py
```diff
@@ -434,7 +432,7 @@ def small_client(
         client.wait_for_workers(10)
         client.restart()
 
-        with upload_cluster_dump(client, small_cluster), benchmark_all(client):
```
ian-r-rose, line 437: Thanks, I've been meaning to make this change.

dashboard.py
```diff
@@ -242,9 +242,13 @@ def make_ab_confidence_map(
             "runtime",
         ]
     ]
+
     runtimes = df["runtime"].unique()
+    if len(runtimes) < 2 or baseline not in runtimes:
```
ian-r-rose, line 247: I've made a similar fix in #432. But will this actually solve the problem? I think the problem that I saw was when `runtimes` was entirely full of `baseline`, not if `baseline` didn't appear.

crusaderky: I now moved this fix to #434 and it does solve the problem

crusaderky:
@ian-r-rose is there anything outstanding here?

ian-r-rose (approved):
LGTM, thanks @crusaderky

PR #434, opened 2022-10-07
Fix dashboard bugs

crusaderky:
- Fix bug in barcharts where upstream had an incorrect tooltip, showing only 1 run
- Fix bug in A/B charts which caused a crash in case of partially missing data
- Add optional dump of filtered pandas data, for debugging purposes

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -171,8 +171,10 @@ def make_barchart(
     tooltip = [
         altair.Tooltip("fullname:N", title="Test"),
         altair.Tooltip("runtime:N", title="Runtime"),
-        altair.Tooltip("dask_version:N", title="Dask"),
-        altair.Tooltip("distributed_version:N", title="Distributed"),
+        altair.Tooltip("min(dask_version):N", title="Dask (min)"),
+        altair.Tooltip("max(dask_version):N", title="Dask (max)"),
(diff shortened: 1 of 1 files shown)


ian-r-rose (approved):

ian-r-rose:
CI looks like it's stuck in a hanging state, even though tests have passed. Merging, thanks @crusaderky

dashboard.py
```diff
@@ -242,12 +244,13 @@ def make_ab_confidence_map(
             "runtime",
         ]
     ]
+
     runtimes = df["runtime"].unique()
-    A = baseline
-    B = next((r for r in runtimes if r != baseline), None)
-    if not B:
```
ian-r-rose, line 248: I still think this fix was necessary -- there are lots of ways to get subsets of data that only include the baseline numbers.

crusaderky: On this line, you'll always get at most baseline + exactly one other runtime.
My change adds a check for the existence of baseline, but does not remove the check for the other runtime.

ian-r-rose: If you have a subset of data with *only* baseline runtimes, your check still causes a `StopIteration` error.

crusaderky: If you have only baseline, then the check 2 lines above `if len(runtimes) < 2` will fail.

PR #436, opened 2022-10-07
Stabilize test_future.py

crusaderky:
Make runtime of `test_single_future` and `test_memory_efficient` more stable

### DIFF
--- a/tests/benchmarks/test_futures.py
+++ b/tests/benchmarks/test_futures.py
@@ -5,7 +5,8 @@ from distributed.utils_test import inc, slowdec, slowinc
 
 def test_single_future(small_client):
     """How quickly can we run a simple computation?"""
-    small_client.submit(inc, 1).result()
+    for i in range(400):
+        small_client.submit(inc, i).result()
 
 
 def test_large_map(small_client):
@@ -28,14 +29,14 @@ def test_large_map_first_work(small_client):
 def test_memory_efficient(small_client):
     """
     We hope that we pipeline xs->ys->zs without keeping all of the xs in memory
-    to start.  This may not actually happen today.
+    to start. This may not actually happen today.
     """
-    xs = small_client.map(np.random.random, [1_000_000] * 100, pure=False)
-    ys = small_client.map(slowinc, xs, delay=0.1)
-    zs = small_client.map(slowdec, ys, delay=0.1)
+    xs = small_client.map(np.random.random, [20_000_000] * 100, pure=False)
+    ys = small_client.map(slowinc, xs, delay=1)
+    zs = small_client.map(slowdec, ys, delay=1)
 
     futures = as_completed(zs)
     del xs, ys, zs  # Don't keep references to intermediate results
 
-    for future in futures:  # pass through all futures, forget them immediately
+    for _ in futures:  # pass through all futures, forget them immediately
         continue
--- a/tests/benchmarks/test_futures.py
+++ b/tests/benchmarks/test_futures.py
@@ -4,8 +4,11 @@ from distributed.utils_test import inc, slowdec, slowinc
 
 
 def test_single_future(small_client):
-    """How quickly can we run a simple computation?"""
-    for i in range(400):
+    """How quickly can we run a simple computation?
+    Repeat the test a few times to get a more sensible
+    cumulative measure.
+    """
+    for i in range(100):
         small_client.submit(inc, i).result()
 
 
--- a/alembic/versions/82e8a0855e91_declare_bankruptcy_for_test_futures.py
+++ b/alembic/versions/82e8a0855e91_declare_bankruptcy_for_test_futures.py
@@ -1,0 +1,28 @@
+"""Declare bankruptcy for test_futures
+
+Revision ID: 82e8a0855e91
+Revises: a8785a7b3cae
+Create Date: 2022-10-18 15:21:42.845790
+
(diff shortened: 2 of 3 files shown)


crusaderky:
## test_single_future
- increased runtime from 0.5~2.5s to 41s
- increased memory from 0.6~1.1 GiB to 1.4 GiB
## test_memory_efficient
- increased runtime from 1.6~2.4s to 15s
- increased average memory from 1.0~1.3 GiB to 9 GiB
- increased peak memory from 1.7~2.0 GiB to 17 GiB

Ready for review and merge.

fjetter:
Do we want to delete existing data from the DB or are we OK having this jump there?

crusaderky:
I declared bankruptcy. Testing it now

tests/benchmarks/test_futures.py
```diff
@@ -5,7 +5,8 @@
 
 def test_single_future(small_client):
     """How quickly can we run a simple computation?"""
-    small_client.submit(inc, 1).result()
+    for i in range(400):
```
jrbourbeau, line 8: A comment as to why we're looping here would be useful for future readers. We might also want to update the test name as it doesn't seem fully right anymore with these changes.

jrbourbeau: Also, you mentioned 400 iterations causes this test to take ~41s. Can we have fewer iterations  while still making the wall time long enough to meaningfully measure differences? For example, is 100 iterations enough?

jrbourbeau (commented):
Thanks @crusaderky. A couple of small comments, otherwise this looks good

crusaderky:
@jrbourbeau code review comments addressed

jrbourbeau (commented):
Thanks @crusaderky! This failure, related to the migration, is showing up in CI

```python
____________________ ERROR at setup of test_shuffle_simple _____________________
[gw0] linux -- Python 3.8.13 /usr/share/miniconda3/envs/test/bin/python

pytestconfig = <_pytest.config.Config object at 0x7f01c5d40df0>
tmp_path_factory = TempPathFactory(_given_basetemp=PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw0'), _trace=<pluggy._tracing.TagTracerSub object at 0x7f01b144abe0>, _basetemp=PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw0'))

    def benchmark_db_engine(pytestconfig, tmp_path_factory):
        """Session-scoped fixture for the SQLAlchemy engine for the benchmark sqlite database.
    
        You can control the database name with the environment variable ``DB_NAME``,
        which defaults to ``benchmark.db``. Most tests shouldn't need to include this
        fixture directly.
    
        Yields
        ------
        The SQLAlchemy engine if the ``--benchmark`` option is set, None otherwise.
        """
    
        if not pytestconfig.getoption("--benchmark"):
            yield
        else:
            engine = sqlalchemy.create_engine(f"sqlite:///{DB_NAME}", future=True)
    
            # get the temp directory shared by all workers
            root_tmp_dir = tmp_path_factory.getbasetemp().parent
            lock = root_tmp_dir / (DB_NAME + ".lock")
    
            # Create the db if it does not exist already.
            with filelock.FileLock(lock):
                if not os.path.exists(DB_NAME):
                    with engine.connect() as conn:
                        conn.execute(sqlalchemy.text("VACUUM"))
    
            # Run migrations if we are not up-to-date.
            with filelock.FileLock(lock):
                current = subprocess.check_output(["alembic", "current"], text=True)
                if "(head)" not in current:
                    p = subprocess.run(["alembic", "upgrade", "head"])
>                   p.check_returncode()

conftest.py:135: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = CompletedProcess(args=['alembic', 'upgrade', 'head'], returncode=255)

    def check_returncode(self):
        """Raise CalledProcessError if the exit code is non-zero."""
        if self.returncode:
>           raise CalledProcessError(self.returncode, self.args, self.stdout,
                                     self.stderr)
E           subprocess.CalledProcessError: Command '['alembic', 'upgrade', 'head']' returned non-zero exit status 255.

/usr/share/miniconda3/envs/test/lib/python3.8/subprocess.py:448: CalledProcessError
---------------------------- Captured stdout setup -----------------------------
FAILED: Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
---------------------------- Captured stderr setup -----------------------------
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
ERROR [alembic.util.messaging] Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
```

Otherwise this looks good to go

tests/benchmarks/test_futures.py
jrbourbeau: Good question -- maybe from some of the plots in https://github.com/coiled/coiled-runtime/issues/420?

Issue #252, opened 2022-08-17
How to handle tests that "should work" but don't?


gjoseph92:
In https://github.com/coiled/coiled-runtime/pull/243#issuecomment-1217395868, I'm adding a workload that should work—and in fact does work on a fresh cluster—but always fails in CI right now due to some combination of
* https://github.com/dask/distributed/issues/4987#issuecomment-877473670
* https://github.com/dask/distributed/issues/5223

In general, how to we want to handle these sorts of tests?

#### 1. Don't run them (skip / don't add at all)

This is simple since then we don't have to change anything right now.

If they're skipped, when we think the issue might be fixed, we could hopefully remember to go un-skip them, see if that passes, and if so, merge the un-skip.

However, that manual process (remember to un-skip this) won't scale well to many tests, is easy to forget, and also doesn't exactly confirm that the thing that you think fixed the problem actually did, since (without even more manual work) you don't have a record of the test failing before the suspected fix was merged.

#### 2. Allow some tests to fail (xfail?), and have a way of displaying that on the dashboard

This would be a little more work, since we'd need to figure out a way to represent failure on the dashboard. But that would also tell a nice story when a lot of Xs turned into a normal line.

It's also nice to know if something unintentionally fixes the test (though this is probably pretty rare). Plus, I think it's generally good to be upfront about things that don't work.

A downside here is increased runtime (and cost) for the tests, running something all the time that we know is going to fail (and the failure mode is probably that it times out after 10min of runtime or something).

Another thing @fjetter mentioned is that we don't want to have a bunch of xfailed tests hanging around forever. **If a test is going to be xfailed, it must have an issue opened for it, and the expectation that we'd actually work on that issue in the near term.** Otherwise, if something doesn't work and we don't expect it will for a long time, it should probably be skipped or not added at all.

cc @hayesgb @jrbourbeau @ian-r-rose @ncclementi

ian-r-rose:
I said a version of this over in #243, but re-stating here:

I very much like the *idea* of having failing tests also show up in whatever charts we are showing, so that we can note when they are fixed. We are already tracking whether a test succeeded or failed, so we could do something like change the shape or color of a marker for failures.

[...]

PR #443, opened 2022-10-12
OLS Workload

hayesgb:
This adds a stability test based on an OLS workflow that is reported to be stressful on the cluster.


tests/stability/test_ols.py
```diff
@@ -0,0 +1,12 @@
+import dask.array as da
+
+from ..utils_test import wait
+
+
+def test_ols(small_client, num_samples=int(1e8), num_coeffs=100, chunksize=int(1e6)):
```
jrbourbeau, line 6: Nit: While defining `num_samples`, `num_coeffs`, and `chunksize` as keyword arguments seems to be working, this isn't typically done when writing `pytest`-style tests. Instead I'd recommend something like (note, I've not run linters on the code below):

```suggestion
def test_ols(small_client):
    num_samples = int(1e8)
    num_coeffs = 100
    chunksize = int(1e6)
```

which is more `pytest`-idiomatic. This should reduce the amount of thought future readers of this test will need.

jrbourbeau (commented):
Thanks @hayesgb! 

Instead of creating a new `test_old.py` module for this test, let's move it to `tests/stability/test_array.py`, which seems more appropriate at the moment. If we end with a bunch of OLS-related tests, then we can pull things into a separate module. Since we're currently reusing clusters within modules, moving to `test_array.py` will also avoid the spinup time for an additional cluster. 

This test is taking ~14 minutes to run. Do we have a sense for what's problematic about this particular example? I wonder if we can reduce the runtime, while still preserving the usefulness as a stability test

hayesgb:
cc: @gjoseph92

hayesgb:
@jrbourbeau -- Thanks for the review.  I've tested this with `worker-saturation` on and the execution time drops to 10-11 minutes.  What are your thoughts on marking this as skipped until that's on be default?

gjoseph92:
xref #252. Ideally I think we’d just add this and then see it start working
better when the default gets changed?

hayesgb:
You're right @gjoseph92 .

After digging a little more, I've also learned this was run on a larger cluster, so I should be able to modify with `scaled_array_shape` to get the array / cluster size closer aligned to the original example.

hayesgb:
This size array (relative to cluster size) is more closely represents the workload of interest.  At this scale, the. workload runs in about 100 seconds.  With `worker-saturation=1.0`, that drops to about 40.

hayesgb:
Any additional concerns here @jrbourbeau ?

ncclementi:
@hayesgb Would you mind reverting this and fixing all the CI commented text, and fixing the test suit to run?

Issue #444, opened 2022-10-12
compare costs w/ non-burstable instance types


ntabris:
Most of the tests currently use `t3`. For these, you pay extra if you have longer durations of high-CPU load.

This is what we're paying extra, in addition to the base `t3` instance cost:

<img width="961" alt="image" src="https://user-images.githubusercontent.com/46289310/195408081-458caf80-50fd-4cc8-8224-1c604a202c16.png">

And here are the base `t3` instance costs for same period:

<img width="959" alt="image" src="https://user-images.githubusercontent.com/46289310/195408263-e96c4763-1846-48ff-bc86-dfa00a088b93.png">

I suspect that switching off `t3` to something w/ higher base cost but no charge for high CPU util would save money. We (probably me) should check this.

ntabris:
This is also maybe a thing to think about in how we (platform) suggest instance types for people, it's definitely possible `t3` would be more expensive for various workloads than `m6i` (which also potentially gets better network bandwidth).

shughes-uk:
I'd take into consideration availability and free tier quota ect for whatever we pick as the default.

ntabris:
At least for coiled-runtime tests, using `m6i` instead of `t3` has reduced cost and also (somewhat to my surprise) makes a non-trivial difference in perf.

Here's chart of t3 "burst" usage/cost:

<img width="849" alt="image" src="https://user-images.githubusercontent.com/46289310/200873840-3fe9aa2b-8fd7-4436-9e16-21a480b20934.png">

We'll need to think more about what this means for instance types on the Coiled platform more generally.

PR #447, opened 2022-10-13
Use m6i (fixed cost) instead of t3 (burstable)

ntabris:
Costs look better, perf (based on single case) also looks better.

Related to #444


ntabris:
TODO: wait a day or two till Cost Explorer has ingested data, then look at costs where `GITHUB_RUN_ID` (tag) = `3251189249` to see total cost for https://github.com/coiled/coiled-runtime/actions/runs/3251189249/

ncclementi:
@ntabris this might help in the future, you can trigger CI by doing `git commit -m "trigger ci" --allow-empty` and that would do it, no need to add an empty file.

ntabris:
This does appear to slightly reduce cost per test.

Using primarily t3, EC2 cost is $5.20 (compute) + $3.37 (other) = $8.57

Using primarily m6i, EC2 cost is $6.46 (compute) + $0.94 (other) = $7.40

That's not a huge difference, but it's enough that I lean toward making this change unless there are any concerns. I'm not yet sure how/whether this change impacts perf results.

ntabris:
This is single data point but wallclock times for this PR are mostly lower. The last point is this PR, earlier points our not this PR.

image

tests/benchmarks/test_parquet.py
```diff
@@ -24,6 +24,7 @@ def parquet_cluster(dask_env_variables):
         package_sync=True,
         environ=dask_env_variables,
         backend_options={"send_prometheus_metrics": True},
+        tags=gitlab_cluster_tags,
```
jrbourbeau, line 27: Should we be changing instance types here too?

conftest.py
```diff
@@ -416,6 +416,20 @@ def _benchmark_all(client):
...
+    tag_names = [
+        "GITHUB_JOB",
+        "GITHUB_REF",
+        "GITHUB_RUN_ATTEMPT",
+        "GITHUB_RUN_ID",
+        "GITHUB_RUN_NUMBER",
+        "GITHUB_SHA",
+    ]
+
+    return {tag: os.environ.get(tag, "") for tag in tag_names}
```
jrbourbeau, line 430: I asked offline, but asking here for completeness: is there a config option we can set for specifying tags globally, instead of needing to pass these through to each cluster instance? 

The answer was no -- though it would be useful in this particular case

jrbourbeau (commented):
Thanks @ntabris, apologies for the delayed response. I'm not very familiar with AWS instances types -- do you know if there are performance differences between `t3` and `m6i` instances (apart from their burstability)? I ask because I think there are tests where the cluster configuration is important

tests/benchmarks/test_parquet.py
ntabris: Nope, I'm just changing places that are currently using t3.

conftest.py
ntabris: No. It's a feature-enhancement we could add without too much trouble, obviously if we added we could later change this code to use that.

ntabris:
> do you know if there are performance differences between `t3` and `m6i` instances (apart from their burstability)?

I've matched vCPU and memory numbers, so tests that are tuned to number of cores or amount of memory should be fine as far as that goes.

The individual cores may be faster (newer gen Intel chips). They can also get more peak network bandwidth.

So I do expect this to affect comparability over time, but I don't expect this to impact things like amount of data/memory pressure/spill or whether a given test is cpu/io/network bound, though it's possible if a test was only slightly more cpu/io/network bound it will be affected.

tests/benchmarks/test_work_stealing.py
```diff
@@ -27,16 +27,18 @@ def test_work_stealing_on_scaling_up(
     upload_cluster_dump,
     benchmark_all,
     dask_env_variables,
+    gitlab_cluster_tags,
 ):
     with Cluster(
         name=test_name_uuid,
         n_workers=1,
         worker_vm_types=["t3.medium"],
```
ncclementi, line 35: In this case, don't we want to change the worker's vm type too for its m6 equivalent?

tests/benchmarks/test_work_stealing.py
```diff
@@ -84,16 +86,21 @@ def clog(n):
...
+    test_name_uuid,
+    upload_cluster_dump,
+    benchmark_all,
+    dask_env_variables,
+    gitlab_cluster_tags,
 ):
     with Cluster(
         name=test_name_uuid,
         n_workers=10,
         worker_vm_types=["t3.medium"],
```
ncclementi, line 98: Same comment than than above, don't we want the m6 equivalent too?

ntabris: same reply, no equivalent of t3.medium

jrbourbeau (approved):
Thanks @ntabris

PR #449, opened 2022-10-14
Add Snowflake benchmarks (rough) [DNM]

phobson:
Couple of things to address with this:

1. Need the snowflake creds added as secrets to the repo
2. I just tacked `dask-snowflake` into the conda install command. Didn't add it to the requirements. I don't think it should be a requirement for the runtime, but it wouldn't be difficult to convince me that it could be
3. Based on user reports, I think it'd be nice to break to the read/write into separate benchmarks by getting more granular with the fixtures and context managers.

### DIFF
--- a/ci/scripts/install_coiled_runtime.sh
+++ b/ci/scripts/install_coiled_runtime.sh
@@ -10,7 +10,7 @@ then
   cat $1
   mamba env update --file $1
 else
-  mamba install -c conda-forge coiled-runtime=$COILED_RUNTIME_VERSION
+  mamba install -c conda-forge coiled-runtime=$COILED_RUNTIME_VERSION dask-snowflake
 fi
 
 # For debugging
--- a/tests/benchmarks/test_snowflake.py
+++ b/tests/benchmarks/test_snowflake.py
@@ -1,0 +1,58 @@
+import os
+import uuid
+
+import dask
+import dask.dataframe as dd
+import pytest
+from dask_snowflake import read_snowflake, to_snowflake
+from distributed import Client
+from snowflake.sqlalchemy import URL
+from sqlalchemy import create_engine
+
+
+@pytest.fixture
+def client():
+    with Client(n_workers=2, threads_per_worker=10) as client:
+        yield client
+
+
+@pytest.fixture
+def table(connection_kwargs):
+    name = f"test_table_{uuid.uuid4().hex}".upper()
+
+    yield name
+
+    engine = create_engine(URL(**connection_kwargs))
+    engine.execute(f"DROP TABLE IF EXISTS {name}")
+
+
+@pytest.fixture(scope="module")
+def connection_kwargs():
+    return dict(
+        user=os.environ["SNOWFLAKE_USER"],
+        password=os.environ["SNOWFLAKE_PASSWORD"],
+        account=os.environ["SNOWFLAKE_ACCOUNT"],
+        database="testdb",
+        schema="public",
+        warehouse=os.environ["SNOWFLAKE_WAREHOUSE"],
+        role=os.environ["SNOWFLAKE_ROLE"],
+    )
+
+
+ddf = dask.datasets.timeseries(
+    start="2000-01-01", end="2000-12-31", freq="1T", partition_freq="1W"
+)
+
+
+def test_write_read_roundtrip(table, connection_kwargs, small_client):
+    to_snowflake(ddf, name=table, connection_kwargs=connection_kwargs)
+
+    query = f"SELECT * FROM {table}"
+    df_out = (
+        read_snowflake(query, connection_kwargs=connection_kwargs)
+        .rename(columns=str.upper)  # snowflake's column casing is weird
+        .sort_values(by="A")  # sort needed to re-sync the partitions
+        .reset_index(drop=True)
+    )
+
+    dd.utils.assert_eq(ddf, df_out, check_dtype=False)


tests/benchmarks/test_snowflake.py
```diff
@@ -0,0 +1,58 @@
...
+import dask
+import dask.dataframe as dd
+import pytest
+from dask_snowflake import read_snowflake, to_snowflake
+from distributed import Client
+from snowflake.sqlalchemy import URL
+from sqlalchemy import create_engine
+
+
+@pytest.fixture
```
ncclementi, line 13: Not sure I follow why are we creating a local cluster and client here.

tests/benchmarks/test_snowflake.py
```diff
@@ -0,0 +1,58 @@
...
+        role=os.environ["SNOWFLAKE_ROLE"],
+    )
+
+
+ddf = dask.datasets.timeseries(
+    start="2000-01-01", end="2000-12-31", freq="1T", partition_freq="1W"
+)
+
+
+def test_write_read_roundtrip(table, connection_kwargs, small_client):
```
ncclementi, line 47: This is more of a question, do we want the test to be a round-trip, or having separate tests for write and read, could give us more information about where things are going wrong?

ci/scripts/install_coiled_runtime.sh
```diff
@@ -10,7 +10,7 @@ then
   cat $1
   mamba env update --file $1
 else
-  mamba install -c conda-forge coiled-runtime=$COILED_RUNTIME_VERSION
+  mamba install -c conda-forge coiled-runtime=$COILED_RUNTIME_VERSION dask-snowflake
```
ncclementi, line 13: Should we ping the dask-snowflake version we are testing? I wonder if this will update/downgrade some of the other packages that are installed with the runtime


I think here we are also missing the case for upstream and latest? not sure what's the best way to tackle this.

ncclementi (commented):
Regarding adding the secrets, do we have the credentials somewhere? Happy to pair for a couple of minutes and add them. I'm not sure where they are. 

Once we add the this will require modifying the `test.yaml` to make it aware of the secrets before running the test somewhere in here https://github.com/coiled/coiled-runtime/blob/595c449f9a8c55ffe23d64c639314c46a4db1540/.github/workflows/tests.yml#L103-L120

tests/benchmarks/test_snowflake.py
phobson: Oops -- we don't need that.

tests/benchmarks/test_snowflake.py
phobson: I don't want the test to be round trip and I agree that individual read/write benchmarks would be better.

My thinking is we should switch out the `small_client` fixture for `small_cluster` and then individually benchmark each step in a single test.

The reason for this is that writing a single test benchmark writing is easy enough, but then we'll need to clean up the test's side effects and remove that table. As a result, when we go to test _reading_ the table, we'll have to redundantly write the table yet again. There might be a clever way around this that I can't see. But individually benchmarking each step in a single test makes sense in my head.

ci/scripts/install_coiled_runtime.sh
phobson: yeah i'll add a pin

phobson:
> Regarding adding the secrets, do we have the credentials somewhere? Happy to pair for a couple of minutes and add them. I'm not sure where they are.
> 
> Once we add the this will require modifying the `test.yaml` to make it aware of the secrets before running the test somewhere in here
> 
> https://github.com/coiled/coiled-runtime/blob/595c449f9a8c55ffe23d64c639314c46a4db1540/.github/workflows/tests.yml#L103-L120

I believe we have the secrets in the existing dask-snowflake repo. I'm not an admin there, so I can't see them.

hayesgb:
@phobson -- Secrets are in the repo.  Is the issue that you can't view them?

phobson:
@hayesgb  that's correct. And unfortunately no one can view them. Once secrets are added, they permanently obfuscated.

hayesgb:
Thanks. I wasn't clear - I was asking if you were unable to see the names of the secrets, not the secrets themselves.  I'll claim jetlag. :-)

phobson:
Ahh sorry not picking up on that. I don't have sufficient privileges on the dask-snowflake repo to even view the names of the secrets

phobson:
Strange SQLAlchemy error that I'm investigating:
https://github.com/coiled/coiled-runtime/actions/runs/3267325445/jobs/5372347495#step:5:65

phobson:
Looks like the above error was recently fixed and that build is already available in conda-forge. I'll kick the CI and see what happens.

ncclementi:
@phobson I just fixed the conflicts (although I see failures). Regarding the tests running, I think there should be skip for the runtime versions where dask snowflake is not going to be there. 

It seems there are package conflicts with the snowflake connector in the pip installation. Not sure exactly what's happening there.

phobson:
@ncclementi thanks for the nudge. Used `pytest.importskip at the top of the file. I'm also not sure what's up with the pip failure.

phobson:
Push Yet Another commit to swap out mean/avg calculations for `median`. I like this because 1) medians are new + difficult for us and 2) snowflake is pretty good at them

phobson:
Summary so far

* If it's not overly burdensome, we'd like to add (dask-)snowflake to the runtime. Blockers would include
  * Significantly increasing the footprint of the package
  * difficulty getting the environment to solved without significant changes to the already-pinned packages
* On my M1 Mac, a fresh coiled-runtime environment adding dask-snowflake (which comes with dependencies) as a requirement increases the size of the environment by 30 MB

```shell
10:25:07 in ~ via 🅒 base
➜ mamba create --name=crt-clean python=3.10 coiled-runtime --quiet --yes

10:26:00 in ~ via 🅒 base took 33s
➜ mamba create --name=crt-sflake python=3.10 coiled-runtime dask-snowflake --quiet --yes

10:29:02 in ~ via 🅒 base took 29s
➜ du -sm $(conda info --base)/envs/crt-clean
1174	/Users/paul/mambaforge/envs/crt-clean

10:29:14 in ~ via 🅒 base
➜ du -sm $(conda info --base)/envs/crt-sflake
1204	/Users/paul/mambaforge/envs/crt-sflake
```

* The pip-build is broken. Currently investigating how to fix (locally)
* The module-level `pytest.importskip` for the snowflake benchmarks seems to be causing a failure for the v0.0.4 and v0.1.0 and I'm not sure why.

phobson:
I'm having trouble build even `main` via pip on my M1 Mac with the following error:

```
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for numcodecs
  Running setup.py clean for numcodecs
Successfully built coiled-runtime
Failed to build lz4 numcodecs
ERROR: Could not build wheels for lz4, which is required to install pyproject.toml-based projects
```

phobson:
My last commit explicitly included `snowflake-connector-python >= 2.8.1` as a coiled-runtime dependency. All of the _run_ requirements were satisfied, save a few:

* snowflake-connector-python
  * requested: >=2.8.1 (initially, [we get get it](https://github.com/coiled/coiled-runtime/actions/runs/3439358429/jobs/5736558130#step:4:52))
  * received: 2.7.1 (happens [much later](https://github.com/coiled/coiled-runtime/actions/runs/3439358429/jobs/5736558130#step:4:391))
* pyarrow
  * requested: ==9.0.0
  * recieved: 5.0.0 (because of the snowflake connector)
    * our log: https://github.com/coiled/coiled-runtime/actions/runs/3439358429/jobs/5736558130#step:4:430
    * Requirements of snowflake-connector-python 5.0.0 https://github.com/snowflakedb/snowflake-connector-python/blob/v2.7.1/setup.py#L54
* numpy
  * requested: ==1.23.3
  * got: 1.19.4 (because of [pyarrow 5.0.0](https://github.com/coiled/coiled-runtime/actions/runs/3439358429/jobs/5736558130#step:4:443))

Trying to figure out what downgrades the snowflake-connector-python package from 2.8 to 2.7

tests/benchmarks/test_snowflake.py
```diff
@@ -59,7 +59,8 @@ def test_read_from_snowflake(perma_table, connection_kwargs, small_client):
 
 
 @pytest.mark.skipif(
-    dask.__version__ < Version("2022.10.0"), reason="median not available in dask"
+    Version(dask.__version__) < Version("2022.10.0"),
+    reason="median not available in dask",
```
ncclementi, line 63:
```suggestion
    Version(dask.__version__) < Version("2022.10.0"),
    reason="No support for median in dask < 2022.10.0",
```

Nit pick: Can we switch the message to be the suggestion to be consistent with https://github.com/coiled/coiled-runtime/blob/32b6c48c75ce54bf915e1bbc353c08f98e031a6f/tests/benchmarks/h2o/test_h2o_benchmarks.py#L96-L99

phobson:
@ncclementi  did some good sleuthing and noticed that other folks are running into issues with the tight arrow pinning in `snowflake-connector-python`.

I left a comment in the issue that Naty found stating that allowing pyarrow 9.0 would benefit us as well
https://github.com/snowflakedb/snowflake-connector-python/issues/1248#issuecomment-1320089689

phobson:
Option discussed during the stand up: we could add snowflake as an optional test dependency instead of adding into the runtime itself.

ncclementi:
Noticing that pyarrow 10 is supposed to be supported now in the snowflake connector see https://github.com/snowflakedb/snowflake-connector-python/pull/1347
should we try to include this one in the tests?

fjetter:
I think this can be closed in favor of https://github.com/coiled/benchmarks/pull/827

PR #451, opened 2022-10-14
Use a single worker for cluster spinup time

jrbourbeau:
Currently we're using the default number of workers (i.e. 4) for measuring how fast a cluster spins up. This PR proposes we only spin up a single worker. This will, I think, provide the same rough time estimate we're after, while using fewer AWS resources. 

cc @ntabris


ntabris (approved):
LGTM, thanks!

ncclementi:
Upstream keeps failing but this is unrelated to the PR, I will go ahead and merge this.

PR #458, opened 2022-10-19
Add raw tabular data and grafana URL

crusaderky:
The new table data is accessible by clicking in the historical plots

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -579,8 +579,11 @@ def make_details_html_report(
         row = ds_row.to_dict()
 
         if cluster_name := row["cluster_name"]:
-            start_ts = int(row["start"].to_pydatetime().timestamp() * 1000)
-            end_ts = int(row["end"].to_pydatetime().timestamp() * 1000)
+            # Add some padding to compensate for clock skew of
+            # GitHub actions vs. Prometheus
+            ts_padding = pandas.Timedelta("5s")
+            start_ts = int((row["start"] - ts_padding).timestamp() * 1000)
+            end_ts = int((row["end"] + ts_padding).timestamp() * 1000)
             row["grafana_url"] = (
                 "http://35.86.202.18:3000/d/eU1bT-nVz/cluster-metrics-prometheus"
                 f"?from={start_ts}&to={end_ts}&var-cluster={cluster_name}"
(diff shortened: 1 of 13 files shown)


crusaderky:
@ntabris could you assist?
I've generated e.g.
http://35.86.202.18:3000/d/eU1bT-nVz/cluster-metrics-prometheus?from=1665790446518&to=1665790694449&var-cluster=test_array-c6668c2c

The timestamps are UTC. However, grafana seems to want local timestamps, as per my browser's timezone, which I can't know since I'm generating the URL statically.

ntabris:
> The timestamps are UTC. However, grafana seems to want local timestamps, as per my browser's timezone, which I can't know since I'm generating the URL statically.

I'm pretty sure your timestamps are just off by an hour.

1665790694449 (what you have as end time) is 2022-10-14T23:38:14.449Z but this cluster wasn't running till 2022-10-15T00:31:10UTC.

crusaderky:
@ntabris Ah yes thank you. My code was accidentally converting the UTC timestamp to a local one.

crusaderky:
This is ready for review and merge.

There's an outstanding problem of severe clock misalignment between github actions (where we record the start and stop timestamps of each test) and prometheus - I've observed 80s differences. As a result the slicing of the grafana plot is very crude as I was forced to add a very large padding.
There's no reason why this can be remediated later though.

dashboard.py
```diff
@@ -526,34 +540,109 @@ def make_ab_html_report(
...
+        row = ds_row.to_dict()
+
+        if cluster_name := row["cluster_name"]:
+            # Add some padding to compensate for clock differences between
+            # GitHub actions and Prometheus (we've observed 80s gaps)
+            ts_padding = pandas.Timedelta("5m")
+            start_ts = int((row["start"] - ts_padding).timestamp() * 1000)
+            end_ts = int((row["end"] + ts_padding).timestamp() * 1000)
+            row["grafana_url"] = (
+                "http://35.86.202.18:3000/d/eU1bT-nVz/cluster-metrics-prometheus"
```
ntabris, line 596: Let's use the new URL I just setup:

```suggestion
                "https://grafana.dev-sandbox.coiledhq.com/d/eU1bT-nVz/cluster-metrics-prometheus"
```

ntabris (approved):
LGTM, thanks @crusaderky!

PR #473, opened 2022-10-24
Log `small_client` fixture phases on the scheduler

hendrikmakait:
This PR logs when we start/finish setup/teardown of the `small_client` fixture on the scheduler. This helps to correlate cluster logs with issues. We have seen some CI failures in the cluster (e.g. #441) where being able to pinpoint when these phases happen would have facilitated debugging.

Merge conflicts with:
* #472


crusaderky:
The problem with `test_name_uuid` is that it's nowhere to be found in the database or in the dashboard.
I've replaced it with test name + session_id.

crusaderky (approved):

PR #484, opened 2022-11-02
Skip test_access_slices until issue 477 is resolved

ncclementi:


phobson (approved):
Looks good until we can figure out what's going on in #477

ncclementi:
Oh it looks like we just hit pip install in the middle of a deployment, I'll retrigger that one later

PR #489, opened 2022-11-03
Measure cluster connect time

crusaderky:
- Measure cluster reconnect time. The test already existed but it was not recording anything.
- Remove auto_benchmark_time fixture, as it was sensitive to ordering issues with other fixtures

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ The most relevant ones are summarized here:
 
 **`benchmark_task_durations`**: Record time spent computing, transferring data, spilling to disk, and deserializing data.
 
-**`auto_benchmark_time`**: Include this fixture to measure the wall clock time of the whole test automatically.
+**`get_cluster_info`**: Record cluster id, name, etc.
 
 **`benchmark_all`**: Record all available metrics.
 
--- a/conftest.py
+++ b/conftest.py
@@ -218,10 +218,6 @@ def benchmark_time(test_run_benchmark):
         def test_something(benchmark_time):
             with benchmark_time:
                 do_something()
-
-    See Also
-    --------
-    auto_benchmark_time
     """
 
     @contextlib.contextmanager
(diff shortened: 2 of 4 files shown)


crusaderky:
Ready for review and merge

ntabris:
@crusaderky I might be confused. There was some rpc call or something that you suspected was significantly affecting wall-clock on some test runs. I had assumed that was what you wanted to measure here, but maybe that's not the case.

crusaderky:
Client connect time to a cluster is a crude proxy for a new RPC call - although it does quite a bit more stuff on top of that.

crusaderky:
I will merge this at COB today unless there are comments

crusaderky:
@ntabris how would you visualize in the static dashboard the data from all other tests? You would need an entirely ad-hoc series of functions in `dashboard.py` in order to show it.

fjetter:
I think visualization and measurement is not necessarily the same thing. We could measure it for every test but only write it to the DB once. Something like

```python

@pytest.mark.fixture(scope="session")
def measure_connect_time(benchmark_db_engine):

    with Session(benchmark_db_engine, future=True) as session, session.begin() as benchmark_db_session:
        run = TestRun(
            session_id=testrun_uid,
            name="connect-time",
            originalname="connect-time",
            path=str(request.node.path.relative_to(TEST_DIR)),
            dask_version=dask.__version__,
            distributed_version=distributed.__version__,
            coiled_runtime_version=COILED_RUNTIME_VERSION,
            coiled_software_name=COILED_SOFTWARE_NAME,
            python_version=".".join(map(str, sys.version_info)),
            platform=sys.platform,
            ci_run_url=WORKFLOW_URL,
        )
        yield run

        rep = getattr(request.node, "rep_setup", None)
        if rep:
            run.setup_outcome = rep.outcome
        rep = getattr(request.node, "rep_call", None)
        if rep:
            run.call_outcome = rep.outcome
        rep = getattr(request.node, "rep_teardown", None)
        if rep:
            run.teardown_outcome = rep.outcome
        yield run
        benchmark_db_session.add(run)

def small_client(..., measure_connect_time):
    start = time()
    with Client(...):
        measure_connect_time.appropriate_field.update(time() - start)
        ... 
```

The code is not functioning but the idea is to have a global measurement we update. We'll write it to the DB as if it was a test case but we will not have an actual test case.

There are possibly more things we would like to collect globally in which case this kind of approach would pay off. If it's just about the one metric, I don't mind running an extra test.

ntabris:
thanks @fjetter, I think I endorse all of what you said in https://github.com/coiled/coiled-runtime/pull/489#issuecomment-1327699785

crusaderky:
Removed the offending test.
The measure of client connect time can be done in a separate PR.
Ready for review and merge again.

crusaderky:
Merging COB today if there are no objections

ncclementi (approved):
LGTM

PR #492, opened 2022-11-04
Fix `test_access_slices`

### LINKED ISSUE #477: test_access_slices - crashed worker - CI failed
[Workflow Run URL](https://github.com/coiled/coiled-runtime/actions/runs/3348927174)

hendrikmakait:
This PR persists the data accessed in `test_access_slices` on the cluster instead of moving it to the client (and dropping it right after). This reduces the memory footprint of the client since the accessed slice takes up ~2.5 GB for `N==700`. This should solve the problem with `pytest-xdist` workers dying.

**Explanation**

This solution is based on theory and trying the change on CI, not any useful information gathered from CI errors. All I can tell is that the `pytest-xdist` workers are dying _for some reason_ with
```
[gw0] linux -- Python 3.9.13 /usr/share/miniconda3/envs/test/bin/python
worker 'gw0' crashed while running 'tests/benchmarks/test_array.py::test_access_slices[700]'
```

Running `test_access_slices` locally with `.compute()` led to the client process starting to eat up a good chunk of memory and was horribly slow. Switching `.compute()` with `.persist()` made the computation return almost instantly and left memory usage flat.

Our Github-hosted CI runners should only have [7 GB of RAM for Linux VMs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), so reading a large dataset in a `pytest` runner is probably not a good idea (and benchmarks network bandwidth more than anything else). I could see this increase the r

 * Fixes #477

### DIFF
--- a/tests/benchmarks/test_array.py
+++ b/tests/benchmarks/test_array.py
@@ -248,13 +248,12 @@ def test_filter_then_average(threshold, zarr_dataset, small_client):
     zarr_dataset[zarr_dataset > threshold].mean().compute()
 
 
-@pytest.mark.skip(reason="this test crashed ci workers -see issue #477")
 @pytest.mark.parametrize("N", [700, 75, 1])
 def test_access_slices(N, zarr_dataset, small_client):
     """
     Accessing just a few chunks of a zarr array should be quick
     """
-    zarr_dataset[:N, :N, :N].compute()
+    zarr_dataset[:N, :N, :N].persist()
 
 
 def test_sum_residuals(zarr_dataset, small_client):


phobson (approved):
@hendrikmakait thanks for figuring this out.

I'd like to shed some light on why this test was added. In recent past, users have expressed frustration that selecting subset of data from a single block of a large zarr file was creating a very large task graph that involved 100% of the blocks in the zarr dataset. 

My understanding is that this is still the case and this test's aim is to provide a baseline for when selecting a small subset from a large zarr file generates an appropriately small task graph.

hendrikmakait:
Thanks for the explanation @phobson! If the problem we test for is the size of the task graph, I'm wondering if it makes more sense to rewrite this test in a way that would assert a reasonable size of the task graph it creates (and XFAIL it)?

phobson:
@hendrikmakait I'm no authority on the matter, but I think we need both. We need a benchmark that shows that that the overall process of reading a small subset of a large zarr file is fastª (by whatever means and implementation necessary), and then probably a test and that confirms that our implementation (smaller graph) is working as expected. 

ª acknowledging that it's not fast right now

gjoseph92:
With this change, we're not waiting for the computation to happen anymore. Furthermore, `persist` doesn't even block until the graph has been uploaded to the scheduler over the network—it just returns as soon as the graph has been processed.

So this does test the issue with processing large graphs client-side https://github.com/dask/dask/issues/8570, but this doesn't seem like a test that should be talking to the cluster (since it doesn't even use it) xref https://github.com/coiled/coiled-runtime/issues/420

Unless you meant to add a `wait` around the persist?

PR #502, opened 2022-11-08
Upgrades latest to dask==2022.10.2

hayesgb:
This adds upgrades dask and distributed to `2022.10.2`.

Note that pyarrow is still at `9.0.0` as v10 is not yet on `conda-forge`


ncclementi (approved):

ncclementi:
thanks @hayesgb

PR #508, opened 2022-11-09
Build docker images for ARM (+x86)

ntabris:
Unless this breaks something (or drastically increases build time?) I'd like us to build both x86 and arm docker images. This would make it much easier to use `coiled-runtime` on Coiled with ARM instances.


ntabris:
@ncclementi is there a way to trigger nightly build action for this MR? Or would it be simplest to merge? (If so, any concerns?)

jrbourbeau (approved):
Thanks @ntabris -- let's give this a try

ntabris:
Didn't work, I'll try what's described in https://github.com/docker/build-push-action/issues/302

PR #512, opened 2022-11-10
Set up QEMU and Buildx for multiarch docker build

ntabris:
Attempting to fix #508 based on https://docs.docker.com/build/ci/github-actions/examples/#multi-platform-images


ntabris:
The relevant workflow is broken so I'm going to merge this, you can yell at me later if that's a bad thing to do.

PR #516, opened 2022-11-11
Build (stable) docker image for ARM (+x86)

ntabris:
This is working on the nightlies... x86 and arm images are both available at https://hub.docker.com/r/coiled/coiled-runtime/tags?page=1&name=nightly-0.2.0a2211110032-py3

And they work on Coiled with Graviton instances:

```python
import coiled

coiled.create_software_environment(
  name='nightly-py310',
  container='coiled/coiled-runtime:nightly-0.2.0a2211110032-py3.10'
)

cluster = coiled.Cluster(
  software='nightly-py310',
  scheduler_vm_types=['t4g.large'],
  worker_vm_types=['t4g.large']
)
```

So let's build the stable released for ARM.

### DIFF
--- a/.github/workflows/software-environments-stable.yml
+++ b/.github/workflows/software-environments-stable.yml
@@ -44,6 +44,12 @@ jobs:
           echo $COILED_SOFTWARE_ENV_NAME
           echo COILED_SOFTWARE_ENV_NAME=$COILED_SOFTWARE_ENV_NAME >> $GITHUB_ENV
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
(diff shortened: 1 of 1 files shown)


jrbourbeau (approved):
Thanks @ntabris

ntabris:
Thanks, @jrbourbeau! Is it easy to manually trigger rebuild for 0.1.0 images? Or is there stuff there that's not idempotent and/or would get changes already in `main` rather than the release tag? (I think there was discussion of doing a new coiled-runtime release in the nearish future so it's okay to wait for that if it's non-trivial to release 0.1.0 for multi-arch.)

PR #514, opened 2022-11-10
Use spot + fallback + multizone for all clusters

ntabris:
Fixes #507

(I also removed backend_option to send prometheus metrics since I now have that set at account-level for `dask-engineering` account.)

PR #524, opened 2022-11-14
Pin bokeh and coiled for 0.1.1 release

ncclementi:
see https://github.com/coiled/coiled-runtime/issues/530

@jrbourbeau this should be a quick review.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set name = "coiled-runtime" %}
 {% set version = "0.2.0" + environ.get("VERSION_SUFFIX", '') %}
-{% set dask_version = environ.get("DASK_VERSION", "2022.10.2") %}
-{% set distributed_version = environ.get("DISTRIBUTED_VERSION", "2022.10.2") %}
+{% set dask_version = environ.get("DASK_VERSION", "2022.6.0") %}
+{% set distributed_version = environ.get("DISTRIBUTED_VERSION", "2022.6.0") %}
 
 package:
   name: {{ name|lower }}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -23,34 +23,33 @@ requirements:
     - python >=3.8,<3.11
     - pip
     - coiled >=0.2.41
-    - numpy ==1.23.3
-    - pandas ==1.5.1
+    - numpy ==1.21.6
+    - pandas ==1.4.2
     - dask =={{ dask_version }}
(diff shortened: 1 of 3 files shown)


recipe/meta.yaml
```diff
@@ -1,7 +1,7 @@
 {% set name = "coiled-runtime" %}
 {% set version = "0.2.0" + environ.get("VERSION_SUFFIX", '') %}
-{% set dask_version = environ.get("DASK_VERSION", "2022.10.2") %}
-{% set distributed_version = environ.get("DISTRIBUTED_VERSION", "2022.10.2") %}
+{% set dask_version = environ.get("DASK_VERSION", "2022.6.0") %}
```
ncclementi, line 3: @jrbourbeau since we are doing a patch, should we bump to 2022.6.1 see https://github.com/coiled/coiled-runtime/issues/306
cc: @hendrikmakait

recipe/meta.yaml
```diff
@@ -23,34 +23,34 @@ requirements:
     - python >=3.8,<3.11
     - pip
     - coiled >=0.2.41
```
ncclementi, line 25: I'm leaving this in as a protective measure

ncclementi:
It looks like a lot of things are broken, but not related to this PR necessarily. Is hard to tell what's happening here.

recipe/meta.yaml
jrbourbeau: Let's stick with 2022.6.0 to keep this PR as minimal as possible. I expect we'll push out another release soon where we update other package versions

jrbourbeau:
@ncclementi  I merged `main` to see if https://github.com/coiled/coiled-runtime/pull/514 will fix the CI failures we were seeing (there's a chance it does)

ncclementi:
> @ncclementi I merged `main` to see if #514 will fix the CI failures we were seeing (there's a chance it does)

You beat me to it. : )

ncclementi:
It's not clear to me what the Assertion errors we have are. @jrbourbeau It seems it comes from the scheduler, any ideas?

ntabris:
Current failure (https://github.com/coiled/coiled-runtime/actions/runs/3483410547/jobs/5826865040) doesn't look related to #514, looks like it could be a genuine issue related to `scheduler.restart()`.

The failed assertion is https://github.com/dask/distributed/blob/69b74f9f0aa46dac7facb960f1b0aaac69121e5e/distributed/scheduler.py#L5805

@fjetter has been working on things related to restart.

PR #529, opened 2022-11-15
Ensure concurrent restarts are not interfering

fjetter:
Explanation why this bypasses the race condition in the comment. See also https://github.com/dask/distributed/issues/7312

Closes https://github.com/coiled/coiled-runtime/issues/468

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
         language_version: python3
         exclude: ^alembic/versions/
   -   repo: https://github.com/pycqa/flake8
-      rev: 4.0.1
+      rev: 5.0.4
       hooks:
       - id: flake8
         language_version: python3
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ repos:
       - id: black
         language_version: python3
         exclude: ^alembic/versions/
-  -   repo: https://gitlab.com/pycqa/flake8
```
fjetter, line 18: I know this should not be in here. I got weird auth errors when downloading from gitlab. I suggest to just use the github mirror. If requested, I can break this out into another PR

fjetter:
Other occurrences of restart should not be affected since this only comes up if we call restart twice in short succession

.pre-commit-config.yaml
ncclementi: I'll give this a try, hope you don't mind pushing to your PR. If things are green I'll merge this in.

ncclementi (approved):

PR #545, opened 2022-11-20
[DNM] test spill test with NVMe instance disks

ntabris:
Let's see how instance types with NVMe affect the spill test.


ncclementi:
@ntabris are you still playing with this PR, or can we close it?

PR #549, opened 2022-11-23
Centralized kwargs to coiled.Cluster

### LINKED ISSUE #483: Configure number of workers and instance types in A/B tests

# Proposed design

```yaml

small_cluster:
  n_workers: 10

# For test_parquet.py
parquet_cluster:
  n_workers: 15
```

crusaderky:
Move the kwargs of all calls to `coiled.Cluster` to a centralized config file, which can then be overridden during A/B tests.
This reduces repetition and enables running A/B tests where what's being tested is not the software, but the infrastructure.

For example:
- 10 workers with 2 threads each vs. 5 workers with 4 threads each
- m6i vs. m6a vs. m5 instances

---
- Partially closes #483.
Out of scope: review all tests to verify that the workload scales dynamically with cluster size.
---
### Practical changes

- In all `benchmarks` and `runtime` tests using the `small_client` fixture, set `spot=true, spot_on_demand_fallback=true, multizone=true`
- In `test_parquet.py`, upgrade `scheduler_vm_types` from m5.xlarge to m6i.xlarge
- In `test_parquet.py` and `test_spill.py`, set `send_prometheus_metrics=true`
- In `test_work_stealing_on_scaling_up` and `test_repeated_merge_spill`, set `spot=true, spot_on_demand_fallback=true, multizone=true`
- In `test_work_stealing_on_straggling_worker`, set `spot=true, spot_on_demand_fallback=true, multizone=true, send_prometheus_metrics=True`


crusaderky:
CC @fjetter @jrbourbeau @ncclementi

conftest.py
```diff
@@ -451,24 +451,39 @@ def dask_env_variables():
...
     module = os.path.basename(request.fspath).split(".")[0]
     with Cluster(
         name=f"{module}-{uuid.uuid4().hex[:8]}",
-        n_workers=10,
-        worker_vm_types=["m6i.large"],  # 2CPU, 8GiB
-        scheduler_vm_types=["m6i.xlarge"],
-        backend_options=backend_options,
-        package_sync=True,
         environ=dask_env_variables,
         tags=gitlab_cluster_tags,
```
fjetter, line 485: Minor concern with this is that our benchmarks would not be able to pick up a difference here. It'd be nice if these kwargs would be hashed and/or stored in the DB somewhere

conftest.py
```diff
@@ -451,24 +451,39 @@ def dask_env_variables():
     return {k: v for k, v in os.environ.items() if k.startswith("DASK_")}
 
 
-@pytest.fixture(scope="module")
-def small_cluster(request, dask_env_variables, gitlab_cluster_tags):
-    # Extract `backend_options` for cluster from `backend_options` markers
-    backend_options = merge(
-        m.kwargs for m in request.node.iter_markers(name="backend_options")
-    )
-    backend_options["send_prometheus_metrics"] = True
```
crusaderky, line 460: This could only work at module level. I could not find it used anywhere and I wonder if there's any actual benefit in retaining the extra complexity? @jrbourbeau

tests/stability/conftest.py
```diff
@@ -1,16 +0,0 @@
...
-        spot=True, spot_on_demand_fallback=True, multizone=True
-    )
-    for item in items:
-        # Add a module-level `spot=True` backend option marker if one doesn't already exist
-        module = item.parent
-        if not any(
-            m.kwargs.get("spot", False)
-            for m in module.iter_markers(name="backend_options")
-        ):
-            module.add_marker(marker)
```
crusaderky, line 16: @jrbourbeau why did you enable spot=True, spot_on_demand_fallback=True, multizone=True for `stability` tests, but not for `benchmark` or `runtime`?

conftest.py
crusaderky: The db stores a link to the stdout/stderr of the run, and in that you have the dump of the kwargs

tests/stability/conftest.py
ntabris: there are some specific tests where I chose not to use spot because it seemed like the tests were more sensitive to cluster start time and there's the potential for some small impact from using spot + fallback (since we might get some spot, then need to make second request to get some on-demand, and I don't know if this is true but it's possible time to provision would be different on spot vs on-demand).

crusaderky: Cluster start time is not included in the test runtime measure though... unless spot instances are frequently forcibly shut down in the middle of a test?

conftest.py
jrbourbeau: This may very well be dead code. If you don't see it being used anywhere, feel free to remove

crusaderky:
Everything works as intended.
A/B test evidence is available at https://github.com/coiled/coiled-runtime/issues/551.
Ready for review and merge.

ncclementi:
This LGTM, I'm not sure if @jrbourbeau wants to give it a last look since he was involved in the initial review.

ncclementi (approved):

cluster_kwargs.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+# The keys 'name', 'environ', and 'tags' must not be used.
+
+# Settings for all clusters, unless overriden below
+default:
+  package_sync: true
+  wait_for_workers: true
+  scheduler_vm_types: [m6i.xlarge]
+  backend_options:
+    send_prometheus_metrics: true
```
ntabris, line 19: I'd remove this, it's now set (by me) at the account level.

ntabris: But also doesn't hurt anything if you want to leave it..

PR #557, opened 2022-11-29
Revert "OLS Workload" fix ci debugging code. 

ncclementi:
Reverts only changes in `test.yml` introduced in coiled/coiled-runtime#443

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,52 +31,52 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ["3.9"]
-        pytest_args: [tests/stability/test_array.py]
+        pytest_args: [tests]
         runtime-version: [upstream, latest, "0.0.4", "0.1.0"]
-        # include:
-        #   # Run stability tests on Python 3.8
-        #   - pytest_args: tests/stability
-        #     python-version: "3.8"
-        #     runtime-version: upstream
-        #     os: ubuntu-latest
-        #   - pytest_args: tests/stability
-        #     python-version: "3.8"
-        #     runtime-version: latest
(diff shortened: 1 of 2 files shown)


ncclementi:
It looks like we have a legit failure in windows https://github.com/coiled/coiled-runtime/actions/runs/3576035693/jobs/6013493204#step:5:269

ncclementi:
@gjoseph92 Do you know if this error is expected on windows? I'm not super familiar with the `scaled_array_shape` utility, but if this is expected to fail on windows?

gjoseph92:
It should not fail on windows and there's no reason why windows would be related to the failure

ncclementi:
> It should not fail on windows and there's no reason why windows would be related to the failure

Well I'm not sure what's the reason for the failure, it looks like the nbytes are not calculated properly. I do not have a way to investigate this without a windows machine, nor the time. I think I will skip the test on windows for now and open an issue about it.

gjoseph92:
> I think I will skip the test on windows for now and open an issue about it

Sure, but I don't see a reason why windows would be related to it. Unless it reproduces on windows I'm not sure why we'd skip it?

ntabris:
haven't looked closely so maybe this is unhelpful but should `actual_nbytes` ever be negative, or is the fact that it is mean there's likely overflow or something like that?

ncclementi:
> Sure, but I don't see a reason why windows would be related to it. Unless it reproduces on windows I'm not sure why we'd skip it?

It reproduces on windows at least on ci, I ran it 4 times, always failed on windows.

gjoseph92:
That's bizarre, thanks for re-running it and confirming. I guess it is some sort of overflow issue on windows? Maybe in the `np.prod`?

PR #562, opened 2022-11-30
Tweak parallelism from A/B config

### LINKED ISSUE #483: Configure number of workers and instance types in A/B tests

crusaderky:
- Partially closes #483 

When one significantly increases the cluster size (#549), the current hardcoded levels of parallelism can cause Coiled to fail to allocate all the necessary workers.

Additionally, pytest-xdist tends to eat the pytest output, making debugging more difficult.
Expose parallelisation levels as knobs in the A/B config file and optionally disable pytest-xdist completely.


ntabris:
What sort of cluster size are you talking about? Do you have example of test runs where Coiled failed to provision the desired number of instances?

crusaderky:
@ntabris in https://github.com/coiled/coiled-runtime/pull/355 we ascertained that AWS DoS protection will kick in beyond a certain, unspecified threshold.

#483 is about having the option to play with 1,000 workers clusters.
This PR makes the difference between 1,000 instances at once and 20,000.

ntabris:
I don't know what you mean by AWS DoS, there are lots of limits and I don't see anything in #355 that describes what you were seeing.

Maybe let's talk before you start making multiple 1000 worker clusters. That's fine to do occasionally, but we should make sure we aren't doing this way more than necessary, and 20 1000 worker clusters is a lot (even if you're making them one at a time).

crusaderky:
My above comment also applies to 100 workers. This PR makes the difference between 100 workers and 2000 workers.

ntabris:
Sure, I'm not saying this PR has a problem, I'm just saying me/platform should be in the loop when you're thinking about doing larger scale testing.

AB_environments/config.yaml
```diff
@@ -14,3 +14,16 @@ categories:
...
+# create the Coiled clusters, and restarting failed jobs won't fix the problem.
+# Additionally, there are problems with Coiled itself triggered by limitations that are
+# never actually reached with real paying users.
+max_parallel:
+  # Number of parallel A/B test jobs per branch.
+  ci_jobs: 5
+  # Number of parallel test_*.py modules per A/B test job.
+  # Each module typically spawns one Coiled cluster at a time.
+  # Set to 1 to disable pytest-xdist.
+  modules_per_job: 4
```
hendrikmakait, line 29: nit: I first ran into the variable name without the description and it took me a moment to figure out how the number of modules would impact parallelism. Would it make sense to rename this to something like `pytest_workers_per_job` to highlight that this is about `pytest-xdist` parallelism?

AB_environments/README.md
```diff
@@ -154,6 +154,9 @@ categories:
   - runtime
   - benchmarks
   - stability
+max_parallel:
```
ncclementi, line 157: The same way we explain opther parameters (e.g category, test_null_hypothesis, etc ) in the README, can we explain these two in the README as well? 

I know there are some comments in the `config.yml` but it'll be useful having them explain in the readme too.

ncclementi (approved):
LGTM, just left a small comment

AB_environments/config.yaml
crusaderky: done

AB_environments/README.md
crusaderky: done

PR #565, opened 2022-11-30
Scale tests with cluster size

### LINKED ISSUE #483: Configure number of workers and instance types in A/B tests

crusaderky:
Partially closes #483 

Review all tests so that they retain a pseudo-constant runtime as the cluster size goes up. Disable those that can't for very large clusters.

Demo with 2x[1] and 5x[2] cluster sizes:
https://github.com/coiled/coiled-runtime/actions/runs/3584320320
[1] 20 workers on small_cluster, 30 workers on parquet_cluster
[2] 50 workers on small_cluster, 75 workers on parquet_cluster

### DIFF
--- a/tests/benchmarks/test_work_stealing.py
+++ b/tests/benchmarks/test_work_stealing.py
@@ -10,7 +10,10 @@ from distributed import Client
 from packaging.version import Version
 from tornado.ioloop import PeriodicCallback
 
+from ..utils_test import run_up_to_nthreads
 
+
+@run_up_to_nthreads("small_cluster", 50, reason="fixed dataset")
 def test_trivial_workload_should_not_cause_work_stealing(small_client):
     root = delayed(lambda n: "x" * n)(utils.parse_bytes("1MiB"), dask_key_name="root")
     results = [delayed(lambda *args: None)(root, i) for i in range(10000)]
@@ -18,6 +21,7 @@ def test_trivial_workload_should_not_cause_work_stealing(small_client):
     small_client.gather(futs)
 
 
+@run_up_to_nthreads("small_cluster", 50, reason="fixed dataset")
 @pytest.mark.xfail(
     Version(distributed.__version__) < Version("2022.6.1"),
     reason="https://github.com/dask/distributed/issues/6624",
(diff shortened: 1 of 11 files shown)


ntabris:
Often I want to run a consistent workload on different sized clusters, in order to compare cost. Please make sure that's still easy, and that it's not too hard to tell for given test run what the exact workload was. (FWIW I'd lean toward parameterizing tests rather than have test just internally decide to do different things on different hardware.)

crusaderky:
> Please make sure that's still easy

**Most** tests in main are already scaling their workload to the cluster size.
This PR just makes it organic throughout **all** tests.

tests/benchmarks/test_work_stealing.py
```diff
@@ -10,14 +10,18 @@
...
+@run_up_to_nthreads("small_cluster", 50, reason="fixed dataset")
 def test_trivial_workload_should_not_cause_work_stealing(small_client):
     root = delayed(lambda n: "x" * n)(utils.parse_bytes("1MiB"), dask_key_name="root")
     results = [delayed(lambda *args: None)(root, i) for i in range(10000)]
     futs = small_client.compute(results)
     small_client.gather(futs)
 
 
+@run_up_to_nthreads("small_cluster", 50, reason="fixed dataset")
 @pytest.mark.xfail(
```
ncclementi, line 25: Since we are working with versions  `> 2022.11.1`  should we remove this. We can do it in a separate PR if it's out of scope.

ncclementi (approved):
This LGTM. 
I have one comment/question for learning purposes: How was the choice made for the max value the n_threads (50 or 100) in each case?

tests/benchmarks/test_work_stealing.py
crusaderky: Yes definitely out of scope

crusaderky:
> I have one comment/question for learning purposes: How was the choice made for the max value the n_threads (50 or 100) in each case?

It was from observing that there was virtually no change between 40 and 100 threads

Issue #526, opened 2022-11-14
Release 0.2.0


jrbourbeau:
https://github.com/coiled/coiled-runtime/pull/488 bumped package versions in preparation for a new runtime release. Given https://github.com/coiled/coiled-runtime/issues/520 we should try to release ASAP. @ncclementi are you aware of any outstanding issues?

ncclementi:
Yes, we have a couple of problems:
1. We are seeing this constantly https://github.com/coiled/coiled-runtime/issues/468 
2. There are valid regressions and we do not know where do they come from see https://github.com/coiled/coiled-runtime/issues/500 more specifically https://github.com/coiled/coiled-runtime/issues/500#issuecomment-1314505305

ncclementi:
Based on the comment above plan of action:
1. Do a patch release (0.1.1) of runtime pinning bokeh to avoid dashboard problems. 
2. Keep studying problems to bump all the packages for release 0.2.0 see https://github.com/coiled/coiled-runtime/issues/500 and https://github.com/coiled/coiled-runtime/issues/468

jrbourbeau:
That sounds like a good plan to me. I'd suggest pinning `bokeh=2.4.3`

ntabris:
this means that, e.g., test shuffle got ~30% slower (with high degree of confidence)?

hayesgb:
@ncclementi 

As we discussed -- running another A/B test that looks at bumping everything excepting the `fsspec` packages and then making a decision seems OK.

Let's also confirm you can successfully complete a `conda build` with both of the build candidate, and the one you intend to test. 

Let's aim for a decision by midday tomorrow.

If `pyarrow==10.0.x`makes its way onto `condo-forge` in early December (as expected) we can do a patch release.

gjoseph92:
Shuffle specifically we can expect to be slower (30% sounds about right to me) with queuing on by default.

If we're worried about versions of (transitive) dependencies having affected performance, @ncclementi maybe you should run the same A/B test, but turn off queuing for 0.2.0. If that looks the same/better, then we can attribute the wall-clock difference to queuing.

ncclementi:
Here is the link to the AB test with no queuing: 

Wall time
![Screen Shot 2022-11-30 at 8 38 16 PM](https://user-images.githubusercontent.com/7526622/204945117-c30d1a38-cca4-4657-a6e1-d9b31f488627.png)

Memory
[...]

PR #570, opened 2022-12-01
Update packages for release 0.2.0

ncclementi:
This PR updates the packages for the 0.2.0 release. I was able to conda-build locally, as well as install all the updated packages with a conda env. 

After this PR is merged, I will proceed with the release by bumping the version in the `setup.py` and the following steps for both pypi and conda-forge. 

For the regressions discussions check https://github.com/coiled/coiled-runtime/issues/526

cc: @hayesgb

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set name = "coiled-runtime" %}
-{% set version = "0.2.0" + environ.get("VERSION_SUFFIX", '') %}
-{% set dask_version = environ.get("DASK_VERSION", "2022.6.0") %}
-{% set distributed_version = environ.get("DISTRIBUTED_VERSION", "2022.6.0") %}
+{% set version = "0.3.0" + environ.get("VERSION_SUFFIX", '') %}
+{% set dask_version = environ.get("DASK_VERSION", "2022.11.1") %}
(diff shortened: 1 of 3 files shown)


ncclementi:
The failure on the pip build is because there was a release of coiled that is already up on pypi but not on conda-forge yet. I would ignore it for the time being. 

```
E       AssertionError: assert <Version('0.2.51')> == <Version('0.2.50')>

tests/runtime/test_build.py:115: AssertionError
```

recipe/meta.yaml
```diff
@@ -1,7 +1,7 @@
 {% set name = "coiled-runtime" %}
-{% set version = "0.2.0" + environ.get("VERSION_SUFFIX", '') %}
-{% set dask_version = environ.get("DASK_VERSION", "2022.6.0") %}
-{% set distributed_version = environ.get("DISTRIBUTED_VERSION", "2022.6.0") %}
+{% set version = "0.3.0" + environ.get("VERSION_SUFFIX", '') %}
```
phobson, line 2: Are we releasing 0.2.0 or 0.3.0?

ncclementi: This is for the nightlies, we usually put it one more up than the release it'll happen.

hayesgb (approved):
LGTM.  Based on conversation here and in #526 I would move forward.

PR #590, opened 2022-12-06
Fix `test_access_slices` - wait for persist call

### LINKED ISSUE #589: `test_access_slices` not doing anything
In this tests https://github.com/coiled/coiled-runtime/blob/69c820bf872eb20af88412674db3a6890469bb58/tests/benchmarks/test_array.py#L251-L256 

We are persisting but not waiting, hence the times are pretty much zero. 

Screen Shot 2022-12-06 at 10 23 37 AM

ncclementi:
- Closes #589

cc: @hendrikmakait


jrbourbeau (commented):
Good catch -- this looks reasonable to me

PR #595, opened 2022-12-06
Test and process result only for latest, upstream and 0.2.1

ncclementi:
- Closes https://github.com/coiled/coiled-runtime/issues/586

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,7 +32,7 @@ jobs:
         os: [ubuntu-latest]
         python-version: ["3.9"]
         pytest_args: [tests]
-        runtime-version: [upstream, latest, "0.1.0", "0.2.0"]
+        runtime-version: [upstream, latest, "0.2.1"]
         include:
           # Run stability tests on Python 3.8
           - pytest_args: tests/stability
@@ -45,11 +45,7 @@ jobs:
             os: ubuntu-latest
           - pytest_args: tests/stability
             python-version: "3.8"
-            runtime-version: "0.1.0"
-            os: ubuntu-latest
-          - pytest_args: tests/stability
-            python-version: "3.8"
-            runtime-version: "0.2.0"
+            runtime-version: "0.2.1"
             os: ubuntu-latest
           # Run stability tests on Python 3.10
           - pytest_args: tests/stability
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -274,7 +266,7 @@ jobs:
           environment-file: ci/environment-dashboard.yml
 
       - name: Generate dashboards
-        run: python dashboard.py -d benchmark.db -o static -b coiled-latest-py3.9 coiled-0.1.0-py3.9
+        run: python dashboard.py -d benchmark.db -o static -b coiled-latest-py3.9 coiled-0.2.1-py3.9
```
ncclementi, line 269: I think this change is needed, but I can be wrong, 
@crusaderky please correct me if I'm wrong.

ncclementi:
In the PR, the artifact, only builds this, 

Screen Shot 2022-12-07 at 10 45 18 AM


but I do not know if on main, because we have the previous deploys it'll keep those too. Does anyone know how the deployment works?

ncclementi:
The error are due to a known cause see https://github.com/coiled/coiled-runtime/issues/521

.github/workflows/tests.yml
crusaderky: Yes. The intent is to show the differences of `latest` and `upstream` vs. the latest published version.

dashboard.py
```diff
@@ -716,7 +716,8 @@ def main() -> None:
     engine = sqlalchemy.create_engine(f"sqlite:///{args.db_file}")
     df = pandas.read_sql(
         "select * from test_run where platform = 'linux' "
-        "and call_outcome in ('passed', 'failed')",
+        "and call_outcome in ('passed', 'failed')"
+        "and coiled_runtime_version >= '0.2.1' ",
```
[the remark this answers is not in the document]
crusaderky: I don't think I like this.
If we don't care about 0.0.4 and 0.1.0 anymore, shouldn't we wipe the database of them with an alembic rule instead?

.github/workflows/tests.yml
crusaderky, line 269:
```suggestion
        # Show differences of upstream vs. latest unpublished vs. latest publshed
        run: python dashboard.py -d benchmark.db -o static -b coiled-latest-py3.9 coiled-0.2.1-py3.9
```

dashboard.py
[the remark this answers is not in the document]
ncclementi: My understanding is that we do not want to run them anymore, but we would probably want to keep the history. That's why I was not touching the database but the plotting part only.

crusaderky (approved):

PR #597, opened 2022-12-07
Test P2P Shuffle in integration tests

### LINKED ISSUE #569: Test P2P Shuffle in integration tests
Parametrize tests to use the `tasks` and `p2p` shuffle

* [x] `test_shuffle`
* [x] `test_join` 
    * [ ] BONUS: Enable larger data sizes for `p2p`
* [x] `test_h2o_benchmarks`

Only run for `2022.11.0` and larger (which includes https://github.com/dask/distributed/commit/9c6904df38495e90af776aecb24f4a495ebda028)

hendrikmakait:
* [x] Closes #569


conftest.py
```diff
@@ -607,7 +607,7 @@ def _upload_cluster_dump(client):
     yield _upload_cluster_dump
 
 
-P2P_AVAILABLE = parse_version(dask.__version__) >= parse_version("2022.11.0")
+P2P_AVAILABLE = Version(dask.__version__) >= Version("2022.11.0")
```
ncclementi, line 610: Just a heads up, I'm working on a PR ([here](https://github.com/coiled/coiled-runtime/pull/595)) to only test `runtime >=0.2.1` which will cover this case as the dask version is 2022.11.1.

hendrikmakait:
Skipping larger data size for join tests due to data transfer to client which OOM-kills the pytest workers (#633).

hendrikmakait:
Ready for review. cc @crusaderky, @ncclementi

tests/benchmarks/test_join.py
```diff
@@ -1,52 +1,51 @@
...
-    ),
-    pytest.param(
-        10,
-        marks=pytest.mark.skip(reason="Does not finish"),
-    ),
+params = [
+    (0.1, "tasks"),
+    # shuffling takes a long time with 1 or higher
+    (0.1, "p2p"),
+    # (1, "p2p"),
```
hendrikmakait, line 11: #633 should allow us to enable this in a follow-up

tests/benchmarks/test_join.py
```diff
@@ -1,52 +1,51 @@
...
-    0.1,
-    pytest.param(
-        1,
-        marks=pytest.mark.skip(reason="Does not finish"),
-    ),
-    pytest.param(
-        10,
-        marks=pytest.mark.skip(reason="Does not finish"),
-    ),
+params = [
```
ncclementi, line 7: Can we put a comment that says what these params are?  I know you can read it in the `pytest.mark()` but it helps, before it was clear because it was just `mem_mult`, something like
```suggestion
# (mem_mult, shuffle)
params = [
```

tests/benchmarks/test_join.py
ncclementi: Shouldn't we mark them then as `marks=pytest.mark.skip(reason="Does not finish, fix when #633 is merged")`

If not, open an issue to state that when the #633 is merged we need to uncomment this

ncclementi (commented):
Left some small comments.

My only concern is that we will be running each of the h2o queries 9 times (csv, pq0.5GB, pq5GB  * no_shuffle, task, p2p) This will increase the time of the whole CI, a lot. For most of the queries and this size of data shuffle and p2p might make no sense. 

Shouldn't we just move to only 1 biggish pq dataset (thinking of the 50GB) and run this only maybe once a week. What would imply running it as a separate workflow probably.

hendrikmakait:
> My only concern is that we will be running each of the h2o queries 9 times (csv, pq0.5GB, pq5GB * no_shuffle, task, p2p) This will increase the time of the whole CI, a lot. For most of the queries and this size of data shuffle and p2p might make no sense.

FWIW, we only run each query 6 times as we do not test the implicit default aka "no_shuffle". IIUC, `tasks` is the default behavior when using a distributed cluster, we just make it explicit now and do not need to test "no_shuffle". While there shouldn't be a difference in performance, we want to avoid possible performance regressions. Regarding frequency/increased CI runtime, I've had a chat with @fjetter about this and while `tasks` is the default behavior and `p2p` is under active development, we feel like it would be a good idea to test both of them frequently.  Note that _until the next release_, `p2p` is only tested for `upstream`. 

> Shouldn't we just move to only 1 biggish pq dataset (thinking of the 50GB) and run this only maybe once a week. What would imply running it as a separate workflow probably.

Adding testing on a large-scale dataset is yet another thing we should add, but that's out of scope for this PR.

hendrikmakait:
> Shouldn't we mark them then as marks=pytest.mark.skip(reason="Does not finish, fix when #633 is merged")

Good point, I meant to do that after creating #633.

tests/benchmarks/test_join.py
```diff
@@ -1,52 +1,57 @@
...
-    0.1,
+# (mem_mult, shuffle)
+params = [
+    (0.1, "tasks"),
+    # shuffling takes a long time with 1 or higher
+    (0.1, "p2p"),
     pytest.param(
-        1,
-        marks=pytest.mark.skip(reason="Does not finish"),
+        (1, "p2p"), marks=pytest.mark.skip(reason="client OOMs, see coiled-runtime#633")
```
hendrikmakait, line 13: XREF: #633

tests/benchmarks/test_join.py
```diff
@@ -1,52 +1,57 @@
...
     ),
     pytest.param(
-        10,
-        marks=pytest.mark.skip(reason="Does not finish"),
+        (10, "p2p"),
+        marks=pytest.mark.skip(reason="client OOMs, see coiled-runtime#633"),
```
hendrikmakait, line 17: XREF: #633

ncclementi:
@hendrikmakait We are having some legit failures in https://github.com/coiled/coiled-runtime/actions/runs/3766959977/jobs/6404088081

```python-traceback
________________ ERROR collecting tests/benchmarks/test_join.py ________________
tests/benchmarks/test_join.py::test_join_big: in "parametrize" the number of names (2):
  ['mem_mult', 'shuffle']
must be equal to the number of values (1):
  ((1, 'p2p'),)
```

The ones in https://github.com/coiled/coiled-runtime/actions/runs/3766959977/jobs/6404088006 are platform related.

ncclementi:
@hendrikmakait I'm trying different combinations here https://github.com/coiled/coiled-runtime/pull/634, but haven't been able to make it work. 

That PR, runs only that test so it's easier to debug. in case you want to mess in there.

hendrikmakait:
@ncclementi: I think I found the solution (interesting that some stuff seems to work but then fails when it comes to running the specific parametrization (or ignoring it). 

Sorry for causing this, I should've really tested "that one last beautification" locally.

ncclementi:
@hendrikmakait no worries, I've been playing with it trying to fix it because I thought you were off for today. I'm testing it over here too as it should fail faster if it doesn't work https://github.com/coiled/coiled-runtime/pull/634

It's very unintuitive :/

ncclementi (approved):

ncclementi:
Ci is green, 
Thanks @hendrikmakait this goes in!

PR #609, opened 2022-12-12
Update to `coiled>=0.2.54` for `package_sync` fix

hendrikmakait:
Extracted from #601


hendrikmakait:
To fix CI, we still need the additional changes in #609. This change has primarily been extracted for visibility.

ncclementi (approved):

PR #624, opened 2022-12-16
Bump std in regressions detection to 3

### LINKED ISSUE #623: Relax regression detection in main
Currently, in main we are raising a regression if the last 3 values go over 2 std. 
https://github.com/coiled/coiled-runtime/blob/e031cf43d06ac5804b96d5aacc87d768858d5591/detect_regressions.py#L46-L48

This is very harsh, let's relax it a bit for now. 

Note: If this doesn't improve CI failures we will turn off regression.

ncclementi:
- Closes #623


hayesgb (approved):
Lgtm

PR #625, opened 2022-12-19
Switch scheduler to m6i.large (2 vCPU) from m6i.xlarge (4 vCPU)

ntabris:
Let's use small scheduler since that doesn't appear to impact perf.

(IIRC we switched from large to xlarge back when we were using t3's, I wouldn't be surprised if it did matter for those.)

### DIFF
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -14,7 +14,7 @@
 default:
   package_sync: true
   wait_for_workers: true
-  scheduler_vm_types: [m6i.xlarge]
+  scheduler_vm_types: [m6i.large]
   backend_options:
     send_prometheus_metrics: true
     spot: true


ntabris:
Schedulers are always on demand. Currently each CI run costs ~$2.40 for on demand `m6i.xlarge`. If we use `m6i.large`, cost is (not surprisingly) ~$1.20.

(Other instances for CI run come to ~$2.70, so currently the scheduler is almost half the cost of cluster. Those instances are much cheaper because they're spot.)

ntabris:
If I'm reading the A/B results correctly it looks like smaller scheduler is usually faster (?!). A couple cases were slower but not by much and not with much confidence.

I think it makes sense to use `m6i.large` rather than `m6i.xlarge`. That sound okay?

jrbourbeau (commented):
Thanks @ntabris! This seems reasonable to me. @hayesgb @crusaderky do you have thoughts here? 

> If I'm reading the A/B results correctly it looks like smaller scheduler is usually faster (?!)

Hmm, hooray? This seems interesting to me

crusaderky:
> If I'm reading the A/B results correctly it looks like smaller scheduler is usually faster (?!). 

No. The null hypothesis is indistinguishable from AB_large, which means everything you think you're seeing is just noise.

ntabris:
Yeah, I don't understand how to read this.

I agree that people expect more cores not to matter for scheduler, but I've also seen that to be false, at least on t3.

crusaderky (approved):

PR #629, opened 2022-12-20
Scale test_spill.py; test highly compressible data

crusaderky:
- Closes https://github.com/dask/distributed/issues/7351
- Add use case to test_spill.py where the data is highly compressible. This is expected to increase pickle time and reduce spill/unspill time.
- Scale test_spill.py with the number of workers
- Speed up dot product test
- Move test module to benchmarks

# Demo
A/B tests with 5 workers (baseline / null hypothesis), 10 workers (2x), 20 workers (4x):
image

### test_dot_product_spill[uncompressible]
image

### test_dot_product_spill[compressible]
image

These results are very interesting - I was *not* expecting to see this much disparity between pickle and unpickle!


crusaderky:
Blocked by https://github.com/coiled/coiled-runtime/issues/630

tests/benchmarks/test_spill.py
```diff
@@ -0,0 +1,91 @@
...
+@pytest.mark.parametrize(
+    "compressible",
+    [pytest.param(False, id="uncompressible"), pytest.param(True, id="compressible")],
+)
+def test_dot_product_spill(spill_client, compressible):
+    """See also test_array.py::test_dot_product
+    for variant that doesn't hit the spill threshold
+    """
+    memory = cluster_memory(spill_client)  # 38.33 GiB
+    shape = scaled_array_shape_quadratic(memory * 0.3, "11.5 GiB", ("x", "x"))
```
crusaderky, line 82: Reduced from 18 GiB

tests/utils_test.py
```diff
@@ -79,15 +78,21 @@ def scaled_array_shape(
     final = tuple(s for s in resolved_shape if s is not None)
     assert len(final) == len(resolved_shape), resolved_shape
 
-    actual_nbytes = np.prod(final) * dtype.itemsize
+    actual_nbytes = math.prod(final) * dtype.itemsize
```
crusaderky, line 81: Fix integer overflow bug in Windows

tests/utils_test.py
```diff
@@ -104,6 +109,18 @@ def scaled_array_shape_quadratic(
     return scaled_array_shape(scaled_nbytes, shape, dtype=dtype, max_error=max_error)
 
 
+def print_size_info(memory: int, target_nbytes: float, *arrs: da.Array) -> None:
+    print(
+        f"Cluster memory: {format_bytes(memory)}, "
+        f"target data size: {format_bytes(int(target_nbytes))}"
```
crusaderky, line 115:
```suggestion
        f"target data size: {format_bytes(target_nbytes)}"
```

crusaderky:
Ready for final review and merge

fjetter:
The spill tests were always a bit slow/heavy. I'm worried that we're introducing now more of them. IIUC the runtimes of the new ones are not significantly faster than before, are they? Can we do something to make them all a bit faster?

tests/benchmarks/test_spill.py
```diff
@@ -0,0 +1,90 @@
...
+@pytest.mark.parametrize(
+    "keep_around", [pytest.param(False, id="release"), pytest.param(True, id="keep")]
+)
+@pytest.mark.parametrize(
+    "compressible",
+    [pytest.param(False, id="uncompressible"), pytest.param(True, id="compressible")],
+)
+def test_spilling(spill_client, compressible, keep_around):
+    memory = cluster_memory(spill_client)  # 36 GiB
+    shape = scaled_array_shape(memory * 1.79, ("x", "x"))  # 64 GiB
```
hendrikmakait, line 56: `memory * 1.79` is oddly specific yet not understandable without the comment. Why not simply use a constant? Is this to enable benchmarking larger clusters with linearly scaling data sizes?

tests/benchmarks/test_spill.py
```diff
@@ -0,0 +1,90 @@
...
+    with Client(spill_cluster) as client:
+        spill_cluster.scale(n_workers)
+        client.wait_for_workers(n_workers)
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+@pytest.mark.parametrize(
+    "keep_around", [pytest.param(False, id="release"), pytest.param(True, id="keep")]
```
hendrikmakait, line 48: Nice readability tweak!

hendrikmakait (approved):
The code generally looks good to me. I share @fjetter's concern about moving these tests to the benchmarks wrt. A/B testing. However, I think that the distinction between benchmarks and stability tests is an artificial one and we may be better served by making it configurable which tests to run in an A/B test. For example, if A/B testing some change that's limited to arrays, I don't need to run H2O benchmarks, or benchmark parquet writes. (Obviously, this is out of scope for this PR.) 

Also, from what I see, the new tests we add take up a limited amount of time compared to the ones that already exist. By reducing the spilling tensordot test size this PR seems to yield a net improvement overall.

tests/benchmarks/test_spill.py
crusaderky: It is to match the previous implementation, so that we don't need to declare bankruptcy on the historical measures.

crusaderky:
@fjetter @hendrikmakait I tried reducing the size of `test_dot_product_spill` from 0.3 * memory to 0.2, but that sends it (just) below the spill threshold.

I can try reduce the number of workers from 5 to 2 and the problem size to match, but I suspect that will reduce data duplication too and thus *increase* overall runtime if the aim is to reach the point where the algorithm spills.

fjetter:
I'm not just concerned about A/B test but I'm concerned about the overall runtime in general. The CI runtime of this run was 47min and the tests added here contribute ~18min, i.e. we spend 38% of the runtime testing spilling. This is not representative, spilling is not *that* important. Overall I find it hard to believe that it is not possible to write something that uses disk but is faster.

- test_spill is nice but considering that it isn't doing anything interesting it is hard to justify 5min runtime
- test_tensordot_stress might be more representative but also a workload (tensordot) that is not super relevant for us right now.

I'm OK with merging this since it actually reduces total runtime (tensordot_stress takes 20min alone on main) but I still believe we should consider dropping these tests entirely or replace them with something better.

crusaderky:
> I'm not just concerned about A/B test but I'm concerned about the overall runtime in general. The CI runtime of this run was 47min and the tests added here contribute ~18min

I think you have it backwards.
### BEFORE (main):
- test_spilling runs in stability in 260+240s
- test_tensordot_stress runs in stability in 1100s
total 1600s per job

stability jobs:
- ubuntu [0.2.1, latest, upstream] [py3.8, 3.9, 3.10]
- [windows, macosx] latest 3.9
total 11 jobs per run
total 1600 * 11 = 17600s per run

### AFTER (this PR):
- test_spilling runs in benchmarks in 250 + 150 + 40 + 40s
- test_dot_product_spill runs in benchmarks in 420 + 360s
total 1260s per job

benchmarks jobs:
- ubuntu [0.2.1, latest, upstream] py3.9
total 3 jobs per run
total 1260 * 3 = 3780s per run

crusaderky:
@fjetter any objections to merging?

fjetter:
> I think you have it backwards.

Maybe I wrote the wrong thing. What I was trying to say

- These tests are slow and I don't love them
- Your PR is actually faster than main which is good
- Since the PR is faster we should merge it. In a follow up, we should consider dropping them or changing them to be even faster.

PR #632, opened 2022-12-23
Remove `test_shuffle.py`

hendrikmakait:
Removes `test_shuffle.py` from the stability tests:

`test_shuffle_simple` should be superseded by `benchmarks.test_dataframe.test_shuffle` which tests on a larger amount of data.

`test_shuffle_parquet` should be superseded by `benchmarks.test_dataframe.test_shuffle` as well as `benchmarks.test_parquet.test_write_wide_data`.


fjetter:
do we want to drop these from the DB as well?

hendrikmakait:
> do we want to drop these from the DB as well?

I'd support that, but since it's an irreversible change with our current setup (i.e., no soft deletes), I wanted to remove the tests first, see if someone complains, and then drop execution history from the DB.

fjetter (approved):

PR #644, opened 2023-01-03
Remove `shuffle` parametrization from H2O benchmarks that do not perform a shuffle

### LINKED ISSUE #639: Remove `shuffle` parametrization from H2O benchmarks that do not perform a shuffle
As it turns out, most H2O groupby benchmark queries do not perform shuffles. As a consequence, the parametrization should be removed (and the respective DB entries cleaned). Additionally, we should add a comment that explains why only a subset of the tests is parametrized with the shuffle method. (It is because we do not specify `split_out`, which defaults to 1 and then does not perform a shuffle on a groupby. groupby.apply and holistic aggregations (i.e., median) always perform a shuffle).

hendrikmakait:
* Closes #639


alembic/versions/a9363331e323_clean_h2o_tests_with_removed_shuffle_.py
```diff
@@ -0,0 +1,35 @@
...
+Revises: c38b9d85915e
+Create Date: 2023-01-03 19:56:22.838577
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'a9363331e323'
```
fjetter, line 13: Likely need to coordinate merges with https://github.com/coiled/coiled-runtime/pull/632 and https://github.com/coiled/coiled-runtime/pull/643

Instead we can also drop the migrations from both PRs and do both changes in one alembic migration

hendrikmakait:
The migration has been coordinated with #632 and is ready for review/merge.

hendrikmakait:
CI failure is unrelated

crusaderky (approved):

PR #645, opened 2023-01-03
Enable larger join tests

### LINKED ISSUE #641: Enable larger joins for `p2p` in `test_join.py`
Now that #633 has been merged, we can test larger joins (https://github.com/coiled/coiled-runtime/blob/2e0f4a6f395868acf20db960588b64d4d1f8c80a/tests/benchmarks/test_join.py#L12-L17)

hendrikmakait:
* Closes #641


hendrikmakait:
@ncclementi: I still have to check why the tests don't run on CI; likely the pytest worker gets OOM-killed.

hendrikmakait:
@ncclementi: As it turns out, one issue with the tests was that the partition size was very small (~8 MiB). I've adjusted the tests now to use larger partitions, let's see what CI says.

hendrikmakait:
I've also dropped the 10x join since it would mean spilling 10-20x the data to disk which would slow tests down significantly. We could (and should) add the occasional BIG data integration test run.

hendrikmakait:
`test_join_big` shows unnecessarily high memory usage with `p2p` due (dask/distributed#7496). It is still performing better, so I count that as a win.

`test_join_big_small` does not show any effect due to early materialization of the smaller dataframe which circumvents a distributed join (#669).

ncclementi:
> `test_join_big` shows unnecessarily high memory usage with `p2p` due ([dask/distributed#7496](https://github.com/dask/distributed/issues/7496)). It is still performing better, so I count that as a win.

This makes sense. Hopefully, when things get fixed we should see a nice drop in the memory.

tests/benchmarks/test_join.py
```diff
@@ -24,17 +19,21 @@ def test_join_big(small_client, mem_mult, shuffle):
...
 
-        df2_big = timeseries_of_size(memory * mem_mult)
+        df2_big = timeseries_of_size(
+            memory * mem_mult, dtypes={str(i): float for i in range(100)}
+        )  # 66.58 MiB partitions
 
         # Control cardinality on column to join - this produces cardinality ~ to len(df)
-        df2_big["x2"] = df2_big["x"] * 1e9
-        df2_big = df2_big.astype({"x2": "int"})
+        df2_big["predicate"] = df2_big["0"] * 1e9
```
ncclementi, line 33: Out of curiosity, why did you choose the name "predicate"?

ncclementi (approved):
This LGTM, 
Only comment is that we should probably add a separate issue to track the case for size=10 as part of integration tests, and then mentioned it on this PR just for completion.

tests/benchmarks/test_join.py
hendrikmakait: That's me coming from a DB background. I wanted a name that's more descriptive in this context than `x2` and since it's the column used in the _join predicate_ (i.e., the expression used to merge the tables/dataframes), that's what I ended up with.

ncclementi: I see. That's good, no need to change it, it was more out of curiosity and I just learn something new :) 
I'm merging this in.

PR #655, opened 2023-01-09
Choose H2O datasets in A/B tests

### LINKED ISSUE #654: Drop CSV and 0.5GB parquet H20 benchmarks
We're currently executing all H20 benchmarks on three different datasets

- 0.5 GB CSV
- 0.5 GB Parquet
- 5 GB Parquet

I believe that the CSV tests do not add significant value and the 0.5GB parquet dataset is simply too small for any optimizations to truly show.

I do not believe they hurt is in terms of runtime too much but they do introduce a certain bias. Looking at an A/B test, for example, they make up ~50% of all of our tests but historically their influence on actual decision making was very small

image

I suggest to drop all 0.5GB tests from our suite.

(At the very least we should exclude them by default from A/B tests to reduce bias and make results more readable)

crusaderky:
- Closes #654
- Disable 0.5 GB datasets by default in A/B
- Simplify running 50 GB tests (cc @hayesgb )

download


AB_environments/README.md
```diff
@@ -115,6 +115,9 @@ setting is high enough.
 The file offers a `categories` list. These are the subdirectories of `tests/`
 which you wish to run.
 
+`h2o_datasets` is a list of datasets to run through in
+`tests/benchmarks/test_h2o.py`. Refer to the file for the full list.
```
crusaderky, line 119:
```suggestion
`tests/benchmarks/test_h2o.py`. Refer to the file for the possible choices.
```

tests/benchmarks/test_h2o.py
```diff
@@ -1,36 +1,46 @@
...
+    "50 GB (parquet)": "s3://coiled-datasets/h2o-benchmark/N_1e9_K_1e2_parquet/*.parquet",
+}
+
+enabled_datasets = os.getenv("H2O_DATASETS")
+if enabled_datasets is not None:
+    enabled_datasets = {k.strip() for k in enabled_datasets.split(",")}
+    if unknown_datasets := enabled_datasets - DATASETS.keys():
+        raise ValueError("Unknown H2O dataset(s): ", unknown_datasets)
+else:
+    enabled_datasets = {"0.5 GB (csv)", "0.5 GB (parquet)", "5 GB (parquet)"}
```
crusaderky, line 27: This is what is used by the non-A/B tests.

@fjetter I think there is *some* value in leaving the 0.5 GB datasets enabled in the general tests. Namely, it shows potential regressions in CSV reading, and it helps appreciating the gap in read speed of parquet vs csv.
Happy to change this default if you have strong opinions though.

tests/benchmarks/test_h2o.py
```diff
@@ -1,42 +1,50 @@
...
+from ..utils_test import run_up_to_nthreads
+
+DATASETS = {
+    "0.5 GB (csv)": "s3://coiled-datasets/h2o-benchmark/N_1e7_K_1e2_single.csv",
+    "5 GB (csv)": "s3://coiled-datasets/h2o-benchmark/N_1e8_K_1e2_single.csv",
+    "50 GB (csv)": "s3://coiled-datasets/h2o-benchmark/N_1e9_K_1e2_single.csv",
+    "0.5 GB (parquet)": "s3://coiled-datasets/h2o-benchmark/N_1e7_K_1e2_parquet/*.parquet",
+    "5 GB (parquet)": "s3://coiled-datasets/h2o-benchmark/N_1e8_K_1e2_parquet/*.parquet",
+    "50 GB (parquet)": "s3://coiled-datasets/h2o-benchmark/N_1e9_K_1e2_parquet/*.parquet",
+}
```
ncclementi, line 22: We also have the datasets that have `id1`, `id2` and `id3` as pyarrow strings for 5GB, 50GB, and even 500GB. I can create 0.5GB if needed, the s3 uri are:

- 5GB uri: `s3://coiled-datasets/h2o-benchmark/pyarrow_strings/N_1e8_K_1e2/`
- 50GB uri: `s3://coiled-datasets/h2o-benchmark/pyarrow_strings/N_1e9_K_1e2/`
- 500GB: `s3://coiled-datasets/h2o-benchmark/pyarrow_strings/N_1e10_K_1e2/`

ncclementi (approved):
I left couple of comments about other available datasets and sizes that we have, I'll leave it up to you what you want to include or not. Otherwise this looks good to me.

ncclementi:
@crusaderky 
We have some conflicts now because I've merged the PR to run the individual A/B tests. 
If you can fix them and include the data paths for the cases that you consider useful for tests, this is good to go.

tests/benchmarks/test_h2o.py
fjetter: I believe there are plenty of benchmarks out there that show the diff between parquet and CSV.
I'm OK keeping them around if they are not too noisy/flaky

crusaderky:
The 50GB datasets consistently fail on a cluster of 200 m6i.large workers. I'd rather treat the proper investigation on what's needed to make them go through as out of scope for the purpose of this PR.

ntabris:
> The 50GB datasets consistently fail on a cluster of 200 m6i.large workers. I'd rather treat the proper investigation on what's needed to make them go through as out of scope for the purpose of this PR.

Which clusters?

I'm looking at all clusters with >100 `m6i.large` workers in the last two weeks and I only see one: [test_h2o-f9339777](https://cloud.coiled.io/dask-engineering/clusters/144993/details)

But I do see two clusters with **1200** `m6i.large` workers, [test_h2o-66b8ca9c](https://cloud.coiled.io/dask-engineering/clusters/145079/details) and [test_h2o-4f710b97](https://cloud.coiled.io/dask-engineering/clusters/145194/details).

Is it possible you were trying 1200 workers instead of 200?

ncclementi:
> The 50GB datasets consistently fail on a cluster of 200 m6i.large workers. I'd rather treat the proper investigation on what's needed to make them go through as out of scope for the purpose of this PR.

@crusaderky when you say they fail is this for all the datasets? 
I would expect this not to be great for the case when the data was saved as categories, and for the case of pyarrow strings, if the code stays the same, this implies that we are not casting the types upon reading, causing these columns to be read as python strings (this will imply way bigger memory usage)

But I have been able to successfully run the queries in https://github.com/coiled/h2o-benchmarks/blob/main/h2o_benchmarks_2022-10-0_pyarrow_str_50GB.ipynb with 10 t3.large workers.

crusaderky:
> @crusaderky when you say they fail is this for all the datasets? 

Complicated answer. After every test, you get a restart, but if a test kills off a worker the next, unrelated test will be messed up too. I've also seen failures in the cluster dump. So short of adding a flag that says "don't call client.restart, but instead nuke the whole cluster and rebuild it from scratch after every test" I'd need to kick off the tests one by one to answer you.
Also, with 200 workers it took more than 2 hours to run and it got killed off by github CI - thus no final report.

crusaderky:
> I'm looking at all clusters with >100 `m6i.large` workers in the last two weeks and I only see one with 200 workers: [test_h2o-f9339777](https://cloud.coiled.io/dask-engineering/clusters/144993/details)
> 
> But I do see two clusters with **1200** `m6i.large` workers, [test_h2o-66b8ca9c](https://cloud.coiled.io/dask-engineering/clusters/145079/details) and [test_h2o-4f710b97](https://cloud.coiled.io/dask-engineering/clusters/145194/details).
> 
> Is it possible you were trying 1200 workers instead of 200?

The overview for all the clusters you linked shows 200. There were 6x calls to `client.restart`, which restarted the worker, not the host - so 1200 worker processes, 200 at a time, on 200 hosts.

crusaderky:
@ncclementi am I clear to merge?

ntabris:
> The overview for all the clusters you linked shows 200. There were 6x calls to client.restart, which restarted the worker, not the host - so 1200 worker processes, 200 at a time, on 200 hosts.

Ick. Okay, yeah, I didn't see that you keep getting 200 new workers, thanks for pointing that out. That's not good that running this killed all your workers and also that those clusters ran for multiple hours.

If the result of merging this is that we'll regularly run clusters like the `test_h2o` clusters that I linked to, I don't think that's good and would ask you hold off on merging.

If this wouldn't be part of regular coiled-runtime runs, then I don't have any immediate concerns.

crusaderky:
No, this is to be explicitly enabled on a ad-hoc basis. We will normally run the 5gb datasets only.

ntabris:
Sounds good, thanks!

ncclementi:
@crusaderky it looks like the failure is unrelated, merging this one! 
Thank you for your work.

PR #675, opened 2023-01-26
 Test integration with xgboost and optuna

crusaderky:
This replicates notebooks "Feature engineering" and "Modelling 2" from https://github.com/coiled/dask-xgboost-nyctaxi, with the exception that it doesn't shuffle around the training and test splits (it would have slown down the test 5x and it wouldn't have been very interesting from a dask point of view).

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -51,6 +51,7 @@ requirements:
     - openssl >1.1.0g
     - optuna ==3.1.0
     - scipy ==1.10.0
+    - sqlalchemy ==1.4.46  # FIXME https://github.com/dask/dask/issues/9896
     - pynvml ==11.4.1
     - bokeh ==2.4.3
 
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -49,6 +49,8 @@ requirements:
     - xgboost ==1.7.1
     - dask-ml ==2022.5.27
     - openssl >1.1.0g
+    - optuna ==3.1.0
+    - scipy ==1.10.0
     - pynvml ==11.4.1
     - bokeh ==2.4.3
 
--- a/tests/benchmarks/test_xgboost.py
+++ b/tests/benchmarks/test_xgboost.py
@@ -1,0 +1,237 @@
+import dask.dataframe as dd
+import numpy as np
+import pandas as pd
+import pytest
+import xgboost.dask
+from dask_ml.metrics import mean_squared_error
+
+from ..utils_test import downscale_dataframe, run_up_to_nthreads, wait
+
+
+@pytest.fixture
+def taxi_zone_lookup():
+    # File from https://www.nyc.gov/site/tlc/about/tlc-trip-record-data.page
+    df = pd.read_csv(
+        "s3://coiled-runtime-ci/xgboost/taxi+_zone_lookup.csv",
+        usecols=["LocationID", "Borough"],
+    )
+    BOROUGH_MAPPING = {
+        "Manhattan": "Superborough 1",
+        "Bronx": "Superborough 1",
+        "EWR": "Superborough 1",
+        "Brooklyn": "Superborough 2",
+        "Queens": "Superborough 2",
+        "Staten Island": "Superborough 3",
+        "Unknown": "Unknown",
+    }
+    df["Superborough"] = [BOROUGH_MAPPING[k] for k in df["Borough"]]
+    df = df.astype({"Borough": "string[pyarrow]", "Superborough": "string[pyarrow]"})
+    return df
+
+
+@run_up_to_nthreads("small_cluster", 200, reason="fixed size dataset")
+def test_preprocess(small_client, taxi_zone_lookup, read_parquet_with_pyarrow):
+    """A typical workflow that preprocesses crude data into a ML-friendly dataframe"""
+    ############
+    # Read input
+    ############
+    ddf = dd.read_parquet(
+        "s3://coiled-runtime-ci/xgboost/nyc-uber-lyft-input.parquet",
+        index=False,
+        columns=[
+            "hvfhs_license_num",
+            "PULocationID",
+            "DOLocationID",
+            "trip_miles",
+            "trip_time",
+            "tolls",
+            "congestion_surcharge",
+            "airport_fee",
+            "wav_request_flag",
+            "on_scene_datetime",
+            "pickup_datetime",
+        ],
+    )
+
+    # Full size dataset on 40x r6i.large
+    ddf = downscale_dataframe(ddf, "580 GiB")
+
+    # The size of the partitions in the input dataset varies substantially, between 22
+    # and 836 MiB. For the purpose of this test we could just rewrite the input in more
+    # palatable chunks. However, having to deal with dishomogeneous and unwieldy input
+    # partitions is a very realistic use case, which we want to replicate.
+    # TODO https://github.com/dask/dask/issues/9849
+    # TODO https://github.com/dask/dask/issues/9850
+    # This line loads the whole dataset into memory and then discards it.
+    ddf = ddf.repartition(partition_size="100MB")
+
+    ####################
+    # Preprocess columns
+    ####################
+    ddf = ddf.assign(
+        accessible_vehicle=ddf.on_scene_datetime.isnull(),
+        pickup_month=ddf.pickup_datetime.dt.month,
+        pickup_dow=ddf.pickup_datetime.dt.dayofweek,
+        pickup_hour=ddf.pickup_datetime.dt.hour,
+    )
+    ddf = ddf.drop(columns=["on_scene_datetime", "pickup_datetime"])
+    ddf["airport_fee"] = ddf["airport_fee"].replace("None", 0).astype(float).fillna(0)
+
+    #############
+    # Filter rows
+    #############
+    ddf = ddf.dropna(how="any")
+    # Remove outliers
+    lower_bound = 0
+    Q3 = ddf["trip_time"].quantile(0.75)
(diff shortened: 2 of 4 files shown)


recipe/meta.yaml
```diff
@@ -49,6 +49,8 @@ requirements:
     - xgboost ==1.7.1
     - dask-ml ==2022.5.27
     - openssl >1.1.0g
+    - optuna ==3.1.0
```
ncclementi, line 52: Just a heads up, this workflow as in the notebooks it needs 
```
xgboost=1.7.3
optuna==3.1.0b0 #this is a nightly
```
and I believe we might need a newer version of dask and distributed. 
For reference this is the [env file](https://github.com/coiled/dask-xgboost-nyctaxi/blob/main/environment.yml) that Greg used

crusaderky: I ran the notebooks just fine with coiled-runtime 0.2.1 plus optuna 3.1.0 though.

tests/conftest.py
```diff
@@ -632,3 +633,30 @@ def shuffle(request):
 def configure_shuffling(shuffle):
     with dask.config.set(shuffle=shuffle):
         yield
+
+
+@pytest.fixture
+def read_parquet_with_pyarrow():
+    """Force dask.dataframe.read_parquet() to return strings with dtype=string[pyarrow]
+    FIXME https://github.com/dask/dask/issues/9840
+    """
```
crusaderky, line 642: CC @jrbourbeau at the moment this hacky fixture is used exclusively by the xgboost preprocess test.

recipe/meta.yaml
ncclementi: Ok then, maybe this was for the threading things. I remember @phobson mentioning this was needed.

tests/benchmarks/test_xgboost.py
```diff
@@ -0,0 +1,237 @@
...
+    return df
+
+
+@run_up_to_nthreads("small_cluster", 200, reason="fixed size dataset")
+def test_preprocess(small_client, taxi_zone_lookup, read_parquet_with_pyarrow):
+    """A typical workflow that preprocesses crude data into a ML-friendly dataframe"""
+    ############
+    # Read input
+    ############
+    ddf = dd.read_parquet(
```
ncclementi, line 38: Is there a reason why here you are not casting the strings to pyarrow strings?

tests/benchmarks/test_xgboost.py
```diff
@@ -0,0 +1,237 @@
...
+        "s3://coiled-runtime-ci/xgboost/nyc-uber-lyft-preprocessed.parquet",
+    )
+
+    # Full size dataset on 40x r6i.large
+    ddf = downscale_dataframe(ddf, "580 GiB")
+
+    # Under the hood, XGBoost converts floats to `float32`.
+    # Let's do it only once here.
+    float_cols = ddf.select_dtypes(include="float").columns.tolist()
+    ddf = ddf.astype({c: np.float32 for c in float_cols})
```
ncclementi, line 169: I think the data was already saved in float32, do we need explicitly np.float32? Not sure I follow this part.

tests/benchmarks/test_xgboost.py
```diff
@@ -0,0 +1,237 @@
...
+
+    # We need the categories to be known
+    categorical_vars = ddf.select_dtypes(include="category").columns.tolist()
+
+    # categorize() reads the whole input and then discards it.
+    # Let's read from disk only once.
+    ddf = ddf.persist()
+    ddf = ddf.categorize(columns=categorical_vars)
```
ncclementi, line 177: If the columns are already a category, is this line doing anything?
ddf.categorize converts into categories the columns, but they already are categories. https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.categorize.html

tests/benchmarks/test_xgboost.py
```diff
@@ -0,0 +1,237 @@
...
+    #########################################
+    def train_model(**study_params):
+        model = xgboost.dask.train(
+            None,
+            {"tree_method": "hist", **study_params},
+            d_train,
+            num_boost_round=4,
+            evals=[(d_train, "train")],
+        )
+        predictions = xgboost.dask.predict(None, model, x_test)
```
ncclementi, line 212: by looking at XGBoost examples we probably should we put a comment explaining why we are not generating the DMatrix for test. (btw I do not know the reason)

tests/benchmarks/test_xgboost.py
ncclementi: I just saw the fixture `read_parquet_with_pyarrow`. We should put a comment saying that we are doing this.

tests/benchmarks/test_xgboost.py
```diff
@@ -0,0 +1,237 @@
...
+    #############################
+    # Dataset load and preprocess
+    #############################
+    # Load feature table generated by previous test
+    ddf = dd.read_parquet(
+        "s3://coiled-runtime-ci/xgboost/nyc-uber-lyft-preprocessed.parquet",
+    )
+
+    # Full size dataset on 40x r6i.large
+    ddf = downscale_dataframe(ddf, "580 GiB")
```
ncclementi, line 164: Are you saying that the cluster needs to have that much memory to process the data? This number seems big based on my experience working with this data in the tutorial. In my case after using pyarrow[strings] the whole dataframe in memory was ~80GB

ncclementi (commented):
I left couple of comments/questions. 

One thing to mention is that we are including new packages, or updating some version in the recipe. Then the test will only run properly for the `latest` and `upstream` case, we should skip it for `0.2.1` .

I'm not sure about the pyarrow string fixture, I'll let @jrbourbeau  comment on that.

tests/benchmarks/test_xgboost.py
crusaderky: Reading the strings as string[python] and then casting them manually to string[pyarrow] would cost you a huge memory spike which, for the 830MiB chunks, you can't afford without killing off the workers.
Adding a comment.

tests/benchmarks/test_xgboost.py
crusaderky: airport_fee is float64

tests/benchmarks/test_xgboost.py
crusaderky: On a to_parquet / read_parquet round-trip, you lose the domains knowledge in the metadata, which the original notebook states xgboost needs to know (I did not investigate the veridicity of this statement).

Calling categorize() again restores the domain info in the metadata, albeit in an unnecessarily expensive way (you could just read them from the first partition).

tests/benchmarks/test_xgboost.py
crusaderky: Yes, correct.
The whole dataframe in memory is ~55GB.
However the xgboost train task will consume heap memory on each worker which is directly proportional to the total amount of data stored on that worker - with a gargantuan multiplier, I must add. xgboost.dask is very, very poorly designed - see https://github.com/coiled/dask-engineering/issues/167 for discussion.

Feel free to verify yourself - you will find this exact code on https://github.com/coiled/dask-xgboost-nyctaxi/pull/9, "Modelling 2" notebook. You can tweak the cluster size and watch it start first randomly, then deterministically kill off the workers with OOM as you reduce it to less than 580gb RAM.

ncclementi: Gotcha, this makes sense, thanks for the explanation.

crusaderky:
Added pin for sqlalchemy<2 (XREF https://github.com/dask/dask/issues/9896)

tests/benchmarks/test_xgboost.py
ncclementi: > the xgboost.dask examples suggests an unnecessary and expensive intermediate step. 

Thanks for clarifying this, I was not aware that step was unnecessary. I assumed their docs were following something standard, but I guess they are not.

tests/benchmarks/test_xgboost.py
ncclementi: Oh, I assumed it was float64 since in the code used to write this data it was supposed to convert it. I'm looking here https://github.com/coiled/coiled-from-prefect/blob/0a19e2daed1afe3fc43501f1ad4c0be470f9c232/flows/coiled_flow.py#L69-L80

I guess something went wrong, in that case, it makes sense to convert it again. That being said I was not able to find in their docs that this is what happens.

ncclementi:
@crusaderky 
Since we are including new packages, or updating some versions in the recipe. Then the test will only run properly for the latest and upstream case, should we skip it for 0.2.1 .?

crusaderky:
> @crusaderky Since we are including new packages, or updating some versions in the recipe. Then the test will only run properly for the latest and upstream case, should we skip it for 0.2.1 .?

no because you could run an A/B test with conda recipe

```yaml
channels:
  - conda-forge
dependencies:
    - coiled-runtime=0.2.1
    - optuna
```

tests/benchmarks/test_xgboost.py
```diff
@@ -0,0 +1,237 @@
...
+    ddf = ddf.persist().repartition(partition_size="100MB")
+    # At this point we would normally finish with to_parquet()
+    wait(ddf, small_client, timeout=600)
+
+
+@run_up_to_nthreads("small_cluster", 200, reason="fixed size dataset")
+def test_optuna_hpo(small_client):
+    # Needs either coiled-runtime >0.2.1 or explicitly adding the package to the A/B
+    # test environment
+    optuna = pytest.importorskip("optuna")
```
crusaderky, line 153: @ncclementi This is skipping the test in 0.2.1

ncclementi (approved):
This LGTM and the Ci failures are unrelated,

PR #694, opened 2023-02-27
Expand coverage of join tests

fjetter:
Haven't put an incredible amount of thought into these but the current benchmarks are very biased towards float / native data types and wide dataframes. This introduces parametrization over both dimensions and adds a trivial set_index benchmark

TODO: Migrations to align with current measurements

### DIFF
--- a/tests/benchmarks/test_join.py
+++ b/tests/benchmarks/test_join.py
@@ -35,24 +35,30 @@ def dtypes(request, num_cols_freq):
     dtypes: dict[str, type] = {
         str(ix): float for ix in range(0, num_cols - num_str_cols)
     }
-    for ix in range(num_cols - num_str_cols, num_str_cols):
+    for ix in range(num_cols - num_str_cols, num_cols):
         dtypes[str(ix)] = str
     return dtypes
 
 
 @run_up_to_nthreads("small_cluster", 40, reason="Does not finish")
-def test_join_big(small_client, memory_multiplier, configure_shuffling, dtypes, num_cols_freq):
+def test_join_big(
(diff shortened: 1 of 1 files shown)


fjetter:
Interesting, my GH actions runners appear to be running OOM

fjetter:
ok, the _culprit_ is the relatively naive logic of `timeseries_of_size`. in my test, I'm reducing the number of columns from 100 to 10 but am keeping `partition_freq` the same which generates 10x more partitions. Particularly with the tasks based algorithm this generates an absurd amount of tasks. Reproducing locally generates about 1.6M tasks for the merge statement which indeed requires a bit of memory for graph generation and serialization... 
Particularly, the way the benchmarks are written, they are calling persist and are waiting for the futures to finish. Persist actually triggers a graph materialization, i.e. these 2.5GB are materialized on the client/GH actions runner and the thing is then shipped to the scheduler. Given that a runner has only [7GB of memory](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) and we're parallelizing, this a guaranteed failure

hendrikmakait:
FWIW, I've run into issues with `partition_freq` a few times myself, we may want to consider adding a keyword like `chunksize` that would take care of automagically creating properly sized chunks if set.

fjetter:
> FWIW, I've run into issues with partition_freq a few times myself, we may want to consider adding a keyword like chunksize that would take care of automagically creating properly sized chunks if set.

I agree. This function should rather accept a "size_partition_in_bytes" parameter or something like that. I opted for the easy way now and am rescaling the `partition_frequency`.

fjetter:
Nope, just scaling the partition_freq is not going to work. Somebody should have a more thorough look at this.

PR #695, opened 2023-02-27
Benchmark P2P rechunking

hendrikmakait:


hendrikmakait:
We probably want to move these tests to benchmarks instead of stability.

fjetter (approved):

PR #709, opened 2023-03-10
Add set_index test

fjetter:
Simplified version of https://github.com/coiled/coiled-runtime/pull/694 that only adds a set_index


hendrikmakait (approved):
LGTM, thanks @fjetter!

hendrikmakait:
Actually, one question: `set_index` is a shuffling operation, not a join, isn't it? I'd move it to `test_dataframe.py`.

fjetter:
> Actually, one question: set_index is a shuffling operation, not a join, isn't it? I'd move it to test_dataframe.py.

Yes, it's a shuffle. Just added it to this file since there is the memory multiplier. In the original attempt adding this I was also paramterizing over the dtypes together with the other tests.
FWIW I think we should rather put all shuffle tests in one test_shuffle instead of having join and dataframe.

PR #714, opened 2023-03-13
Add test variants with compressible data

### LINKED ISSUE #696: Add test variants with compressible data
Many tests start from `da.random.random`, which is uncompressible data and can exhibit a fundamentally differently performance profile upon network transfer and spill/unspill.

For example, today we completely ignore the real life impact of https://github.com/dask/distributed/issues/7589.

The only tests that have a toggle for compressible data are in `test_spill.py`; we should replicate their design in test_array and possibly others.

crusaderky:
Closes #696


crusaderky:
I ran an A/B tests on #7593 and I'm observing a very modest (5%), but consistent speedup in `test_filter_then_average`. The test uses data that is compressible at 37%. The other tests do not show any kind of change - including those running on data that is compressible at 99%. The reason is that the data compressible at 37% takes 140ms per chunk to compress, whereas identically sized data full of ones takes 14ms per chunk. 

I need to scrap the current algorithm and synthetically create something similar to the zarr dataset.

crusaderky:
Holy guacamole :scream: 

image

crusaderky:
This is ready for review and merge. 
Discussion on the findings on https://github.com/dask/distributed/issues/7655

milesgranger (approved):
LGTM. :+1:

PR #722, opened 2023-03-17
Some benchmarks with arrow strings

j-bennet:
Some simple benchmarks with Arrow strings.

### DIFF
--- a/tests/benchmarks/test_arrow.py
+++ b/tests/benchmarks/test_arrow.py
@@ -5,54 +5,48 @@ from ..utils_test import cluster_memory, timeseries_of_size, wait
 
 
 @pytest.fixture(params=[True, False], ids=["string[pyarrow]", "object"])
-def series_with_client(request, small_client):
+def series(request, small_client):
     memory = cluster_memory(small_client)
     df = timeseries_of_size(memory)
     series = df.name
     if request.param:
         series = series.astype(pd.StringDtype("pyarrow"))
     series = series.persist()
-    yield series, small_client
+    yield series
 
 
-def test_unique(series_with_client):
+def test_unique(series, small_client):
     """Find unique values"""
-    series, client = series_with_client
     result = series.unique()
-    wait(result, client, 10 * 60)
+    wait(result, small_client, 10 * 60)
 
 
-def test_contains(series_with_client):
+def test_contains(series, small_client):
     """String contains"""
-    series, client = series_with_client
     result = series.str.contains("a")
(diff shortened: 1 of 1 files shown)


ntabris:
Try again, that user quota was updated a while ago, it appears the error is from ~1 week ago.

j-bennet:
@ntabris Now I get a different error: `coiled.errors.ClusterCreationError: Cluster status is error (reason: Timed out while waiting for workers/scheduler) (cluster_id: 181566)`

ntabris:
Not sure what happened there. The scheduler started downloading senv and then we don't hear from it or get more logs. You might just try re-running.

```
coiled cluster logs --cluster 181566 --workers none --combined --since "2023-03-31 17:30:08"
```

```
(scheduler)     [2023-03-31 17:30:08,478][INFO    ] Fetching download url
  [2023-03-31 17:30:09,010][INFO    ] Downloading & extracting 46 chunk(s)
```

@shughes-uk or @dan-blanchard can you see if @j-bennet needs help next week.

j-bennet:
@ntabris This PR keeps having problems with clusters. Now there's this:

```
FAILED tests/benchmarks/test_work_stealing.py::test_work_stealing_on_scaling_up - coiled.errors.ClusterCreationError: Cluster status is error (reason: Scheduler Stopped -> Environment failed to download, check cluster logs) (cluster_id: 184069)
```

ntabris:
Samantha and Dan are the people for package sync issues anyway

tests/benchmarks/test_arrow.py
```diff
@@ -0,0 +1,58 @@
+import pandas as pd
+import pytest
+
+from ..utils_test import cluster_memory, timeseries_of_size, wait
+
+
+@pytest.fixture(params=[True, False], ids=["string[pyarrow]", "object"])
+def series_with_client(request, small_client):
```
hendrikmakait, line 8: Instead of returning both the series and the client from this fixture, you can also re-use the client fixture here _and_ in the test. We do this in a few places in the benchmarks. Is there a specific reason you did not go for this pattern here?

shughes-uk:
Let me know if there's anymore package sync issues. Thing should be resolved.

j-bennet:
@shughes-uk The CI is green, thanks!

tests/benchmarks/test_arrow.py
j-bennet: No good reason. I updated the fixture.

PR #724, opened 2023-03-21
Add matplotlib arxiv workflow

jrbourbeau:
This PR adds a new representative workflow for an embarrassing parallel computation. Based on https://www.coiled.io/blog/how-popular-is-matplotlib.


cluster_kwargs.yaml
```diff
@@ -31,6 +31,12 @@ parquet_cluster:
   n_workers: 15
   worker_vm_types: [m5.xlarge]  # 4 CPU, 16 GiB
 
+# For tests/workflows/test_embarrassingly_parallel.py
+embarrassingly_parallel_cluster:
+  n_workers: 100
+  backend_options:
```
ncclementi, line 37: Is there a reason we are not specifying the VM types here? Using the default t3 will cause a lot of volatility in the times and trigger regressions. I'd recommend using `m6i.xlarge` and that will take ~4 min total

tests/conftest.py
```diff
@@ -530,6 +530,7 @@ def s3():
     return s3fs.S3FileSystem(
         key=os.environ.get("AWS_ACCESS_KEY_ID"),
         secret=os.environ.get("AWS_SECRET_ACCESS_KEY"),
+        requester_pays=True,
```
ncclementi, line 533: How does this affect the tests that do not require this?

cluster_kwargs.yaml
jrbourbeau: I want this to be representative of what a naive user might do. My current thinking is that they'll just use the default worker specification. Though maybe you or someone else have a different perspective. Definitely open to thoughts from others

tests/conftest.py
jrbourbeau: My guess is we'll pay for more transfer fees (https://s3fs.readthedocs.io/en/latest/#requester-pays-buckets) -- though this is probably already happening for datasets in the Coiled AWS account. Some public datasets require us to specify `requester_pays=True`. We didn't have a mechanism for passing `s3fs` options through, so I just hard coded it. cc @ntabris for thoughts on if this is problematic

ntabris: not sure if it's problematic. this feels like something that would make sense to have configurable per-test though, right? (that feels to me like the easier way of making sure it isn't more problematic, but of course, it would be someone else doing that work so of course it feels easier to me.)

ncclementi: I agree with Nat, the only test that needs `requester_pays` is the one you are adding here.

cluster_kwargs.yaml
ncclementi: It makes sense, although this will likely cause regression detections. I was able to see a difference of double the time between a run that Mat did and one I did with the defaults. 
Maybe @ntabris has some opinions on this?

ncclementi:
> Hmm we're getting `PermissionError: Access Denied` errors when running `test_embarassingly_parallel` here. I can run this example successfully on my laptop when using the `oss` Coiled AWS account. @ncclementi any ideas on what I might be doing wrong here?

Can you point me to the error? I can't find it in the previous CI runs. The only thing that comes to my head is how the `request_pays=True` is passed through to the cluster, maybe something wrong with that?

The other thing you can try locally is using the credentials of the bot and see if you can replicate it lcoally, they are on onepassword in the dask-eng vault.

jrbourbeau:
https://github.com/coiled/coiled-runtime/actions/runs/4475231686/jobs/7864609830 has the permissions error 

> he only thing that comes to my head is how the request_pays=True is passed through to the cluster, maybe something wrong with that?

I agree that seems like the place where things should be going wrong, but given I can run this test successfully locally, it makes me think it's more a credentials issue and `request_pays=` is being passed off appropriately 

> The other thing you can try locally is using the credentials of the bot and see if you can replicate it lcoally, they are on onepassword in the dask-eng vault.

Good idea. I need to do a few things to access onepassword on my machine. You mind trying out using the bot credentials? We could screenshare for 5 minutes to see if we can reproduce

tests/conftest.py
jrbourbeau: I've added a new `s3_factory` fixture for creating non-default `S3FileSystem` instances. Note it's function scoped instead of session scoped like the existing `s3` fixture (changing `s3` to be function scoped breaks other fixtures where `s3` is currently used)

tests/conftest.py
```diff
@@ -526,11 +526,17 @@ def s3_storage_options():
 
 
 @pytest.fixture(scope="session")
-def s3():
-    return s3fs.S3FileSystem(
-        key=os.environ.get("AWS_ACCESS_KEY_ID"),
-        secret=os.environ.get("AWS_SECRET_ACCESS_KEY"),
-    )
+def s3(s3_storage_options):
+    return s3fs.S3FileSystem(**s3_storage_options)
```
jrbourbeau, line 530: Note this change is just cosmetic. We define `key` / `secret` above in `s3_storage_options` and then do the same thing here again. This change just reuses `s3_storage_options` instead of grabbing the environment variables again. Should be okay as both fixtures are session scoped

jrbourbeau:
Alright, it looks like has fixed the permissions issue -- restarting those failing CI builds...

jrbourbeau:
cc @douglasdavis since we were talking about this example offline

cluster_kwargs.yaml
```diff
@@ -31,6 +31,12 @@ parquet_cluster:
   n_workers: 15
   worker_vm_types: [m5.xlarge]  # 4 CPU, 16 GiB
 
+# For tests/workflows/test_embarrassingly_parallel.py
+embarrassingly_parallel_cluster:
+  n_workers: 100
```
fjetter, line 36: I am slightly concerned about cost and CI runtime. 
Looking at the cluster, this uses up about 50-100 credits and I believe this one job runs for 10-20min.
I don't think this should run on every commit.

tests/workflows/test_embarrassingly_parallel.py
```diff
@@ -0,0 +1,89 @@
...
+        client.wait_for_workers(n_workers)
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+def test_embarassingly_parallel(embarrassingly_parallel_client, s3_factory):
+    # How popular is matplotlib?
+    s3 = s3_factory(requester_pays=True)
+    directories = s3.ls("s3://arxiv/pdf")
```
fjetter, line 45: IIUC we'll just let this run on everything that's there. I don't think this is necessary and considering that every month a new directory pops up this would also increase our runtime every month which is bad for a benchmark

cluster_kwargs.yaml
mrocklin: We should probably separate the real-world benchmarks from the artificial ones, and then only run the real-world ones (probably they're larger in general) on-demand ?

fjetter: it should be easy to do, this is all pytest stuff so we can use markers, paths, etc.

I'm fine with running on-demand and maybe once per week or release

ncclementi: For the record, the cost is about $1-3 bucks per run when using the t3 instances the times are in the order of 5-10 min. I suggested using m6i instances which get a more consistent time usage ~4min every time, but James mentioned he wanted to use the default t3s. For some info on cost and times, you can see this comment from Sarah who run this workflow multiple times for the arm vs intel comparison. 
https://github.com/coiled/platform/issues/645#issuecomment-1459019060

tests/workflows/test_embarrassingly_parallel.py
ncclementi: What if we make a copy of that into our own bucket, like a snapshot in time, and that way we also avoid the whole `request_payer` situation?

cluster_kwargs.yaml
jrbourbeau: I've updated things so that workflows (i.e. all tests in the `tests/workflows/` directory) are only run when requested and on the nightly cron job. Locally you can request workflows be run by adding the `--run-workflows` flag to your `pytest ...` command, and on PRs you can request workflows be run by adding the `workflows` label

tests/workflows/test_embarrassingly_parallel.py
jrbourbeau: 1991 corresponds to the beginning of the dataset. I think we want to analyze the full dataset (up to some time cutoff so we have a consistent data volume). The 2004/2005 lower bound you're referencing is just for visualization purposes. If you look at the full notebook https://github.com/mrocklin/arxiv-matplotlib/blob/main/arxiv-aws.ipynb you'll see an earlier plot that does back to 1991

ncclementi: What I meant to say is that the curve is pretty flat at the beginning and probably including less data for those years would save us some money, if we are running this once a day that is a cost of ~$2 * 365 = ~$730 a year for this workflow only.

jrbourbeau: Ah, I see what you mean. In general, I agree with this sentiment. Though in this particular case, there's not much data before 2004. Comparing back to back runs on 1991-2022 data and 2004-2022 data, I saw a rough price difference of ~$1.66 and ~$1.74, respectively.

Thinking about it more, extending down into the 90s actually has some value as it let's us confirm that the `filename_to_date` utility is working as expected. I've added a few light validation `assert`s to the end of this test. 

We can always revisit the subset of data we use here and frequency we run the workflows in the future if we want to do some price optimization

jrbourbeau: Just checking in @ncclementi does ^ seem reasonable to you?

ncclementi: This makes sense, thanks @jrbourbeau 
That was my last comment, I think this is good to go!

ncclementi (approved):
LGTM! Thank you @jrbourbeau

PR #727, opened 2023-03-22
Run benchmarks with t3 instead of m6i

ntabris:
I'd like a new run with t3 instances so I can screenshot a comparison of the costs in Cost Explorer, etc.

### DIFF
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -14,7 +14,7 @@
 default:
   package_sync: true
   wait_for_workers: true
-  scheduler_vm_types: [m6i.large]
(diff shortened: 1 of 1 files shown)


fjetter:
If you want an A/B test you'll need to modify a couple of other files. Otherwise this is done and we can close this

ntabris:
I didn't want A/B but I was hoping for clean run and I keep getting test failures.

ncclementi:
@ntabris Just to make sure, this is a DNM branch right? 
Would you mind adding that to the title? something like [DNM] just to avoid possible confusion

PR #728, opened 2023-03-23
Switch Coiled account to one specifically for benchmarks

ntabris:
Using a designated Coiled account for CI benchmarking will make permission boundary easier so we can make more of the benchmarking metrics public.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -85,7 +85,7 @@ def get_coiled_runtime_version():
             return "unknown"
 
 
-dask.config.set({"coiled.account": "dask-engineering"})
+dask.config.set({"coiled.account": "dask-benchmarks"})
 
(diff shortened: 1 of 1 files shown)


ntabris:
CI seems happy and I see lots of clusters in https://cloud.coiled.io/clusters?account=dask-benchmarks, with metrics in the new Prometheus data source, so I think this is good to merge.

The clusters are running in the same AWS account (w/ same AWS creds) as before, and CI is still using the bot user which I added to this Coiled account.

@jrbourbeau @fjetter I've made you both admins on the `dask-benchmarks` Coiled account so you can add/invite other folks to that team (and it should show up in the Coiled app dropdown menu).

dan-blanchard (approved):

PR #733, opened 2023-03-24
For new test runs use new Grafana URL

ntabris:
Haven't tested this locally.

1679590932198 is earliest timestamp (in ms, like what Grafana uses) of metrics that went to new datasource.

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -9,7 +9,7 @@ import operator
 import pathlib
 from collections.abc import Callable
 from textwrap import dedent
-from typing import Any, Literal, NamedTuple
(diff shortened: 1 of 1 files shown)


ntabris:
@jrbourbeau test failures shouldn't be relevant. I looked at output html, links to old clusters went to old Grafana and links to new clusters went to new Grafana, so this should be good to merge as far as I can tell.

jrbourbeau (commented):
Thanks @ntabris. Agreed the CI failures look unrelated. Though it looks like we should increase the `dask-benchmarks` CPU limit. Will ping you offline about that

jrbourbeau (approved):

PR #737, opened 2023-03-28
Add workflow reading csv, cleaning, saving to parquet

douglasdavis:
This workflow reads the GDELT project csv (actually `sep="\t"`) dataset which is on S3. The plan is to do some cleaning and save the transformed data to parquet.


douglasdavis:
Closing to base the PR off of the branch in #724

PR #739, opened 2023-03-28
Don't test against 0.2.1 runtime release

jrbourbeau:
This is an incremental change related to https://github.com/coiled/coiled-runtime/issues/732

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -252,7 +252,7 @@ jobs:
           environment-file: ci/environment-dashboard.yml
 
       - name: Generate dashboards
-        # Show differences of upstream vs. latest unpublished
+        # Show differences of upstream vs. latest
         run: python dashboard.py -d benchmark.db -o static -b coiled-latest-py3.9
 
       - name: Upload artifact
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -260,8 +252,8 @@ jobs:
           environment-file: ci/environment-dashboard.yml
 
       - name: Generate dashboards
-        # Show differences of upstream vs. latest unpublished vs. latest publshed
-        run: python dashboard.py -d benchmark.db -o static -b coiled-latest-py3.9 coiled-0.2.1-py3.9
+        # Show differences of upstream vs. latest unpublished
+        run: python dashboard.py -d benchmark.db -o static -b coiled-latest-py3.9
```
jrbourbeau, line 256: I think this change is okay from skimming `dashboard.py`, but @crusaderky could you confirm?

.github/workflows/tests.yml
```diff
@@ -260,8 +252,8 @@ jobs:
           environment-file: ci/environment-dashboard.yml
 
       - name: Generate dashboards
-        # Show differences of upstream vs. latest unpublished vs. latest publshed
-        run: python dashboard.py -d benchmark.db -o static -b coiled-latest-py3.9 coiled-0.2.1-py3.9
+        # Show differences of upstream vs. latest unpublished
```
crusaderky, line 255:
```suggestion
        # Show differences of upstream vs. latest
```

PR #746, opened 2023-03-30
Don't use coiled-runtime in A/B tests + general version bump

crusaderky:


jrbourbeau (commented):
Thanks @crusaderky. I'm not familiar with the ins-and-outs of our A/B testing system, but in general I'm +1 for a general version bump

crusaderky:
I just realised that this doesn't, in fact, close #757. I'm bringing it back to draft status until we figure out what's going on.

AB_environments/AB_baseline.conda.yaml
```diff
@@ -5,7 +5,43 @@
...
 dependencies:
+    # Copy-paste from recipe/meta.yaml, minus dask packages (see at bottom)
     - python=3.9
-    # Here we're using coiled-runtime for the sake of convenience, but we didn't have
-    # to. We could have also pointed to a nightly build.
-    - coiled-runtime=0.2.1
+    - pip
+    - coiled >=0.2.54
+    - numpy ==1.23.5
+    - pandas ==1.5.3
```
ncclementi, line 13: pandas 2.0 is out on pip, probably it'll be out on conda-forge soon, should we bump this too?

AB_environments/AB_baseline.conda.yaml
```diff
@@ -5,7 +5,43 @@
...
+    - toolz ==0.12.0
+    - zict ==2.2.0
+    - xgboost ==1.7.4
+    - dask-ml ==2023.3.24
+    - openssl >1.1.0g
+    - optuna ==3.1.0
+    - scipy ==1.10.1
+    - sqlalchemy ==1.4.46
+    - pynvml ==11.5.0
+    - bokeh ==2.4.3
```
ncclementi, line 40: @jrbourbeau is bokeh 3 working now? I thought I saw a PR about it but can't find it at the moment.

jrbourbeau: Yeah, `bokeh=3.1` is allowed in `distributed` `main` now -- though that change hasn't been released yet (xref https://github.com/dask/distributed/pull/7413)

AB_environments/AB_baseline.conda.yaml
crusaderky: Shouldn't we wait for full integration in dask and distributed CI?

jrbourbeau: I'm in favor of bumping to 2.0 now that it's officially out (assuming that it doesn't break CI here). The `dask` test suite is _generally_ passing with `pandas=2.0`, there are a small number of `xfail`s + follow-up issues, but largely things work with `dask.dataframe`. There's one small fix in flight for the `distributed` test suite (it's a tests-only fix).

Having `pandas=2.0` will also make it possible to use the new `dataframe.convert-string` config option in other tests here.

crusaderky:
Ah.

> The following packages are incompatible
> ├─ pandas 2.0.0  is requested and can be installed;
> └─ xarray 2023.1.0  is uninstallable because it requires
>    └─ pandas >=1.3,<2a0 , which conflicts with any installable versions previously reported.

For the time being, this means we either drop xarray completely (a few tests will need to be skipped) or we don't upgrade pandas. I'm midly in favour of the latter.
After we drop the metapackage, it will mean that the recipe for python 3.8 won't feature xarray.

crusaderky:
This one's ready for review and merge

AB_environments/AB_baseline.conda.yaml
crusaderky: See below. Can we leave the temporary removal of xarray - and bump to pandas 2 - to a separate PR?

jrbourbeau: Sounds good 👍

jrbourbeau (approved):
Thanks @crusaderky 

Just noting that the two CI build failures are unrelated and have been fixed elsewhere already

PR #752, opened 2023-03-30
Add rechunking workflow

jrbourbeau:
xref https://github.com/coiled/coiled-runtime/issues/725

### DIFF
--- a/tests/workflows/test_rechunk.py
+++ b/tests/workflows/test_rechunk.py
@@ -1,0 +1,43 @@
+import uuid
+
+import coiled
+import dask.array as da
+import pytest
+from dask.distributed import Client
+
+
+@pytest.fixture(scope="module")
+def rechunk_cluster(
+    dask_env_variables,
+    cluster_kwargs,
+    github_cluster_tags,
+):
+    with coiled.Cluster(
+        f"test-rechunk-{uuid.uuid4().hex[:8]}",
+        environ=dask_env_variables,
+        tags=github_cluster_tags,
+        **cluster_kwargs["embarrassingly_parallel_cluster"],
+    ) as cluster:
+        yield cluster
+
+
+@pytest.fixture
+def rechunk_client(
+    rechunk_cluster,
+    cluster_kwargs,
+    upload_cluster_dump,
+    benchmark_all,
+):
+    n_workers = cluster_kwargs["embarrassingly_parallel_cluster"]["n_workers"]
+    with Client(rechunk_cluster) as client:
+        rechunk_cluster.scale(n_workers)
+        client.wait_for_workers(n_workers)
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+def test_rechunk(rechunk_client, s3_url):
+    x = da.random.random((5_000, 5_000, 5_000, 2), chunks="50 MB")  # 1.82 TiB
+    y = x.rechunk("200 MB")
+    y.to_zarr(s3_url)


tests/workflows/test_rechunk.py
```diff
@@ -0,0 +1,43 @@
...
+    with Client(rechunk_cluster) as client:
+        rechunk_cluster.scale(n_workers)
+        client.wait_for_workers(n_workers)
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+def test_rechunk(rechunk_client, s3_url):
+    x = da.random.random((5_000, 5_000, 5_000, 2), chunks="50 MB")  # 1.82 TiB
```
jrbourbeau, line 41: TODO: Replace this with some public Zarr dataset on AWS

tests/workflows/test_rechunk.py
```diff
@@ -0,0 +1,43 @@
...
+        client.wait_for_workers(n_workers)
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+def test_rechunk(rechunk_client, s3_url):
+    x = da.from_zarr("s3://mur-sst/zarr", component="sea_ice_fraction")  # 3.80 TiB
+    y = x.rechunk("200 MB")
+    y.to_zarr(s3_url)
```
mrocklin, line 43: @rabernat we're trying to establish a set of benchmarks for common workloads faced by Dask users.  Ideally these are more realistic and less of toy examples.  

We were inspired by your thoughts on reading in a dataset stored in one chunking and then performing an analysis in another.  This is a quick proxy to that.  Do you happen to have anything that we could use that would be better?  In terms of complexity we'd love something in between 20-200 lines of code.

jrbourbeau:
Interestingly I'm not able to reproduce the `zarr.errors.ReadOnlyError: object is read-only
` failures that we're seeing in CI locally (even when using the CI bot's credentials). Here's an example of me running the rechunk test on my laptop https://cloud.coiled.io/clusters/180635/overview?account=dask-benchmarks. Will need to look into why this is happening

rabernat:
> Interestingly I'm not able to reproduce the `zarr.errors.ReadOnlyError: object is read-only ` failures

These have been cropping up everywhere. 🤦  See https://github.com/zarr-developers/zarr-python/issues/1353. Should be fixed in Zarr 2.14.2.

jrbourbeau:
Ah, great. Thanks @rabernat!

tests/workflows/test_rechunk.py
rabernat: The most difficult / pathological case to test is a "full shuffle" rechunk, where every input chunk goes into every output chunk.

I'd look at [this dataset](https://pangeo-forge.org/dashboard/feedstock/81). 

An array is at `s3://yuvipanda-test1/cmr/gpm3imergdl.zarr/HQprecipitation` ... not quite as big as MUR but chunked contiguously in space.

Screen Shot 2023-03-30 at 8 34 37 PM

Try something like

```python
ar = da.from_zarr("s3://yuvipanda-test1/cmr/gpm3imergdl.zarr/HQprecipitation")
ar.rechunk((-1, 36, 180))
```

mrocklin: Thanks @rabernat .  We're also trying to capture a broader set of common operations to keep things real-world-ish.  These might eventually graduate into some sort of notebook repository with worked examples that people can take a look at.  Is there some icing we could put on this particular cake?  Maybe some common analysis and an image that would be produced for example?

mrocklin: Or, maybe stated differently, @rabernat , what are 1-3 workflows you'd like run regularly as part of Dask's benchmark suite.  These get used day-to-day by engineers as they run AB tests on various changes that they want to make.  Workloads in this suite will naturally be optimized over time.

rabernat: I appreciate the chance to contribute to this. I'll try to follow up with more detail in the coming days.

jrbourbeau: Thanks for the pointer @rabernat 🚀 Btw, what region is the dataset located in? I didn't see a region reference on the pangeo-force page or the corresponding feedstock 

> Try something like

That example seems to run well (albeit with lots of transfers) when I try it here. I'm currently using a large cluster (~750GiB of memory) which is much bigger than the dataset itself (~200 GiB). Under what conditions do you tend to see bad performance? Maybe on a cluster with less memory than the dataset itself?

jrbourbeau: Just wanted to check in here -- @rabernat do you have additional details that'd be useful here (e.g. some common follow-up operations/plots like Matt mentioned)?

jrbourbeau: Just another gentle ping @rabernat

hendrikmakait:
@jrbourbeau: Is there anything you need help with to get this over the finish line?

tests/workflows/test_rechunk.py
rabernat: Sorry for my slow response here. I would recommend you look at https://github.com/pangeo-data/distributed-array-examples/issues. This is our best attempt to collect the sort of workflows you're looking for from the Pangeo Community.

PR #764, opened 2023-04-03
Exploratory DataFrame analysis

j-bennet:
Exploratory analysis on the NYC Uber/Lyft dataset.

Roughly based on @ncclementi 's dataframe tutorial: https://github.com/coiled/dask-tutorial/blob/main/2-Get_better-at-dask-dataframes.ipynb.

* Xref https://github.com/coiled/coiled-runtime/issues/725.

### DIFF
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -42,7 +42,7 @@ embarrassingly_parallel_cluster:
 
 # For tests/workflows/test_uber_lyft.py
 uber_lyft_cluster:
-  n_workers: 10
+  n_workers: 20
   worker_vm_types: [m6i.xlarge]  # 4 CPU, 16 GiB
 
 # For test_spill.py
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -42,9 +42,8 @@ embarrassingly_parallel_cluster:
 
 # For tests/workflows/test_uber_lyft.py
 uber_lyft_cluster:
-  n_workers: 30
-  backend_options:
-    region: "us-east-1"  # Same region as dataset
+  n_workers: 10
+  worker_vm_types: [m6i.xlarge]  # 4 CPU, 16 GiB
 
 # For test_spill.py
 spill_cluster:
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -40,6 +40,10 @@ embarrassingly_parallel_cluster:
   backend_options:
     region: "us-east-1"  # Same region as dataset
 
+# For tests/workflows/test_uber_lyft.py
+uber_lyft_cluster:
+  n_workers: 30
+
 # For test_spill.py
 spill_cluster:
   n_workers: 5
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -43,6 +43,8 @@ embarrassingly_parallel_cluster:
 # For tests/workflows/test_uber_lyft.py
 uber_lyft_cluster:
   n_workers: 30
+  backend_options:
+    region: "us-east-1"  # Same region as dataset
 
 # For test_spill.py
 spill_cluster:
--- a/tests/workflows/test_uber_lyft.py
+++ b/tests/workflows/test_uber_lyft.py
@@ -38,13 +38,13 @@ def uber_lyft_client(
 
 
 @pytest.fixture
-def ddf(uber_lyft_client):
+def ddf(uber_lyft_client, read_parquet_with_pyarrow):
     """NYC taxi Uber/Lyft dataset"""
-    paths = [
-        f"s3://coiled-datasets/mrocklin/nyc-taxi-fhv/{year}-*.parquet"
-        for year in range(2019, 2023)
-    ]
-    return dd.read_parquet(paths, engine="pyarrow", storage_options={"anon": True})
+    return dd.read_parquet(
+        "s3://coiled-datasets/uber-lyft-tlc/",
+        engine="pyarrow",
+        storage_options={"anon": True},
+    )
 
 
 def test_mean_tips(ddf):
--- a/tests/workflows/test_uber_lyft.py
+++ b/tests/workflows/test_uber_lyft.py
@@ -1,0 +1,67 @@
+import uuid
+
+import coiled
+import dask.dataframe as dd
+import pytest
+from dask.distributed import Client
+
+
+@pytest.fixture(scope="module")
+def uber_lyft_cluster(
+    dask_env_variables,
+    cluster_kwargs,
+    github_cluster_tags,
+):
+    with coiled.Cluster(
+        f"uber-lyft-{uuid.uuid4().hex[:8]}",
+        environ=dask_env_variables,
+        tags=github_cluster_tags,
+        **cluster_kwargs["uber_lyft_cluster"],
+    ) as cluster:
+        yield cluster
+
+
+@pytest.fixture
+def uber_lyft_client(
+    uber_lyft_cluster,
+    cluster_kwargs,
+    upload_cluster_dump,
+    benchmark_all,
+):
+    n_workers = cluster_kwargs["uber_lyft_cluster"]["n_workers"]
+    with Client(uber_lyft_cluster) as client:
+        uber_lyft_cluster.scale(n_workers)
+        client.wait_for_workers(n_workers)
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+@pytest.fixture
+def ddf(uber_lyft_client):
+    """NYC taxi Uber/Lyft dataset"""
+    paths = [
+        f"s3://coiled-datasets/mrocklin/nyc-taxi-fhv/{year}-*.parquet"
+        for year in range(2019, 2023)
+    ]
+    return dd.read_parquet(paths, engine="pyarrow")
+
+
+def test_mean_tips(ddf):
+    """How many passengers tip"""
+    (ddf.tips != 0).mean().compute()
+
+
+def test_max_pay(ddf):
+    """Max driver pay"""
+    ddf.driver_pay.max().compute()
+
+
+def test_max_distance(ddf):
+    """Ride count per rideshare company"""
+    ddf.groupby("hvfhs_license_num").count().compute()
+
+
+def test_longer_trips(ddf):
+    """How many passengers ride over 5 miles"""
+    (ddf.trip_miles > 5).mean().compute()
--- a/tests/workflows/test_uber_lyft.py
+++ b/tests/workflows/test_uber_lyft.py
@@ -37,91 +37,32 @@ def uber_lyft_client(
             yield client
 
 
-def test_explore_dataset(uber_lyft_client):
+def test_exploratory_analysis(uber_lyft_client):
     """Run some exploratory aggs on the dataset"""
 
     # NYC taxi Uber/Lyft dataset
-    columns = [
-        "hvfhs_license_num",
-        "tips",
-        "base_passenger_fare",
-        "driver_pay",
-        "trip_miles",
-        "trip_time",
-        "shared_request_flag",
-    ]
-    ddf = dd.read_parquet("s3://coiled-datasets/uber-lyft-tlc/", columns=columns)
+    df = dd.read_parquet(
+        "s3://coiled-datasets/uber-lyft-tlc/", storage_options={"anon": True}
+    )
 
-    # convert object column to arrow string
-    ddf["hvfhs_license_num"] = ddf["hvfhs_license_num"].astype("string[pyarrow]")
-
+    # Preprocessing:
+    #   - Add a column to indicate company, instead of license number
+    #   - Add a column to indicate if a tip was given
     taxi_companies = {
         "HV0002": "Juno",
         "HV0003": "Uber",
         "HV0004": "Via",
         "HV0005": "Lyft",
     }
+    df["company"] = df.hvfhs_license_num.replace(taxi_companies)
+    df["tipped"] = df.tips > 0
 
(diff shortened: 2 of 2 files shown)


tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,67 @@
...
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+@pytest.fixture
+def ddf(uber_lyft_client):
+    """NYC taxi Uber/Lyft dataset"""
+    paths = [
+        f"s3://coiled-datasets/mrocklin/nyc-taxi-fhv/{year}-*.parquet"
```
ncclementi, line 44: Couple of questions here
- I'm confused about the `year` arg. each of the `.parquet` in `"s3://coiled-datasets/mrocklin/nyc-taxi-fhv/` is a day-partitioned file

Is there a reason why we are creating a list of paths? you can read the whole dataset as  
```python
dd.read_parquet("s3://coiled-datasets/mrocklin/nyc-taxi-fhv/, engine="pyarrow")
```

- Is there a specific reason on why using this dataset? 
 
The partition sizes here are not of great size, they are quite small because it's partitioned  by day, and are kind of imbalanced a bit. 

`s3://coiled-datasets/uber-lyft-tlc/` this one is better partitioned (partition size ~100MB)

ncclementi: Also, the dask-dataframe tutorial is based on this dataset. We are having a session on Wednesday in case you want to listen in and take a look and what this dataset looks like originally. The materials for reference are here https://github.com/coiled/dask-tutorial/blob/main/2-Get_better-at-dask-dataframes.ipynb, happy to chat about this dataset and workflow if needed.

cluster_kwargs.yaml
```diff
@@ -40,6 +40,12 @@ embarrassingly_parallel_cluster:
   backend_options:
     region: "us-east-1"  # Same region as dataset
 
+# For tests/workflows/test_uber_lyft.py
+uber_lyft_cluster:
+  n_workers: 30
+  backend_options:
+    region: "us-east-1"  # Same region as dataset
```
ncclementi, line 47:
```suggestion
    region: "us-east-2"  # Same region as dataset
```
The dataset being used at the moment, the `s3://coiled-datasets/mrocklin/` is in `us-east-2` as most of the datasets inside `coiled-datasets`

cluster_kwargs.yaml
```diff
@@ -40,6 +40,12 @@ embarrassingly_parallel_cluster:
   backend_options:
     region: "us-east-1"  # Same region as dataset
 
+# For tests/workflows/test_uber_lyft.py
+uber_lyft_cluster:
+  n_workers: 30
```
ncclementi, line 45: Is this one arbitrary? 
I have recommendations about the cluster size and instance type. In the tutorial we run this with 
```
    n_workers=15,
    worker_vm_types=["m6i.xlarge"],
``` 
but we are persisting the data, and casting objects to pyarrow strings, and other conversions to optimize memory.

j-bennet: I based this on @mrocklin 's blog post draft:
https://docs.google.com/document/d/15VepxvOZoadTqM82cedxsyReq_hKZmMVZiKeFS_pADk/edit#
he's using 30 workers for that. I'll try your settings.

cluster_kwargs.yaml
j-bennet: And `s3://coiled-datasets/uber-lyft-tlc/`?

tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,67 @@
...
+        for year in range(2019, 2023)
+    ]
+    return dd.read_parquet(paths, engine="pyarrow", storage_options={"anon": True})
```
ncclementi, line 47: We have a fixture to read parquet using string pyarrows, See https://github.com/coiled/coiled-runtime/blob/decbba56de7158b590b1e79fd7b6eb490838df09/tests/conftest.py#L674-L698

and this is how you can use it https://github.com/coiled/coiled-runtime/blob/8f159e9d0e1ea97874fd751fce3607c3f025ff0a/tests/benchmarks/test_xgboost.py#L33

tests/workflows/test_uber_lyft.py
j-bennet: Thanks for the tip @ncclementi, I can switch to `s3://coiled-datasets/uber-lyft-tlc/`.

I think smaller partition sizes correlate with smaller worker sizes in @mrocklin 's example. If I use `s3://coiled-datasets/uber-lyft-tlc/`, I will probably need `m6i.xlarge` like you said.

cluster_kwargs.yaml
ncclementi: Same, if you have doubts about where they are in the AWS ui, once you are in the bucket/folder you can go over to properties, and see where it is. 
Screen Shot 2023-04-03 at 7 16 11 PM

tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_max_distance(ddf):
+    """Ride count per rideshare company"""
+    ddf.groupby("hvfhs_license_num").count().compute()
+
+
+def test_longer_trips(ddf):
+    """How many passengers ride over 5 miles"""
+    (ddf.trip_miles > 5).mean().compute()
```
jrbourbeau, line 67: Let's move these all into a single test. We'll eventually move the workflow in this file into its own notebook example that we want to be representative of what a common Dask DataFrame user might do as part of an exploratory analysis. I think we want a single test here that captures what that would look like. My current thinking is something relatively straightforward along the lines of (clearly it's just pseudocode below):

```python
# Load data
ddf = dd.read_parquet(...)

# Compute interesting quantities to gain insight
tips_fraction = (ddf.tips != 0).mean()
x = ...
y = ...
results = dask.compute(tips_fraction, x, y, ...)
```

tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,67 @@
...
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+@pytest.fixture
+def ddf(uber_lyft_client, read_parquet_with_pyarrow):
+    """NYC taxi Uber/Lyft dataset"""
+    return dd.read_parquet(
+        "s3://coiled-datasets/uber-lyft-tlc/",
+        engine="pyarrow",
```
jrbourbeau, line 45: Should be able to drop this since `"pyarrow"` is the default engine (when `pyarrow` is installed) 

```suggestion
```

tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,67 @@
...
+        storage_options={"anon": True},
```
jrbourbeau, line 46: Is this needed to access the dataset? If so, no worries, let's keep it. If not, then let's remove it

jrbourbeau (commented):
Thanks @j-bennet

tests/workflows/test_uber_lyft.py
ncclementi: Are you asking about the `storage_options={"anon": True}` ? That shouldn't be needed. The dataset is public and things should work, I recall having some issues in the past and this fixed it. I think there was an issue with CI at the time.

tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,67 @@
...
+    with Client(uber_lyft_cluster) as client:
+        uber_lyft_cluster.scale(n_workers)
+        client.wait_for_workers(n_workers)
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+@pytest.fixture
+def ddf(uber_lyft_client, read_parquet_with_pyarrow):
```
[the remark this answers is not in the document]
ncclementi: No, but they will cast to `pyarrow-strings`, at least we are encouraging people to do so, this is an intermediate step until we can use the config easily without relying on pandas 2.0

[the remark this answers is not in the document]
jrbourbeau: Fair point. For now I'd encourage us to not add any large workarounds for issues that a common user would run into. We should do the straightforward thing in our workflows here and then work upstream in `dask`, `distributed`, etc. to make the representative code here work well.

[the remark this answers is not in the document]
ncclementi: in that case you'll need to crank up the number of workers because of memory pressure, you'll likely hit spilling.

j-bennet:
@dan-blanchard @shughes-uk can you help? I'm getting these:

```
ERROR tests/benchmarks/test_spill.py::test_spilling[uncompressible-release] - coiled.errors.ClusterCreationError: Cluster status is error (reason: Scheduler Stopped -> Environment failed to download: seeking backwards is not allowed) (cluster_id: 185977)
```

shughes-uk:
https://github.com/coiled/platform/issues/1027

feel free to retry individual jobs that fail

tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,97 @@
...
+
+    taxi_companies = {
+        "HV0002": "Juno",
+        "HV0003": "Uber",
+        "HV0004": "Via",
+        "HV0005": "Lyft",
+    }
+
+    # add a column to indicate company, instead of license number
+    ddf["company"] = ddf.hvfhs_license_num.replace(taxi_companies)
```
jrbourbeau, line 75: Let's construct all the columns in the DataFrame before `persist`ing

tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,97 @@
...
+    """NYC taxi Uber/Lyft dataset"""
+    columns = [
+        "hvfhs_license_num",
+        "tips",
+        "base_passenger_fare",
+        "driver_pay",
+        "trip_miles",
+        "trip_time",
+        "shared_request_flag",
+    ]
```
jrbourbeau, line 51: Why select just these columns? My sense is most folks will just read the whole dataset in and then select the columns they need afterwards. Over time, once we have a more generic column selection / predicate pushdown / etc., these types of things will automatically improve for users.

tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,97 @@
...
+    with Client(uber_lyft_cluster) as client:
+        uber_lyft_cluster.scale(n_workers)
+        client.wait_for_workers(n_workers)
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+@pytest.fixture
+def ddf(uber_lyft_client):
```
jrbourbeau, line 41: Let's just inline this in the test. IIUC I don't think this needs to be in a fixture

jrbourbeau (commented):
Thanks for the updates @j-bennet, this is shaping up nicely. 

@ncclementi @mrocklin do you have thoughts on what's here? You've both played around with this dataset a bit and have a good sense for what exploring DataFrame users are likely to do

tests/workflows/test_uber_lyft.py
j-bennet: I'm only calculating a few metrics here and not using all the columns, so it makes more sense to load a subset to decrease the size of data. We would recommend the same for users and their real-world workflows, this is also what @ncclementi does in her [tutorial](https://github.com/coiled/dask-tutorial/blob/main/2-Get_better-at-dask-dataframes.ipynb).

ncclementi:
@jrbourbeau I'd recommend any of the things we do in the tutorial. https://github.com/coiled/dask-tutorial/blob/main/2-Get_better-at-dask-dataframes.ipynb

ncclementi:
The pip failure should be fixed on main now, if you merge main on to this branch that should fix CI.

j-bennet:
> I'd recommend any of the things we do in the tutorial. https://github.com/coiled/dask-tutorial/blob/main/2-Get_better-at-dask-dataframes.ipynb

@ncclementi I added more aggs, take a look. It doesn't encapsulate the whole tutorial, but it's a bit more representative.

ncclementi (approved):
Outside my comment on the quantile calculation, and the outliers. This LGTM. 
I'd suggest two options;
1- try to compute quantiles using dask even though they are approximate
2- Leave it as is, but put a note on that.

tests/workflows/test_uber_lyft.py
```diff
@@ -0,0 +1,129 @@
...
+    # what's the largest and average pay and tips that driver got, by company
+    ddf.groupby("company").agg(
+        {"driver_pay": ["max", "mean"], "tips": ["max", "mean"]}
+    ).compute()
+
+    # tip percentage mean of trip with tip
+    ddf.loc[lambda x: x.tip_flag].groupby("company").tip_percentage.mean().compute()
+
+    # sample the dataset to filter outliers, result is a pandas dataframe
+    df_tiny = (
```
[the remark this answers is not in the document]
j-bennet: @ncclementi This is interesting. I'll try and see if dask produces the same result with quantiles as pandas on this dataset. I thought that calculating quantiles on a large dataset just takes too long.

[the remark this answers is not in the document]
j-bennet: I tried calculating quantile with `dask`, yikes. It works fine on `base_passenger_fare` and `trip_miles`, both default method and `tdigest`. However, on `tips`, both`dask` method and `tdigest` return `array([0., 0.])`.

It looks like for now, we have to keep the sampling approach.

jrbourbeau (approved):
Thanks @j-bennet! 

We spoke offline about simplifying this workflow a bit. I've gone ahead and pushed a commit that does this

PR #782, opened 2023-04-11
Rename repo to `coiled/benchmarks`

jrbourbeau:
Closes https://github.com/coiled/coiled-runtime/issues/750


jrbourbeau:
Note there are a few other "coiled runtime" occurrences that I'll handle in a follow-up PR. This PR is just to make sure tests pass after the `coiled-runtime` --> `benchmarks` name change

PR #785, opened 2023-04-12
Add `optuna` workflow

jrbourbeau:
This adds a new `xgboost` (non-dask) + `optuna.integrations.dask` workflow 

xref https://github.com/coiled/benchmarks/issues/725

### DIFF
--- a/tests/workflows/test_optuna.py
+++ b/tests/workflows/test_optuna.py
@@ -1,0 +1,82 @@
+import uuid
+
+import coiled
+import optuna
+import pytest
+import xgboost as xgb
+from dask.distributed import Client, wait
+from optuna.integration.dask import DaskStorage
+from optuna.samplers import RandomSampler
+from sklearn.datasets import fetch_covtype
+from sklearn.model_selection import KFold, cross_val_score
+from sklearn.preprocessing import LabelEncoder
+
+
+@pytest.fixture(scope="module")
+def optuna_cluster(
+    dask_env_variables,
+    cluster_kwargs,
+    github_cluster_tags,
+):
+    with coiled.Cluster(
+        f"optuna-{uuid.uuid4().hex[:8]}",
+        environ=dask_env_variables,
+        tags=github_cluster_tags,
+        **cluster_kwargs["optuna_cluster"],
+    ) as cluster:
+        yield cluster
+
+
+@pytest.fixture
+def optuna_client(
+    optuna_cluster,
+    cluster_kwargs,
+    upload_cluster_dump,
+    benchmark_all,
+):
+    n_workers = cluster_kwargs["optuna_cluster"]["n_workers"]
+    with Client(optuna_cluster) as client:
+        optuna_cluster.scale(n_workers)
+        client.wait_for_workers(n_workers)
+        client.restart()
+        with upload_cluster_dump(client), benchmark_all(client):
+            yield client
+
+
+def test_optuna(optuna_client):
+    # We use a random sampler with a seed to get deterministic results.
+    # This is just for benchmarking purposes.
+    study = optuna.create_study(
+        direction="maximize", storage=DaskStorage(), sampler=RandomSampler(seed=2)
+    )
+
+    def objective(trial):
+        # Dataset (241.59 MiB) from http://archive.ics.uci.edu/ml/datasets/covertype
+        X, y = fetch_covtype(return_X_y=True)
+
+        # Format training labels
+        le = LabelEncoder()
+        y = le.fit_transform(y)
+
+        # Get hyperparameter values for this trial and model score
+        params = {
+            "n_estimators": trial.suggest_int("n_estimators", 2, 10),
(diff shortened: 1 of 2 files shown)


tests/workflows/test_optuna.py
```diff
@@ -0,0 +1,82 @@
...
+def test_optuna(optuna_client):
+    # We use a random sampler with a seed to get deterministic results.
+    # This is just for benchmarking purposes.
+    study = optuna.create_study(
+        direction="maximize", storage=DaskStorage(), sampler=RandomSampler(seed=2)
+    )
+
+    def objective(trial):
+        # Dataset (241.59 MiB) from http://archive.ics.uci.edu/ml/datasets/covertype
+        X, y = fetch_covtype(return_X_y=True)
```
jrbourbeau, line 55: I'd like to eventually either find a hosted dataset on AWS or host this dataset ourselves. Though I'm also fine with using this `sklearn` utility for the time being. Seems good enough for a first pass

jrbourbeau:
Also wanted to note that I ran into https://github.com/optuna/optuna/issues/4604 while making this example

tests/workflows/test_optuna.py
ncclementi: for aws hosted: what are the requirements/characteristics of a good dataset for this? 
to host this one: what is stopping us from doing so?

jrbourbeau: > what are the requirements/characteristics of a good dataset for this?

The dataset needs to fit comfortably onto a single machine. That said, it'd also be good if the dataset wasn't trivial (the covertype dataset has 581k rows). I also like that the covertype dataset requires some, but not lots, of preprocessing. 

> what is stopping us from doing so?

Nothing that I'm aware of. Just having done it yet

tests/workflows/test_optuna.py
```diff
@@ -0,0 +1,93 @@
...
+        le = LabelEncoder()
+        y = le.fit_transform(y)
+
+        # Get hyperparameter values for this trial and model score
+        params = {
+            "n_estimators": trial.suggest_int("n_estimators", 2, 10),
+            "max_depth": trial.suggest_int("max_depth", 2, 10),
+            "learning_rate": trial.suggest_float("learning_rate", 1e-8, 1.0, log=True),
+            "subsample": trial.suggest_float("subsample", 0.2, 1.0),
+            "eval_metric": "mlogloss",
```
ncclementi, line 78: Is there a reason we are going off the default here? I'm not super familiarized with the problem, I just know the default for the Classifier is `logloss`. I was looking in the [docs](https://xgboost.readthedocs.io/en/stable/parameter.html) what `mlogloss` was but it links to something on sklearn so I got confused

tests/workflows/test_optuna.py
```diff
@@ -0,0 +1,93 @@
...
+            "n_jobs": 1,  # Avoid thread oversubscription
+        }
+        model = xgb.XGBClassifier(**params)
+        cv = KFold(n_splits=3, shuffle=True, random_state=2)
+        score = cross_val_score(model, X, y, cv=cv, scoring="accuracy")
```
ncclementi, line 83: Why do we choose accuracy here as the score? I'm not super familiarized with the model and the problem but in a similar (I think) [classification problem](https://docs.coiled.io/user_guide/examples/dask-optuna-hpo.html) we have used "neg_log_loss", does it matter here?

tests/workflows/test_optuna.py
```diff
@@ -0,0 +1,93 @@
...
+        # Get hyperparameter values for this trial and model score
+        params = {
+            "n_estimators": trial.suggest_int("n_estimators", 2, 10),
+            "max_depth": trial.suggest_int("max_depth", 2, 10),
+            "learning_rate": trial.suggest_float("learning_rate", 1e-8, 1.0, log=True),
+            "subsample": trial.suggest_float("subsample", 0.2, 1.0),
+            "eval_metric": "mlogloss",
+            "n_jobs": 1,  # Avoid thread oversubscription
+        }
+        model = xgb.XGBClassifier(**params)
```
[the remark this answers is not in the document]
jrbourbeau: Not really. We've written a bunch of small examples like this in the past. This is mostly just off of memory.

tests/workflows/test_optuna.py
jrbourbeau: I initially chose `mlogloss` because this isn't a binary classification problem, but yeah, I think you're right that the default `logloss` is probably more appropriate

tests/workflows/test_optuna.py
jrbourbeau: Updated to not specify `scoring=` at all. Instead `sklearn` will just use the default scoring metric for the `xgb.XGBClassifier` model (I think `logloss`?)

ncclementi: yupe, I checked and it is it is `logloss`

PR #787, opened 2023-04-13
PyTorch + Optuna GPU workflow

### LINKED ISSUE #759: PyTorch hyperparameter optimization
We have Optuna. We use it in https://github.com/coiled/dask-xgboost-nyctaxi for XGBoost but we should also show how to use it in more vanilla settings with a model that can be trained on a single machine, presumably a GPU. Let’s use PyTorch for this.

- [ ] Train some PyTorch GPU model that fits on a single GPU with Optuna for HPO on a cluster

xref https://github.com/coiled/coiled-runtime/issues/725

milesgranger:
Will close #759

### DIFF
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -56,7 +56,7 @@ uber_lyft:
 
 # For tests/workflows/test_pytorch_optuna.py
 pytorch_optuna:
-  n_workers: 4
+  n_workers: 10
   worker_gpu: 1
   worker_options:
     # Making workers single-threaded to avoid GPU contention. See discussion in


milesgranger:
While I cannot get this to run as a workflow currently, will sync with @ntabris later to figure out the `ModuleNotFound` error, here is an example of the notebook here running on Coiled. Task queue isn't so fascinating to look at but at least the GPUs are busy. 

Could adapt it to use software environments like @mrocklin did in his example to make it easier but interestingly enough for @ntabris maybe is from package_sync, the workers got the CPU only torch install and not GPU, hence the `PipInstall` plugin to force reinstalling torch on the workers then things were good. :+1: 

ezgif-4-e06adc537b

cc @mrocklin @jrbourbeau

mrocklin:
Would this be appropriate for a demo at a conference booth do you think @milesgranger ?  If so, any interest in also adding this to https://github.com/coiled/examples/blob/main/pytorch.ipynb ?

milesgranger:
Sure, I think I could dress it up a bit more to make the iterations over the model training faster to get more action in the task chart; but try and keep the GPUs busy. I suppose it ought to use software environments as well if used in examples/conference demo then. I'll adjust. :+1:

tests/conftest.py
```diff
@@ -88,7 +88,7 @@ def get_coiled_runtime_version():
             return "unknown"
 
 
-dask.config.set({"coiled.account": "dask-benchmarks"})
+dask.config.set({"coiled.account": "dask-engineering"})
```
ntabris, line 91: please don't change this

ntabris: or rather, could you please change this back?

milesgranger: Yes sorry, just caught that. Thanks! :)

milesgranger:
@jrbourbeau are you satisfied with the updates you made? I have no additional changes I'd like to see, so good for a last review if you are. :)

cluster_kwargs.yaml
```diff
@@ -54,6 +54,13 @@ uber_lyft:
   n_workers: 20
   worker_vm_types: [m6i.xlarge] # 4 CPU, 16 GiB (preferred default instance)
 
+# For tests/workflows/test_pytorch_optuna.py
+pytorch_optuna:
+  n_workers: 4
+  worker_gpu: 1
+  worker_options:
+    nthreads: 1
```
jrbourbeau, line 62: Why is `nthreads=1` needed here?

tests/workflows/test_pytorch_optuna.py
```diff
@@ -0,0 +1,208 @@
...
+from dask.distributed import PipInstall, wait
+
+NC = 3  # Number of channels in the training images. For color images this is 3
+NZ = 100  # Size of z latent vector (i.e. size of generator input)
+
+
+# We create a fake image dataset. This ought to be replaced by
+# your actual dataset or pytorch's example datasets. We do this here
+# to focus more on computation than actual convergence.
+class FakeImageDataset(torch.utils.data.Dataset):
```
jrbourbeau, line 19: Do you have any suggestions on a real dataset to use here?

tests/workflows/test_pytorch_optuna.py
```diff
@@ -0,0 +1,208 @@
...
+        # state size. ``(NC) x 64 x 64``
+    )
+    model.apply(weights_init)
+    return model
+
+
+def get_discriminator(trial):
+    # Create and randomly initialize discriminator model
+    activations = ["Sigmoid"]
+    activation = trial.suggest_categorical("discriminator_activation", activations)
```
jrbourbeau, line 77: We're always selecting the `nn.Sigmoid` activation here -- are there others we could add?

tests/workflows/test_pytorch_optuna.py
```diff
@@ -0,0 +1,208 @@
...
+            if trial.should_prune():
+                raise optuna.exceptions.TrialPruned()
+
+        return score
+
+    study = optuna.create_study(
+        storage=optuna.integration.DaskStorage(), direction="minimize"
+    )
+
+    n_trials = 100
```
jrbourbeau, line 201: This number of trials is much larger than the possible combinations of `trial.suggest_*` options in the objective function. Are there other hyperparameters that are commonly tuned?

jrbourbeau (commented):
Thanks @milesgranger

ci/environment.yml
```diff
@@ -1,11 +1,14 @@
 channels:
   - conda-forge
+  - pytorch
```
jrbourbeau, line 3: Oh, also do we need the `pytorch` channel here, or is `conda-forge` sufficient?

milesgranger: I believe so, [pytorch quickstart](https://pytorch.org/get-started/locally/) has a nifty tool which explicitly states the channels, and skimming anaconda it seems packages like pytorch-cuda is only on the pytorch channel.

cluster_kwargs.yaml
milesgranger: To prevent any concurrent access to the GPU per VM, since we're using single GPU instances here... could be possible to modify if the VMs had multiple devices I think.

tests/workflows/test_pytorch_optuna.py
milesgranger: Yesh... could. But I approached this workflow with the objective of making the most use of the GPUs as possible and not complicating it with downloading external data (which my first attempt revield being hit with rate limits, so could have to copy to our S3 I suppose)

In general though, the layout follows the standard PyTorch pattern of Dataset/DataLoader so anyone looking at it as reference would have a short way to go to get their dataset.

With that said, the image dataset isn't entirely random, in that it sorta moves the values of the image to the value of the label; but I yield to your judgement, if it's better to have real data I'm happy to adjust.

tests/workflows/test_pytorch_optuna.py
milesgranger: Added some more, and also to address the other comment about adding more hyperparameters in general.
https://github.com/coiled/benchmarks/pull/787/commits/55d8c1906d27c110734c637dd5a4594a789c79e5

tests/workflows/test_pytorch_optuna.py
milesgranger: https://github.com/coiled/benchmarks/pull/787/commits/55d8c1906d27c110734c637dd5a4594a789c79e5

ci/environment.yml
milesgranger: nvm, I retract that, as we're only installing pytorch which is on conda-forge
https://github.com/coiled/benchmarks/pull/787/commits/7c5b94e2132d2452814cd1ca8c0ab5e5dc122906

cluster_kwargs.yaml
ntabris: Do you want single worker per GPU? Or do you want single worker per host VM, and PyTorch will manage across all GPUs on the host? Or not sure?

tests/workflows/test_pytorch_optuna.py
mrocklin: I'll push back here.  While I appreciate having a demo during conferences that has high utilization (thanks for that, it was useful), I also appreciate having workflows that represent real workloads.  

If, for example, we find that it's very hard to saturate a GPU due to data loading then as an engineering team we should focus much more on data loading.  If we find that in real world use cases data loading is not important then we can ignore it comfortably.  

Having a realistic workload with realistic data loading helps us, I think, to answer questions like these.

These workflows serve multiple use cases.  For notebooks I think demo-ability is useful.  For the pytest benchmarks I think that fidelity to the real-world is probably more useful.

fjetter: I agree with Matt here. We want real data for this and if necessary we'll copy to our S3

cluster_kwargs.yaml
fjetter: Is this based on a recommendation from our docs?

* If so, all good
* If not and this is indeed how one has to set it up, we should document this

tests/workflows/test_pytorch_optuna.py
milesgranger: Let it be written then, I'll update. 👍

cluster_kwargs.yaml
milesgranger: It appears having more than 1 thread per worker creates some sort of contention for the GPU and the utilization drops to ~40%, otherwise w/ 1 thread per worker I get nearly 100%. 

I did not get this from any documentation, nor did I locate it in the docs now; so will be sure to add a note about this experience in a follow-up PR there.

tests/workflows/test_pytorch_optuna.py
milesgranger: Updated in https://github.com/coiled/benchmarks/pull/787/commits/ca8b0fd7805fe3e6df44289a11c2de3f7f569318

cluster_kwargs.yaml
milesgranger: Nvm, after going to make a PR to update docs, I saw it was (kinda) there after-all. Granted it doesn't explicitly say 1 thread per worker, but makes the overall point. https://docs.dask.org/en/latest/gpu.html?highlight=gpu#restricting-work

milesgranger:
@ntabris Seems Windows package-sync doesn't like torchvision: https://github.com/coiled/benchmarks/actions/runs/4871118668/jobs/8687866811?pr=787

I can move the imports inside of the task function and handle the install using the existing work-around for `PipInstall` for settling the GPU versions of packages.

cluster_kwargs.yaml
ntabris: I believe `dask-cuda` also does that for you. It might make more sense to try that rather than doing this manually.

milesgranger: Okay, so you propose these lines should be removed, add `dask-cuda` then use the settings described here: https://docs.rapids.ai/api/dask-cuda/nightly/api/#client-initialization

As it appears the other configuration for it is to replace `LocalCluster` with `LocalCUDACluster` which doesn't seem like what we want(?).

ntabris: Add `dask-cuda` and then set worker class when making cluster, as shown in https://docs.coiled.io/user_guide/gpu.html

milesgranger: Thanks. https://github.com/coiled/benchmarks/pull/787/commits/be11248966d5ec684562a3d5ae2f152a71b72c88

mrocklin: I had tried this earlier, and ran into conda solve issues.  Bringing in the RAPIDS channel was unpleasant.  If it works now then grand.  I also think that this is a "I wanted a banana, but I got a gorilla holding a banana" situation.

milesgranger: Ah, indeed. [Seems that's true here then. ](https://github.com/coiled/benchmarks/actions/runs/4872308102?pr=787)

jrbourbeau: Thanks for the context all. Sounds like we should just stick with `nthreads=1` for now. I added a small comment here that mentions why we're making things single-threaded (linking back to this thread on GitHub).

jrbourbeau (commented):
Thanks @milesgranger! I went ahead and added the `workflows` label to make sure this test is run in CI. This is looking really close. Anything outstanding that you'd like to address here? Otherwise, I'd suggest we call this good and follow-up as needed

> Seems Windows package-sync doesn't like torchvision: https://github.com/coiled/benchmarks/actions/runs/4871118668/jobs/8687866811?pr=787

Unless @ntabris thinks theres a quick workaround for this on the package sync side, then your move `import` + `PipInstall` suggestion seems good to me

milesgranger:
Okay @jrbourbeau I've removed the torchvision dep, and installed via `PipInstall` at cluster runtime for now.

I do see now running it locally (via Coiled), and on the benchmarks run before, [it's failing with "No module named 'tests'](https://cloud.coiled.io/clusters/207039/overview?account=dask-benchmarks&tab=Logs&filterPattern=):
```
distributed.protocol.pickle - INFO - Failed to deserialize b'\x80\x05\x95\x87\x11\x00\x00\x00\x00\x00\x00\x8c\x1e....'
  File "/opt/coiled/env/lib/python3.9/site-packages/distributed/protocol/pickle.py", line 94, in loads
    return pickle.loads(x, buffers=buffers)
ModuleNotFoundError: No module named 'tests'
```

jrbourbeau:
My guess is that this is related to `client.upload_file` not including the scheduler

milesgranger:
@jrbourbeau It was due to referencing functions/variables outside of the test function is all, the simple example code here will cause the same error `No module named 'tests'`: https://github.com/coiled/benchmarks/pull/787/commits/9e73511a304dba107d40befe83e183216ca0d118

Very tedious to debug that one, sunk some time into it. :(

milesgranger:
Ready for another look, after sorting out the urllib3 issue by ensuring <2.0.0 is installed. 
Now using 'real' data, we still manage to make good use of the GPUs:


ezgif-2-2f6fac7f5a

hendrikmakait (approved):
Generally LGTM, I'd like to see the number of workers increase before merging.

.github/workflows/tests.yml
```diff
@@ -97,7 +97,7 @@ jobs:
           # Need to use `main` branch of optuna until https://github.com/optuna/optuna/pull/4589
           # is released in `optuna=3.2.0`. Currently only doing this when workflows are run
           # as that's the only place where the optuna + dask integration is being used.
-          python -m pip install git+https://github.com/optuna/optuna
+          python -m pip install git+https://github.com/optuna/optuna.git@378508ab6bddbad182bdfa0e8b3ad4bbb7040f00
```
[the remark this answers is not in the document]
jrbourbeau: This is the commit where the fix referenced in the commit above was included. We could install directly from `main`, but pinning to this commit seemed favorable because it's less of a moving target.

milesgranger:
@jrbourbeau @hendrikmakait I think we might be good here? Let me know if any other changes are requested! :)

Issue #732, opened 2023-03-24
Stop maintaining `coiled-runtime` metapackage


jrbourbeau:
_There have been some offline conversations / emails on this topic. I'm summarizing those here._ 

Today we have a lot of infrastructure / CI jobs around maintaining the `coiled-runtime` metapackage, but we're not really getting much value out of it anymore*. Coiled has moved away from `coiled-runtime` in favor of `package_sync` and we haven't really been actively maintain it. There are also some cases where the metapackage slows down our benchmarking efforts, where we do find lots of value (e.g. https://github.com/coiled/coiled-runtime/issues/650). 

If we're not getting value out of the metapackage, and in some cases it slows things down that we do care about, we should just drop the metapackage. Instead, we could just keep a set of environment files we use for benchmarking and are comfortable updating / patching as needed (e.g. `numba` doesn't have Python 3.11 support, so we just comment it out of the 3.11 environment file until it's available). This will let us clean things up a bit here, remove several CI jobs, and adapt more quickly on benchmarking (which is the main thing we care about at this point). 

*@crusaderky mentioned he finds being able to easily install a collection of tested packages useful, for example, when doing demos. He finds the metapackage useful for this purpose. He also said that pointing `mamba` at an environment file in this repo is an okay compromise.

dchudz:
Would we continue pushing a docker image with the environment that the benchmarks use, or no? (I'm not saying we need to continue that, but various folks do find it useful.)

jrbourbeau:
I don't think we've needed to touch that CI job in a long time, so we could probably keep it around if it's useful. Who is using those images?

ntabris:
I currently use them when I want to boot an ARM cluster, but I think we'll have package and/or "create software environment" for ARM in next couple weeks, so I personally care less about multiarch images going away.

jrbourbeau:
Sounds good. We can definitely coordinate to make sure things are smooth on you end. Is there anything else platform folks are using `coiled-runtime` for?

crusaderky:
Another good reason for dropping the metapackage, besides numba:
xarray stopped supporting Python 3.8, while AFAIK we don't expect to drop it anytime soon.

So we need a Python 3.8-specific recipe that pins either an older version of xarray, or no xarray at all.

mrocklin:
Is there anything else to do here?  It sounds like platform wants the docker image and James thinks that that's fine.  I think that we're good?

jrbourbeau:
[...]

PR #793, opened 2023-04-13
Remove metapackage

jrbourbeau:
xref https://github.com/coiled/benchmarks/issues/732


jrbourbeau (commented):
@crusaderky could I get your thoughts on this (generally)? I'd think this will simplify things

jrbourbeau:
I should also mentioned that I spoke with @ntabris offline and he's okay dropping the docker images at this point (xref https://github.com/coiled/benchmarks/pull/793)

.github/workflows/tests.yml
```diff
@@ -94,14 +86,13 @@ jobs:
...
 
-      - name: Run Coiled Runtime Tests
+      - name: Run tests
         id: test
         env:
           DASK_COILED__TOKEN: ${{ secrets.COILED_BENCHMARK_BOT_TOKEN }}
           AWS_ACCESS_KEY_ID: ${{ secrets.RUNTIME_CI_BOT_AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.RUNTIME_CI_BOT_AWS_SECRET_ACCESS_KEY }}
-          COILED_RUNTIME_VERSION: ${{ matrix.runtime-version }}
-          DB_NAME: ${{ matrix.os }}-${{ matrix.runtime-version }}-py${{ matrix.python-version }}.db
```
crusaderky, line 104: This was used to label the timeseries. It's also needed for compatibility with A/B testing. No problem in renaming it ofc.

ci/environment.yml
```diff
@@ -1,13 +1,46 @@
...
+  - toolz ==0.12.0
+  - zict ==2.2.0
+  - xgboost ==1.7.4
+  - dask-ml ==2023.3.24
+  - openssl >1.1.0g
+  - optuna ==3.1.0
+  - scipy ==1.10.1
+  - sqlalchemy ==1.4.46  # FIXME https://github.com/dask/dask/issues/9896
+  - pynvml ==11.5.0
+  - bokeh ==2.4.3  # FIXME https://github.com/dask/distributed/issues/7173
```
crusaderky, line 36: #804

ci/environment.yml
```diff
@@ -1,13 +1,46 @@
 channels:
   - conda-forge
 dependencies:
+  - python >=3.8,<3.11  # numba is not yet available on Python 3.11
+  - coiled
+  - numpy ==1.23.5
+  - pandas ==1.5.3  # 2.0.0 requires a newer version of xarray
+  - dask ==2023.3.2
+  - distributed ==2023.3.2
```
crusaderky, line 9: #804

ci/scripts/run_tests.sh
```diff
@@ -4,13 +4,7 @@ set -o xtrace
...
 BENCHMARK="${BENCHMARK:-false}"
 
-# Ensure we run additional tests when testing the latest coiled-runtime
-if [[ $COILED_RUNTIME_VERSION = 'latest' || $COILED_RUNTIME_VERSION = 'upstream' ]]
-then
-  EXTRA_OPTIONS="--run-latest"
-else
-  EXTRA_OPTIONS=""
-fi
+EXTRA_OPTIONS="--run-latest"
```
crusaderky, line 7: The whole point of it was to have the test suite run on very obsolete versions of dask. It feels redundant now and I think it should be removed completely.

dashboard.py
```diff
@@ -735,17 +712,15 @@ def main() -> None:
     df = pandas.read_sql(
         "select * from test_run where platform = 'linux' "
         "and call_outcome in ('passed', 'failed')"
-        "and coiled_runtime_version >= '0.2.1' ",
+        "and coiled_runtime_version = 'upstream'",
```
crusaderky, line 715: I must have missed this, is it hardcoded somewhere?
Also, as mentioned I think we need to preserve the distintion between latest and upstream.
Finally, I don't think this will work with A/B tests.

dashboard.py
```diff
@@ -762,12 +737,12 @@ def main() -> None:
...
     # Generate HTML pages
-    runtimes = sorted(df.runtime.unique(), key=runtime_sort_key)
+    runtimes = sorted(df.runtime.unique(), key=natural_sort_key)
     for runtime in runtimes:
         make_timeseries_html_report(df, output_dir, runtime)
 
-    for (runtime, fullname), df2 in df.groupby(["runtime", "fullname"]):
-        make_details_html_report(df2, output_dir, runtime, fullname)
+    # for (runtime, fullname), df2 in df.groupby(["runtime", "fullname"]):
+    #     make_details_html_report(df2, output_dir, runtime, fullname)
```
crusaderky, line 745: Forgot to uncomment?

crusaderky (changes requested):

.github/workflows/tests.yml
```diff
@@ -24,7 +24,7 @@ defaults:
 
 jobs:
   tests:
-    name: Tests - ${{ matrix.runtime-version }} ${{ matrix.os }} py${{ matrix.python-version }}
+    name: Tests - ${{ matrix.os }} py${{ matrix.python-version }}
```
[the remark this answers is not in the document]
jrbourbeau: Could you elaborate more on what value you see on continuing to test a latest and upstream build? Given our release cadence, I imagine we'll be updating the definition of latest roughly every two weeks without thinking much about it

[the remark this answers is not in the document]
crusaderky: The benefit is that if somebody merges into main a performance regression, with `upstream` the CI bot will open a ticket for us the next morning, whereas without somebody will need to remember to manually update the environment here. In case of perceivedly uneventful releases, this may potentially not happen for a few months.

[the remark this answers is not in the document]
jrbourbeau: Ah, I see. Sorry, it may not be clear based on the current state of this PR. I'd like to only have an `upstream` build that tests the latest `main` of `dask` / `distributed`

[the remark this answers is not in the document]
crusaderky: Ah, I see. I'm happy to drop `latest`.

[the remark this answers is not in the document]
jrbourbeau: 👍 

This PR has gotten stale. I'll bring it back to life either later today or tomorrow

crusaderky:
@jrbourbeau sorry, I didn't have time to look at this. I'll be away until the 22nd; you may want to pull in somebody else for review.

crusaderky:
@jrbourbeau see my review commit.

- Could you produce evidence that this doesn't cause regressions in the A/B tests? Your version couldn't work as it was hardcoding the coiled_runtime_environment.
- As per @scharlottej13's comment, could you review the README?

crusaderky (approved):

jrbourbeau:
Thanks for getting this across the finish line @crusaderky

Issue #756, opened 2023-03-31
ValueError: backend_options should be an instance of coiled.BackendOptions


crusaderky:
I'm failing to run any kind of A/B tests on Python 3.10 or 3.11.
`coiled.Cluster` fails with `ValueError: backend_options should be an instance of coiled.BackendOptions`.
e.g. https://github.com/coiled/coiled-runtime/actions/runs/4574323397
This only happens on A/B tests, and only on Python 3.10 and 3.11.
Regular PR/overnight tests on 3.10, jupyter notebooks on 3.10, and 3.8/3.9 A/B tests are fine.

Everything uses coiled-0.5.9.

CC @fjetter @shughes-uk @ntabris

dchudz:
@crusaderky we can improve the error message here, but to immediately tell you what's wrong I'd need some logging or a minimal reproducer 

what's ` **cluster_kwargs["small_cluster"],` in the failing example?

jrbourbeau:
I was seeing this over in https://github.com/coiled/coiled-runtime/pull/752 yesterday, but now no longer am (at least the latest CI run passed). Was this fixed on the `coiled` side, or is this just a happy accident?

dchudz:
I'm not aware of any relevant coiled-side changes.

crusaderky:
These are the kwargs:
```yaml
# Settings for all clusters, unless overriden below
default:
  package_sync: true
  wait_for_workers: true
  scheduler_vm_types: [m6i.large]
  backend_options:
    send_prometheus_metrics: true
    spot: true
    spot_on_demand_fallback: true
    multizone: true
  n_workers: 10
  worker_vm_types: [m6i.large]  # 2CPU, 8GiB
```

crusaderky:
Reproduced again today:
https://github.com/coiled/benchmarks/actions/runs/4691097776
https://github.com/coiled/benchmarks/tree/guido/AB_crash/AB_environments

shughes-uk:
Can you print the `backend_options` setting in those tests? Really need to understand the shape of the object being passed in

crusaderky:
> These are the kwargs:
> 
> ```yaml
>   package_sync: true
>   wait_for_workers: true
>   scheduler_vm_types: [m6i.large]
>   backend_options:
>     send_prometheus_metrics: true
>     spot: true
>     spot_on_demand_fallback: true
>     multizone: true
>   n_workers: 10
[...]

PR #794, opened 2023-04-14
Dump extra info in github artifacts

crusaderky:
Dump all kwargs passed to `coiled.Cluster`, as well as the output of `mamba env export`, to the github artifacts. This can be useful to debug issues.

XREF #756


phofl (commented):
lgtm

hendrikmakait (approved):
Thanks, @crusaderky!

PR #795, opened 2023-04-14
Remove redundant `python-version` from CI

crusaderky:
Remove python-version from CI scripts when it's already pinned in the conda recipe. This change is purely cosmetic; `conda-incubator/setup-miniconda@v2` does the right thing and ignores it.

### DIFF
--- a/.github/workflows/ab_tests.yml
+++ b/.github/workflows/ab_tests.yml
@@ -62,14 +62,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Set up environment
-        uses: conda-incubator/setup-miniconda@v2
-        with:
-          miniforge-variant: Mambaforge
-          use-mamba: true
-          condarc-file: ci/condarc
-          environment-file: ci/environment.yml
-
       - name: Create null hypothesis as a copy of baseline
         if: matrix.runtime-version == 'AB_null_hypothesis'
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/ab_tests.yml
```diff
@@ -69,7 +68,6 @@ jobs:
           miniforge-variant: Mambaforge
           use-mamba: true
           condarc-file: ci/condarc
-          python-version: ${{ matrix.python-version }}
```
jrbourbeau, line 72: I could be missing something, but it looks like `python` is unpinned in `ci/environment.yml`

https://github.com/coiled/benchmarks/blob/278d82b3e282f59e97ed789c82f5ef018ff6c69b/ci/environment.yml#L8

My guess is there's another environment file that comes into play here, but just wanted to double check

jrbourbeau (commented):
Thanks @crusaderky

.github/workflows/ab_tests.yml
crusaderky: Ah. I just realised. I'm reworking and retesting.

jrbourbeau:
FWIW I'm definitely not against removing redundant information, but wanted to point out https://github.com/coiled/benchmarks/pull/793 where I'm working on removing the metapackage altogether.

crusaderky:
This needs more tweaking.

The main branch currently installs `ci/environment.yml` with Python 3.9, which is then updated with `AB_environments/AB_*.conda.yaml`. This may update the major python version, which is at the very least a major source of concern (see #756).

On the other hand, there are version conflicts between the two files, e.g. pinned versions of `sqlalchemy` and `filelock` that get unpinned on the `mamba env update`.

PR #800, opened 2023-04-19
Add context when using default worker instance type

jrbourbeau:
`m6i.xlarge` is now our default instance type, so we should need to specify these `worker_vm_types` anymore 

cc @ntabris

### DIFF
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -34,16 +34,12 @@ parquet_cluster:
 # For tests/workflows/test_embarrassingly_parallel.py
 embarrassingly_parallel_cluster:
   n_workers: 100
-  # TODO: Remove the `m6i.xlarge` worker specification below
-  # once it's the default worker instance type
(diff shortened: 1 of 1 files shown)


ntabris:
The current default on Coiled is `["m6i.xlarge", "t3.xlarge"]`. I think this is likely to be better for regular Coiled use because you'll get t3 workers if m6i aren't available. But it might not be what you want for a benchmark, since this will potentially lead to unpredictable variation in runtime.

jrbourbeau:
Ah, I see. Thanks for clarifying @ntabris. In that case, I agree sticking with `m6i.xlarge` will make for better benchmarking and also be close-ish to what users also get by default. I've updated this PR to just tweak the comments around the worker instance specification

ntabris (approved):
LGTM

ntabris:
Thanks, @jrbourbeau! Apologies for not making that clearer sooner.

Issue #751, opened 2023-03-30
Image analysis


jrbourbeau:
There is a surprisingly large community of people using Dask for bio-medical imaging. This includes applications like fMRI brain scans, and very high resolution microscopy (3d movies at micro resolution of cells). These folks often want to load in data, apply image processing filters across that data using map_overlap, and then visually explore the result. They want this processing done with human-in-the-loop systems.

A representative example would be to:

1. Load data with `from_zarr`
2. Perform `map_overlap` with a trivial function
3. Then select a sequence of regions one after the other

jrbourbeau:
@GenevieveBuckley this is part of our broader effort to include large-scale, representative user workflows in examples and decisions that inform future development of Dask (xref https://github.com/coiled/coiled-runtime/issues/725). I'd love to get your thoughts on what you think makes a good representative image analysis workflow. Thoughts on a nice public dataset? Maybe you already have a notebook that shows this type of workflow off?

GenevieveBuckley:
I'd say typical image analysis workflows often have this structure:
1. Loading data
2. Preprocessing images *(an interactive preview here would help choosing the right parameters)*
3. Identifying objects *(also a good spot for interactive use, you almost always need to manually correct segmentations in some way)*
4. Measuring objects

Loading data 
 * You can probably skip most of the annoying parts of this, which include moving your data, and probably converting it into a different format (like zarr) before you load and process it. 

Preprocessing images could include things like:
 * Filtering images (gaussian, median, etc. dask-image has several of these basic filters available)
  * Denoising images with deep learning (eg: [Noise2Self](https://github.com/czbiohub/noise2self), [Noise2Void](https://github.com/juglab/n2v))
 * Deconvolving images (there's a [blogpost with an ITK deconv example here](https://blog.dask.org/2019/08/09/image-itk), there's also a lot of CUDA deconvolution options eg [pycudadecon](https://github.com/tlambert03/pycudadecon))

Identifying objects usually involves
 * Segmentation of objects. (Here's a very basic scikit-image [label image](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_label.html) example). 
 * Cleaning up the segmentation - often there is manual correction that needs to be done. 

Measuring objects
 * What people want to measure depends on their use case, and the question they're asking
[...]

PR #801, opened 2023-04-20
[WIP] Image analysis workflow

jrbourbeau:
This is an initial pass at the example outlined here https://github.com/coiled/benchmarks/issues/751#issuecomment-1514193327. Could still use some finishing steps, but this contains (I think) the bulk of the logic. Pushing up now for any early feedback. 

cc @GenevieveBuckley @mrocklin for visibility

### DIFF
--- a/tests/workflows/test_image_analysis.py
+++ b/tests/workflows/test_image_analysis.py
@@ -1,0 +1,30 @@
+import dask
+import dask.array as da
+import numpy as np
+from dask_image import ndfilters, ndmeasure, ndmorph
+
+
+def test_BBBC039(small_client):
+    images = da.from_zarr(
+        "s3://coiled-datasets/BBBC039", storage_options={"anon": True}
+    )
+    smoothed = ndfilters.gaussian_filter(images, sigma=[0, 1, 1])
+    thresh = ndfilters.threshold_local(smoothed, block_size=images.chunksize)
+    threshold_images = smoothed > thresh
+    structuring_element = np.array(
+        [
+            [[0, 0, 0], [0, 0, 0], [0, 0, 0]],
+            [[0, 1, 0], [1, 1, 1], [0, 1, 0]],
+            [[0, 0, 0], [0, 0, 0], [0, 0, 0]],
+        ]
+    )
+    binary_images = ndmorph.binary_closing(
+        threshold_images, structure=structuring_element
+    )
+    label_images, num_features = ndmeasure.label(binary_images)
+    index = np.arange(num_features)
+    # FIXME: Only selecting the first few images due to cluster idle timeout.
+    # Maybe sending large graph? Need to investigate a bit.
+    area = ndmeasure.area(images[:3], label_images[:3], index)
+    mean_intensity = ndmeasure.mean(images[:3], label_images[:3], index)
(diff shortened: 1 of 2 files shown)


tests/workflows/test_image_analysis.py
```diff
@@ -0,0 +1,30 @@
+import dask
+import dask.array as da
+import numpy as np
+from dask_image import ndfilters, ndmeasure, ndmorph
+
+
+def test_BBBC039(small_client):
+    images = da.from_zarr(
+        "s3://coiled-datasets/BBBC039", storage_options={"anon": True}
+    )
```
jrbourbeau, line 10: I converted the original `.tif` dataset to Zarr format and uploaded them to a public S3 bucket. This had the side-effect of bypassing https://github.com/dask/dask-image/issues/84.

tests/workflows/test_image_analysis.py
```diff
@@ -0,0 +1,30 @@
...
+            [[0, 0, 0], [0, 0, 0], [0, 0, 0]],
+        ]
+    )
+    binary_images = ndmorph.binary_closing(
+        threshold_images, structure=structuring_element
+    )
+    label_images, num_features = ndmeasure.label(binary_images)
+    index = np.arange(num_features)
+    # FIXME: Only selecting the first few images due to cluster idle timeout.
+    # Maybe sending large graph? Need to investigate a bit.
```
GenevieveBuckley, line 27: I have no useful suggestions, but it's great* that this has possibly already identified some kind of problem your users might occasionally run into.

* You know, great your users, but admittedly not so great for the person who now needs to investigate it 😆 Have fun with that, James!

tests/workflows/test_image_analysis.py
```diff
@@ -0,0 +1,30 @@
+import dask
+import dask.array as da
+import numpy as np
+from dask_image import ndfilters, ndmeasure, ndmorph
+
+
+def test_BBBC039(small_client):
```
GenevieveBuckley, line 7: Suggest adding a link to [those talk slides](https://genevievebuckley.github.io/dask-image-talk-2020/) and/or [the repo](https://github.com/GenevieveBuckley/dask-image-talk-2020). That way if anyone needs to work on this benchmark later, they can look there for context/answers there before quizzing James about it.

tests/workflows/test_image_analysis.py
```diff
@@ -0,0 +1,30 @@
...
+    images = da.from_zarr(
+        "s3://coiled-datasets/BBBC039", storage_options={"anon": True}
+    )
+    smoothed = ndfilters.gaussian_filter(images, sigma=[0, 1, 1])
+    thresh = ndfilters.threshold_local(smoothed, block_size=images.chunksize)
+    threshold_images = smoothed > thresh
+    structuring_element = np.array(
```
GenevieveBuckley, line 14: Suggest adding a short comment: to this line about why we're not using the default structuring element. Eg:
```python
# Since this image stack appears to be 3-dimensional,
# we sandwich a 2d structuring element in between zeros
# so that each 2d image slice has the binary closing applied independently
```
Apparently I only ever said that verbally during the talk, may as well write it down.

PR #802, opened 2023-04-21
Add pytest marker/fixture for client creation

jrbourbeau:
Currently there's quite a bit of boilerplate code that's getting copy-pasted in each workflow module. This PR moves that code into a pytest marker + fixture for reuse / a more pleasant interface. 

Steps here are to:

1. Add named cluster spec to `cluster_kwargs.yaml`
2. Mark test with `@pytest.mark.client(<spec-name>)`
3. Add the `client` fixture to the test (the materialized `client` will be attached to a cluster with the given spec)

This seems like a decent improvement to me

### DIFF
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -32,19 +32,19 @@ parquet_cluster:
   worker_vm_types: [m5.xlarge]  # 4 CPU, 16 GiB
 
 # For tests/workflows/test_embarrassingly_parallel.py
-embarrassingly_parallel_cluster:
+embarrassingly_parallel:
   n_workers: 100
   worker_vm_types: [m6i.xlarge] # 4 CPU, 16 GiB (preferred default instance)
   backend_options:
     region: "us-east-1"  # Same region as dataset
 
-# For tests/workflows/test_optuna.py
-optuna_cluster:
+# For tests/workflows/test_xgboost_optuna.py
+xgboost_optuna:
   n_workers: 50
   worker_vm_types: [m6i.xlarge]  # 4 CPU, 16 GiB (preferred default instance)
 
 # For tests/workflows/test_uber_lyft.py
-uber_lyft_cluster:
+uber_lyft:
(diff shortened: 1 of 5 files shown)


tests/workflows/test_embarrassingly_parallel.py
```diff
@@ -74,11 +42,11 @@ def extract(filename: str, fs):
                             out.append((member.name, b"matplotlib" in data.lower()))
                 return out
 
-    futures = embarrassingly_parallel_client.map(extract, directories, fs=s3)
+    futures = client.map(extract, directories, fs=s3)
```
jrbourbeau, line 45: Using `client` instead of, for example, `embarrassingly_parallel_client`, also seems like a small victory

cluster_kwargs.yaml
```diff
@@ -32,19 +32,19 @@ parquet_cluster:
...
-embarrassingly_parallel_cluster:
+embarrassingly_parallel:
   n_workers: 100
   worker_vm_types: [m6i.xlarge] # 4 CPU, 16 GiB (preferred default instance)
   backend_options:
     region: "us-east-1"  # Same region as dataset
 
-# For tests/workflows/test_optuna.py
-optuna_cluster:
+# For tests/workflows/test_xgboost_optuna.py
```
ncclementi, line 41: Just making sure this is just a change in the comment, right? Because if  the name of the test changed, then there should be a DB migration.

ncclementi (approved):
This LGTM, I see CI is green so this is working. I left only one comment but the change looks neat and witty : )

cluster_kwargs.yaml
jrbourbeau: Yeah, this is just a comment update. The test module is actually already named `test_xgboost_optuna.py`, I just forgot to update this comment in the PR that added the test.

jrbourbeau:
Thanks for the review @ncclementi!

PR #803, opened 2023-04-21
Upgrade to zict 3.0

crusaderky:
Avoids unnecessary deep-copy when unspilling uncompressed numpy arrays (dask/zict#74)

### DIFF
--- a/AB_environments/AB_baseline.conda.yaml
+++ b/AB_environments/AB_baseline.conda.yaml
@@ -31,7 +31,7 @@ dependencies:
     - cloudpickle ==2.2.1
     - tornado ==6.2
     - toolz ==0.12.0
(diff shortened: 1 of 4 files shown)


jrbourbeau (approved):
Fine by me if tests pass. Btw, should we expect to see corresponding performance improvements in certain benchmarks?

crusaderky:
> Fine by me if tests pass. Btw, should we expect to see corresponding performance improvements in certain benchmarks?

"expect" is a big word. This prevents a memory spike on unspill, so we *may* see a reduction in peak memory usage by 120 MiB per worker. Or not, if the peak is caused by something else that isn't unspilling.

crusaderky:
I looked at the plots. Can't see a difference.

PR #805, opened 2023-04-23
Run A/B tests with minimal environments

crusaderky:
Run A/B tests with just dask, distributed, coiled, and s3fs in the environment file

Tested with A/B recipes:
- dask, distributed, coiled, s3fs
- dask, distributed, coiled, s3fs, xarray
- dask, distributed, coiled, s3fs, xarray, cftime, zarr
- dask, distributed, coiled, s3fs, xgboost, dask-ml, optuna
- dask, distributed, coiled, s3fs, scipy


tests/stability/test_array.py
```diff
@@ -5,11 +5,19 @@
 
 from ..utils_test import cluster_memory, scaled_array_shape, wait
 
+try:
+    import scipy  # noqa: F401
+
+    has_scipy = True
+except ImportError:
+    has_scipy = False
```
crusaderky, line 13: Didn't use importorskip because, since there's only one test in this module, the cluster would be started needlessly

crusaderky:
Ready for review

tests/stability/test_array.py
j-bennet: ❤️ I would suggest you add this as an explicit comment in code, otherwise someone (like me) may have an autopilot moment and replace it with `importorskip` without thinking.

AB_environments/README.md
```diff
@@ -46,6 +46,23 @@ forks, e.g.
       - dask ==2023.4.2
       - git+https://github.com/yourname/distributed@803c624fcef99e3b6f3f1c5bce61a2fb4c9a1717
 ```
+
+You may also ignore the recipe file and go for a barebones environment. The bare
```
j-bennet, line 50: Can you elaborate why one might want to do that?

j-bennet (approved):
👍 from me; I would like to add a couple of informative comments, but that should not be blocking.

AB_environments/README.md
crusaderky: For example because you want to see how pandas performs without pyarrow (fall back to fastparquet), how xarray performs without h5netcdf (fall back to scipy), etc

ci/environment.yml
```diff
@@ -7,8 +7,7 @@ dependencies:
   - pytest-xdist
   - python
   - pyyaml
-  - sqlalchemy
+  - sqlalchemy ==1.4.46  # FIXME https://github.com/dask/dask/issues/9896
```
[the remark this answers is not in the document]
crusaderky: Yes, as soon as the next dask release is out we can clean up here. Until it is though I'd rather keep the same pin everywhere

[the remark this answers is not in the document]
phofl: Yep this was meant as a reminder, not a request to change

PR #819, opened 2023-04-27
Fix confidence interval plots

hendrikmakait:
An alternative approach to #817 (which has been developed in parallel).


crusaderky (commented):
I'm not too happy about this PR because
1. it relies on the user not putting any comment on the same line
2. it doesn't work e.g. with `python >=3.8,<3.11`

Overall I think my version has the benefit of simplicity and has trivial maintenance cost.

hendrikmakait:
Superseded by #817

Issue #821, opened 2023-04-27
s3 throughput much higher on some workers?


crusaderky:
![image](https://user-images.githubusercontent.com/6213168/234855554-3ef3056d-c704-4620-8196-6aec73063a27.png)

In this test,
- [s3fs] is a raw file read with fsspec from s3:// of a 562 MiB file, from a single dask task
- [pandas] is a call to pandas that reads and decodes the whole parquet dataset from a single dask task
- [dask] parallelizes the read and decode across 5 different workers (the file is chunked).

The file: `s3://coiled-runtime-ci/ookla-open-data/type=fixed/year=2022/quarter=1/2022-01-01_performance_fixed_tiles.parquet`

The data seems to indicate that, depending on which worker you land, you will get much faster or slower throughput - with up to a factor of 3x variance (8s ~ 22s).

Is this something we're already aware of? Is it to be expected?

CC @ntabris

ntabris:
Just be clear, you're thinking about the variance across distinct runs of the same test?

crusaderky:
@ntabris yes

ntabris:
For the record...

@crusaderky and I had a quick chat and looked at the network throughput metrics on a few runs. The rate metrics we have aren't ideal because the test is so short, but we would still expect to see higher variance in the metrics than we see if network throughput was the primary explanation.

I told @crusaderky that my impression is that the network throughput we get on Coiled reading from s3 could be much better, and pointed him to various related issues:

- https://github.com/coiled/general/issues/154
- https://github.com/fsspec/s3fs/issues/677
- https://github.com/coiled/parquet-scaling-best-practices/issues/11

He mentioned that he might try something like `subprocess.check_call("aws s3 cp s3:// /dev/null", shell=True)` to see what the theoretical limit is, I'd also be interested in seeing that.

My impression is that we aren't opening enough concurrent connection to get good throughput, but I haven't spent a lot of time investigating.

crusaderky:
I inspected the cluster dumps.
In a "fast" run of the [s3fs] variant, which took end-to-end 6.5s, the task was in executing state for 6.48s.
In a "slow" run, which took end-to-end 16.3s, the task was in executing state for 16.25s.
I don't have the fine performance metrics in the dumps but I don't expect them to show anything other than `thread-noncpu`.
So 99% of the runtime - and of the variance - is confirmed to be caused by these 2 lines:
[...]

PR #825, opened 2023-05-02
Test boto3 performance throughput

crusaderky:
Meter boto3 performance when reading a single large file from s3.
This is to have a baseline to compare potential fsspec problems against.

- XREF https://github.com/coiled/benchmarks/issues/821
- XREF https://github.com/coiled/general/issues/154
- XREF https://github.com/fsspec/s3fs/issues/677


crusaderky:
CC @ntabris

tests/benchmarks/test_parquet.py
```diff
@@ -1,17 +1,20 @@
 """
 Parquet-related benchmarks.
 """
+import io
 import uuid
 
+import boto3
```
crusaderky, line 7: dependency of coiled

tests/benchmarks/test_parquet.py
```diff
@@ -1,17 +1,20 @@
...
 import dask.dataframe as dd
 import dask.datasets
 import distributed
+import fsspec
```
crusaderky, line 11: dependency of dask

mrocklin:
ooh, I'm excited to see the results of this

crusaderky:
I ran it 10 times and the results seem to conclude that

- there is no noticeable difference between boto3 and fsspec
- there is a "warm-up" period where s3 will become progressively faster, e.g. your data shifts towards the top of some frequently used cache - until the cache doesn't work for whatever reason and it slows down again
- some tests will be noisier than others just in virtue of where they are in the test suite - first test is always the noisiest because of the above point
- there is no difference in overall performance between reading the file in 5 chunks from 5 different workers vs just sequentially from a single worker
- decoding the dataframe with pyarrow takes a huge amount of time, and parallelizing the effort across 5 workers (`[dask]`) makes a difference

image

crusaderky:
Ready for review and merge

phofl:
> decoding the dataframe with pyarrow takes a huge amount of time, and parallelizing the effort across 5 workers ([dask]) makes a difference

I guess you are referring to the conversion from the Arrow Table to NumPy arrays? This is one of the best use-cases for the general purpose ``dtype_backend="pyarrow"`` in pandas 2.0. This is a no-copy op in this scenario and is pretty fast (I saw speedups up to 10 times the NumPy performance for the whole read_parquet operation in pandas).

Do you think it makes sense to add one benchmark for this case as well? Would have to use pandas 2.0 though

milesgranger (approved):
LGTM

ntabris:
What do you think about having each test download the file N times (say, 4 or 8)? That might make tests more consistent since (I'm guessing) majority of those downloads will be after s3 has moved file to hot storage.

FWIW the question I was interested in is whether we'd get much better network throughput by using multiple concurrent downloads. I don't think this answers that.

crusaderky:
> What do you think about having each test download the file N times (say, 4 or 8)? That might make tests more consistent since (I'm guessing) majority of those downloads will be after s3 has moved file to hot storage.

I thought about that, but I think we're already spending enough time on this test...

ntabris:
> I thought about that, but I think we're already spending enough time on this test...

As in developer time or computer time?

crusaderky:
> > I thought about that, but I think we're already spending enough time on this test...
> 
> As in developer time or computer time?

both

PR #826, opened 2023-05-03
[DNM] Try arxiv with more threads than cores

ntabris:


mrocklin:
Oh fun.  I'm curious to see what happens here.

PR #827, opened 2023-05-03
Snowflake workflow, v1, read/write only

j-bennet:
Snowflake workflow using NYC bike data (csv).

This version is simply benchmarking reading and writing to Snowflake, without extras.

Extras will be added in v2.

**Important**: `snowflake-connector-python` requires `pandas >1.0.0,<1.6.0`, and we want to use 2.0 in dask. This PR can't be merged as-is. The Snowflake workflow would need its own CI job with its own separate conda requirements.

**Update 5/16/2023:**

`snowflake-connector-python` just bumped the reqs to `pandas>=1.0.0,<2.1.0`.

https://github.com/snowflakedb/snowflake-connector-python/pull/1496


tests/workflows/test_snowflake.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # repartition to ensure even chunks
+    ddf = ddf.repartition(partition_size="100Mb")
+
+    input_npartitions = ddf.npartitions
+
+    wait(ddf)
+
+    # save data to Snowflake
+    to_snowflake(ddf, name=table, connection_kwargs=connection_kwargs)
```
mrocklin, line 117: I think that we care about reading from snowflake much much more than we care about writing to it.  I would love to have a workflow that focuses just on reading.  I think that that will help us to isolate those parts of the workflow (I also expect writing to cost much more than reading, and so dominate the benchmark).

With that in mind, maybe we can upload a dataset into Snowflake and then just keep it there?

tests/workflows/test_snowflake.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    # maybe plot roundtrips vs one-way
+    import seaborn as sns
+    sns.barplot(roundtrips, x="is_roundtrip", y="count")
+    """
+
+    # explore trip duration
+    ddf_out["trip_duration"] = ddf_out["ended_at"] - ddf_out["started_at"]
+    ddf_out["trip_minutes"] = ddf_out["trip_duration"].apply(
+        lambda x: x.total_seconds() // 60, meta=("trip_duration", int)
```
mrocklin, line 142: Using the apply here is probably going to be very slow.  I recommend using  `.dt.total_seconds` instead.  In general any time pandas uses apply it runs at Python rather than Cython speeds, and so is very slow.

tests/workflows/test_snowflake.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+
+    # filter out incorrect station IDs
+    ddf = ddf[(ddf.start_station_id != -1) & (ddf.end_station_id != -1)].reset_index(
+        drop=True
+    )
+
+    # create boolean is_member and drop member_casual
+    ddf["is_member"] = ddf.member_casual.apply(
+        lambda x: x == "member", meta=("is_member", bool)
```
phofl, line 105: I'd recommend using 
```
ddf.member_casual == "member"
``` 

apply is slow unfortunately

recipe/meta.yaml
```diff
@@ -26,6 +27,9 @@ requirements:
     - numpy ==1.23.5
     - pandas ==1.5.3  # 2.0.0 requires a newer version of xarray
     - dask =={{ dask_version }}
+{% if python_version in ["3.8", "3.9"] -%}
+    - dask-snowflake
+{%- endif %}
```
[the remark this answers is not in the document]
j-bennet: Initially I didn't have this clause here, and CI broke trying to install requirements:

https://github.com/coiled/benchmarks/actions/runs/4875870950/jobs/8698634340

I'm pretty sure it's because `snowflake-connector-python` pins `numpy` to `<1.25.0`:

https://github.com/snowflakedb/snowflake-connector-python/blob/97c25993eb3fb43e0df634fd41f98f9e9ebf20bd/setup.cfg#L85

Lower version of `numpy` can't run on `python>3.9`.

CI still breaks though, there must be something else I'm missing.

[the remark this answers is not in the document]
j-bennet: Definitely looking forward to https://github.com/coiled/benchmarks/pull/793!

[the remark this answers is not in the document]
phofl: There is something weird going on though, the job is pulling numpy 1.19, which indeed isn't provided for python 3.10, but snowflake-connector-python allows 1.24 for example, which is the newest released version of numpy, something else seems to restrict numpy

tests/workflows/test_snowflake.py
```diff
@@ -0,0 +1,172 @@
...
+    ddf = ddf.repartition(partition_size="100Mb")
+
+    wait(ddf)
+
+    # save data to Snowflake
+    to_snowflake(ddf, name=table, connection_kwargs=connection_kwargs)
+
+
+@pytest.mark.client("snowflake")
+def test_read_explore(client, connection_kwargs, table):
```
hendrikmakait, line 123: This workflow feels like something that I personally would solve in Snowflake directly.

Is there a way we can make this more iterative/explorative so that we'd benefit from having intermediate results available on the cluster? Alternatively, could we make more use of different Dask libraries/integrations? This might look more similar to https://github.com/Snowflake-Labs/snowpark-python-demos.

hendrikmakait: I'm not sure what the exact motivation of this is, maybe we could also go the other way and dumb it down until it becomes a narrow test of Snowflake read performance.

j-bennet: > Is there a way we can make this more iterative/explorative so that we'd benefit from having intermediate results available on the cluster

That't why I added quantiles and filtering out outliers.

> I'm not sure what the exact motivation of this is, maybe we could also go the other way and dumb it down until it becomes a narrow test of Snowflake read performance

I'm a little struggling with this myself. It seems that everyone has a different idea about the purpose of this, and how complicated / representative of a real-world it should look. This is an attempt to find a balance between "very dumb read and calculate a single mean or sum" and "read and do something very complicated, like train an ML model, or perform a full-on ETL pipeline".

Maybe @jrbourbeau and @mrocklin can advise?

hendrikmakait:
@j-bennet: I suppose your work will supersede #449?

j-bennet:
> @j-bennet: I suppose your work will supersede #449?

Only if it's merged. :)

I didn't know we had a previous attempt! Should I borrow some things from there?

hendrikmakait:
> I didn't know we had a previous attempt! Should I borrow some things from there?

Honestly, I don't know. I've only stumbled across this myself while checking out open PRs.

ncclementi:
I knew about this workflow and brought it a couple of times on standup, more recently when the snowflake connector started accepting pyarrow 10 [see this comment](https://github.com/coiled/benchmarks/pull/449#issuecomment-1450903771) sorry I didn't mention it again, I assumed you and the team were aware of it and decided to start from a clean place. 

The main issue that Paul landed in his workflow was related to getting the right dependencies see https://github.com/coiled/benchmarks/pull/449#issuecomment-1310769736. The snowflake connector has very specific pins for pyarrow, so we will probably hit this issue again everytime we want to be on the latest pyarrow. See https://github.com/snowflakedb/snowflake-connector-python/blob/c371d9fb01f57df59847e61c5315754c2f9b6fad/pyproject.toml#L9 and related discussions https://github.com/snowflakedb/snowflake-connector-python/issues/1248. 

When it comes to the code in the workflow itself all the operations in the workflow, could be done in Snowflake. I wonder if starting with a simple read and write first is ok, and then moving to a second one more "realistic" like what we talked in the last meeting. @jrbourbeau do you have more context here that you can provide?

j-bennet:
@ncclementi 

> When it comes to the code in the workflow itself all the operations in the workflow, could be done in Snowflake. I wonder if starting with a simple read and write first is ok, and then moving to a second one more "realistic" like what we talked in the last meeting.

Yes, I agree. Let's at least get the read/write working for v1. I updated the PR ti simplify, now on to fixing CI.

tests/workflows/test_snowflake.py
```diff
@@ -0,0 +1,110 @@
+import os
+import uuid
+
+import dask.dataframe as dd
+import pandas as pd
+import pytest
+from dask.distributed import wait
+
+snowflake = pytest.importorskip("snowflake")
```
ncclementi, line 9: @j-bennet It looks like this line is causing problems and the tests are being skipped see https://github.com/coiled/benchmarks/actions/runs/4942229870/jobs/8835569338?pr=827#step:6:1068 not sure

j-bennet: Yes, I don't think that the condition in `meta.yaml` worked:

[recipe/meta.yaml](https://github.com/coiled/benchmarks/pull/827/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9a)

Not sure why, must have something to do with `{% set python_version = environ.get("PYTHON_VERSION", '') %}`

ncclementi: Maybe you need to use `{{ python_version }}` to get the variable value in the if condition? 
notice that's what we do in  `dask =={{ dask_version }}`

j-bennet: I got the jinja template right, finally. Now there are dependency errors and it can't solve for the requirements. Will try to fix.

ncclementi: yes this is the snowflake connector being annoying. It looks like we moved to pandas 2.0 recently in the benchmarks, and it's complaining about that

j-bennet: I saw, we want pandas == 2.0.1, but snowflake-connector-python wants it lower.

```
│     └─ snowflake-connector-python 3.0.3 would require
│        └─ pandas >1.0.0,<1.6.0 , which can be installed;
```

they have an open issue about supporting 2.0, and no visible movement on it:

https://github.com/snowflakedb/snowflake-connector-python/issues/1506

j-bennet: I guess if we want to have a Snowflake workflow, we'd have to drop to pandas 1.6, which would be a bummer. Perhaps we should have a separate CI job for it.

ncclementi: Having a separate CI job makes sense to me as it's the only solution, and this will keep happening with snowflake because the connector is always behind with the dependencies support. That being said. I think before doing anything further, this deserves a broader discussion, maybe at some point next week with the team? 
It'll probably be good having a summary in the main conversation of the PR

j-bennet: I added a note in summary.

tests/workflows/test_snowflake.py
```diff
@@ -0,0 +1,110 @@
...
+                is_member boolean not null
+            )"""
+    )
+    yield name
+    # after the data is written, delete table
+    engine.execute(f"DROP TABLE IF EXISTS {name}")
+
+
+@pytest.mark.client("snowflake")
+def test_write(client, connection_kwargs, table):
```
hendrikmakait, line 58: nit:
```suggestion
def test_etl_into_snowflake(client, connection_kwargs, table):
```

This test isn't _just_ writing data to Snowflake, it also involves reading a lot of data from csv and a bit of filtering/transforming.

tests/workflows/test_snowflake.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # create boolean is_member and drop member_casual
+    ddf["is_member"] = ddf.member_casual == "member"
+
+    ddf = ddf.drop(columns="member_casual")
+
+    # repartition to ensure even chunks
+    ddf = ddf.repartition(partition_size="100Mb")
+
+    wait(ddf)
```
hendrikmakait, line 94: Is there a need to wait here?

tests/workflows/test_snowflake.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # save data to Snowflake
+    to_snowflake(ddf, name=table, connection_kwargs=connection_kwargs)
```
hendrikmakait, line 97: Just to be safe, this is synchronous and only returns once all data has been written, correct?

tests/workflows/test_snowflake.py
```diff
@@ -0,0 +1,110 @@
...
+
+@pytest.mark.client("snowflake")
+def test_read(client, connection_kwargs):
+    """Read and explore NYC bike dataset from Snowflake"""
+    table = "citibike_tripdata"  # persistent table
+
+    df = read_snowflake(
+        f"SELECT * FROM {table}",
+        connection_kwargs=connection_kwargs,
+        npartitions=200,
```
hendrikmakait, line 108: I suspect users are more likely to specify `partition_size` than `npartitions` similar to the repartition in the write workload. Should we change (or simply remove this, from the docs it looks like we default to `100 MiB`).

tests/workflows/test_snowflake.py
j-bennet: Yes.

tests/workflows/test_snowflake.py
j-bennet: The default is None:
https://github.com/coiled/dask-snowflake/blob/fbfa369f78170015cfdf23cd1cc7918b1deb8fdf/dask_snowflake/core.py#L210
but it's a good point, most people would probably use `100MiB`. I'll change that.

.github/workflows/tests.yml
```diff
@@ -93,17 +101,29 @@ jobs:
           echo $EXTRA_OPTIONS
           echo EXTRA_OPTIONS=$EXTRA_OPTIONS >> $GITHUB_ENV
 
+      - name: Install Optuna
+        if: steps.workflows.outcome == 'success' && matrix.extra != 'snowflake'
```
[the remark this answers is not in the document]
j-bennet: More explicit how?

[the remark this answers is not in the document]
hendrikmakait: The comment works for now, later we may want to make `"optuna"` its own `matrix.extra` which we can explicitly check for.

tests/workflows/test_snowflake.py
hendrikmakait: Thanks for clarifying!

tests/workflows/test_snowflake.py
hendrikmakait: Sorry, what I meant was that we default within the function body:

https://github.com/coiled/dask-snowflake/blob/fbfa369f78170015cfdf23cd1cc7918b1deb8fdf/dask_snowflake/core.py#L263-L264

Issue #808, opened 2023-04-24
AB test copy-on-write in `pandas`


jrbourbeau:
I'm curious if we'll see performance improvements here 

xref https://github.com/dask/dask/issues/10218

cc @phofl

fjetter:
We talked about this offline a bit. The theory so far is that we could possibly see substantial improvements since some keys may now actually share memory since some keys would just be views instead of deepcopies.

However a concern was raised that this would also mess with distributed instrumentation about memory, particularly managed vs unmanaged memory. This would not be a dealbreaker of course.

jrbourbeau:
> However a concern was raised that this would also mess with distributed instrumentation about memory, particularly managed vs unmanaged memory

Just to clarify, the concern here is that `distributed` may overestimate things because `sizeof` isn't aware of views vs. non-views?

crusaderky:
> 



> > However a concern was raised that this would also mess with distributed instrumentation about memory, particularly managed vs unmanaged memory
> 
> Just to clarify, the concern here is that `distributed` may overestimate things because `sizeof` isn't aware of views vs. non-views?

Yes, I think we'll need some variant of `sizeof` that returns a dict of {buffer id: nbytes} and a change in SpillBuffer to match

crusaderky:
Let's keep the discussion on pitfalls etc. on https://github.com/dask/dask/issues/10218

phofl:
Here is a link to the benchmarks: https://github.com/coiled/benchmarks/actions/runs/4809354163 

Results are a bit mixed, but A/B tests don't cover most of the methods where CoW would make a difference

crusaderky:
FWY I typically run at least 5, better if 7, runs in A/B. 3 aren't enough to get any statistically significant measure and can easily lead to incorrect conclusions.

phofl:
Thx, will rerun with a higher repeat setting

crusaderky:
In the above link the dashboard is lacking the confidence report due to a bug that's been resolved in main.
I've generated it locally.

Note how the very impressive speedup on the new test `test_dataframe_cow_chain` is accompanied by an equally extreme increase in memory usage.

![image](https://github.com/coiled/benchmarks/assets/6213168/b5371776-eaed-4744-b83d-0bc060237cc1)
![image](https://github.com/coiled/benchmarks/assets/6213168/3689e5d4-f272-4130-8bbe-8e1c0fd2b485)

phofl:
[...]

PR #829, opened 2023-05-05
Add suggestion for an initial repeat value

phofl:
cc @crusaderky 

Based on https://github.com/coiled/benchmarks/issues/808#issuecomment-1532284254

This would have been helpful to me.


hendrikmakait (approved):
Thanks, @phofl!

PR #832, opened 2023-05-05
Add Copy-on-Write method chain to benchmarks

phofl:
These are all methods that will get a speedup by CoW.


hendrikmakait (approved):
Thanks, @phofl!

PR #841, opened 2023-05-10
Workflow: CSV to parquet

milesgranger:
Continuation from #738


tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,140 @@
...
+    ("ActionGeo_ADM1Code", "string[pyarrow]"),
+    ("ActionGeo_Lat", "Float64"),
+    ("ActionGeo_Long", "Float64"),
+    ("ActionGeo_FeatureID", "string[pyarrow]"),
+    ("DATEADDED", "Int64"),
+    ("SOURCEURL", "string[pyarrow]"),
+])
+
+
+def test_from_csv_to_parquet(from_csv_to_parquet_client, s3_factory, s3_url):
```
fjetter, line 104: this test appears to be a bit convoluted.

The ordinary read_csv should just be one of

* `dd.read_csv("s3://gdelt-open-data/events/")`
* `dd.read_csv(files)`

why is this back and forth conversion and the `download` function required?

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,140 @@
...
+                path,
+                sep='\t',
+                names=COLUMNSV1.keys(),
+                dtype=COLUMNSV1,
+                storage_options=s3.storage_options,
+                on_bad_lines="skip"
+            )
+            # TODO: Want to raise ValueError right away, not
+            # inside the graph later which persist (or nothing) would cause.
+            df = df.compute()
```
fjetter, line 123: I don't completely understand why you're doing this but I suggest to either
* Use a `worker_client` explicitly, see https://distributed.dask.org/en/stable/task-launch.html
* Specify sync/threads scheduler explicilty.

The implicit behavior works but has quite a few shortcomings.

milesgranger:
Just to clarify, this is a WIP and not ready for a proper review... only posted to show the reproducer of the TCP error when running on Coiled is all. :)

tests/workflows/test_from_csv_to_parquet.py
fjetter: I'm mostly raising this in case it is related to the problem you were mentioning.

cluster_kwargs.yaml
```diff
@@ -66,3 +66,10 @@ test_work_stealing_on_straggling_worker:
 test_repeated_merge_spill:
   n_workers: 20
   worker_vm_types: [m6i.large]
+
+# For tests/workflows/test_from_csv_to_parquet.py
+from_csv_to_parquet_cluster:
+  n_workers: 30
+  worker_vm_types: [t3.medium]  # 2CPU, 4GiB
```
ncclementi, line 73: Are we not using the default instances on purpose here, is there a reason for using the burstable ones?

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,143 @@
...
+        # Some bad files have '#' in numeric values
+        converters={
+            col: lambda v: float(v.replace("#", "") or "NaN")
+            for col, dtype in SCHEMA.items()
+            if dtype != "string[pyarrow]"
+        },
+    )
+
+    # Now we can safely convert the numeric columns
+    df = df.astype(
```
ncclementi, line 129: it looks like we have pandas 2.0 in the benchmarks now, is there a reason for not using the config to use pyarrow strings?
```python
dask.config.set({"dataframe.convert-string": True});
```

ncclementi (commented):
Thanks @milesgranger 
This LGTM, left a couple of comments to consider.

cluster_kwargs.yaml
milesgranger: No reason other than ignorance. :sweat_smile:

cluster_kwargs.yaml
```diff
@@ -66,3 +66,9 @@ test_work_stealing_on_straggling_worker:
 test_repeated_merge_spill:
   n_workers: 20
   worker_vm_types: [m6i.large]
+
+# For tests/workflows/test_from_csv_to_parquet.py
+from_csv_to_parquet_cluster:
+  n_workers: 30
```
ncclementi, line 72: Sorry @milesgranger I think we still want to explicitly set them with the default, to avoid mixing of instance types. Basically, the same as we do in the case of test_xgboost or embarrasingly_parallel. 
```suggestion
  n_workers: 30
  worker_vm_types: [m6i.xlarge]  # 4 CPU, 16 GiB (preferred default instance)
```

tests/workflows/test_from_csv_to_parquet.py
milesgranger: We could for sure, but given the explicit `SCHEMA` and the required `converters` to parse strings with erroneous `#` in float fields in certain files followed by then explicitly casting them to their correct dtypes afterwards; does setting this config offer any additional benefit? (I genuinely don't know in this case)

cluster_kwargs.yaml
ncclementi, line 72: Sorry @milesgranger I think we still want to explicitly set them with the default, to avoid mixing of instance types. Basically, the same as we do in the case of test_xgboost or embarrasingly_parallel. 
```suggestion
  n_workers: 30
  worker_vm_types: [m6i.xlarge]  # 4 CPU, 16 GiB (preferred default instance)
```

tests/workflows/test_from_csv_to_parquet.py
milesgranger: Oh, or maybe that string being passed to the converter function is then a pyarrow string..

ncclementi: TBH I'm not sure, probably @jrbourbeau or @j-bennet will know better here.

milesgranger: Just in case it does: https://github.com/coiled/benchmarks/pull/841/commits/db73e85e9d9adef68ae179ec03899868e0872f0f

j-bennet: If you already read your data with correct dtypes, you would not need to do `astype`. Lines 131-133 become unnecessary.

How do you read your data with correct dtypes? The converters should return a correct dtype:

```
converters = {}
for field, dtype in SCHEMA.items():
    if dtype == "Int64":
        converters[field] = lambda x: int(x.replace("#", "") or "NaN")
    elif dtype == "Float64":
        converters[field] = lambda x: float(x.replace("#", "") or "NaN")
df = dd.read_csv("data.csv", converters=converters)
```

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,145 @@
...
+        {col: dtype for col, dtype in SCHEMA.items() if dtype != "string[pyarrow]"}
+    )
+
+    df = df.map_partitions(
```
j-bennet, line 135: Why can't you call `drop_duplicates` on the dataframe?

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,145 @@
...
+        lambda xdf: xdf.drop_duplicates(subset=["SOURCEURL"], keep="first")
+    )
+    df["national_paper"] = df.SOURCEURL.str.contains(
+        "washingtonpost|nytimes", regex=True
+    )
+    df = df[df["national_paper"]]
```
j-bennet, line 141: I'm confused. Why read all columns if you're subsetting it to a single column?

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,145 @@
...
+    df = df.persist()
+    assert len(df)
+
+    df.to_parquet(f"{s3_url}/from-csv-to-parquet/", write_index=False)
```
j-bennet, line 145: I don't see cleanup for written parquet anywhere, did I miss it?

j-bennet (commented):
Left some comments, we can also chat in Slack or Google Meet, if that helps.

j-bennet:
I just noticed that CI doesn't actually test workflows, unless it's a nightly job, or pr is marked with `workflows` label:

https://github.com/coiled/benchmarks/blob/3062113a0a33e69c049379817ac7de49ef76004f/.github/workflows/tests.yml#L87-L89

You probably want that label (TIL).

tests/workflows/test_from_csv_to_parquet.py
milesgranger: https://github.com/coiled/benchmarks/pull/738#discussion_r1156711292

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,145 @@
...
+        dtype={
+            col: dtype for col, dtype in SCHEMA.items() if dtype == "string[pyarrow]"
+        },
+        storage_options=s3.storage_options,
+        on_bad_lines="skip",
+        # Some bad files have '#' in numeric values
+        converters={
+            col: lambda v: float(v.replace("#", "") or "NaN")
+            for col, dtype in SCHEMA.items()
+            if dtype != "string[pyarrow]"
```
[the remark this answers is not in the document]
milesgranger: Yepsi pepsi, would like to remove erroneous `#` from all numeric columns, and converting the string (not `Int64` at this point) to float covers both cases.

tests/workflows/test_from_csv_to_parquet.py
milesgranger: It's not selecting a single column, it's boolean indexing here.

tests/workflows/test_from_csv_to_parquet.py
milesgranger: https://github.com/coiled/benchmarks/blob/3062113a0a33e69c049379817ac7de49ef76004f/tests/conftest.py#L608

tests/workflows/test_from_csv_to_parquet.py
milesgranger: This would produce 'int64' and 'float64'  not 'Int64' and 'Float64' dtypes, and feel like this is maybe splitting hairs a bit at this point. `astype` is a very common operation and seems like, if anything, would be beneficial to have in the workflow anyhow.

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,145 @@
...
+):
+    s3 = s3_factory(anon=True)
+    files = s3.ls("s3://gdelt-open-data/events/")[:1000]
+    files = [f"s3://{f}" for f in files]
+
+    df = dd.read_csv(
+        files,
+        sep="\t",
+        names=SCHEMA.keys(),
+        # 'dtype' and 'converters' cannot overlap
```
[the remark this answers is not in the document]
milesgranger: This doesn't appear to be true; it fails, specifically the line treating `SOURCEURL` as a string, turns out to be `dtype('O')`, somewhat unexpected, no?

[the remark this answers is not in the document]
milesgranger: The above, depends on dask sampling to determine the column to be 'string-like' as according to the documentation, so it may have not interpreted that column as a string-like data from it's sample. In such a case the explicit dtypes avoids the ambiguity?

milesgranger:
Sorry, I was sort of under the impression that (the label) was added towards the end when not many more adjustments were required so as to not waste runs(?). I've been running it from local machine though.

j-bennet:
> Sorry, I was sort of under the impression that (the label) was added towards the end when not many more adjustments were required so as to not waste runs(?). I've been running it from local machine though.

Ah. :) In that case you're all set. Just making sure.

tests/workflows/test_from_csv_to_parquet.py
j-bennet: True.

tests/workflows/test_from_csv_to_parquet.py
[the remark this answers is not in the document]
j-bennet: Th

> This doesn't appear to be true; it fails, specifically the line treating `SOURCEURL` as a string, turns out to be `dtype('O')`, somewhat unexpected, no?

Quite unexpected. This option should convert `object` dtypes to `string[pyarrow]`. It doesn't look at data, only at dtypes. Makes me think that the setting somehow didn't apply.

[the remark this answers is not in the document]
milesgranger: I suppose this is happening because the environments `coiled-runtime-0-2-1-py310.yml` ends up with dask==2022.11.1 which I suppose doesn't include this configuration setting. Is that correct?

milesgranger:
The [failing test is unrelated](https://github.com/coiled/benchmarks/actions/runs/4978373648/jobs/8908626137?pr=841) it seems.

tests/workflows/test_from_csv_to_parquet.py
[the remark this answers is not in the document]
j-bennet: Yes, the setting is available since `2023.2.1`. But if it's not available, wouldn't you get an "unknown setting" error?

j-bennet (approved):
LGTM, but before you merge, I'd still like to resolve the `dataframe.convert-string` issue. If it's getting applied, you should not need to explicitly convert object columns to `string[pyarrow]`. If it's not getting applied, take it out, it's misleading.

tests/workflows/test_from_csv_to_parquet.py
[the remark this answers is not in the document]
milesgranger: Doesn't appear so:
```python
In [1]: import dask

In [2]: with dask.config.set({'foo.bar': True}):
   ...:     pass
   ...:

In [3]:
```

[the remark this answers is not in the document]
j-bennet: Hah. TIL. Well, in that case I suggest taking `configure_use_pyarrow_strings` out. It's useless anyway.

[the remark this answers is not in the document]
milesgranger: Okay dokay! https://github.com/coiled/benchmarks/pull/841/commits/19606da825f0c2d1c5ccc3d8fcd7be6795f676e3

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,139 @@
...
+    from_csv_to_parquet_cluster,
+    cluster_kwargs,
+    upload_cluster_dump,
+    benchmark_all,
+):
+    n_workers = cluster_kwargs["from_csv_to_parquet_cluster"]["n_workers"]
+    with Client(from_csv_to_parquet_cluster) as client:
+        from_csv_to_parquet_cluster.scale(n_workers)
+        client.wait_for_workers(n_workers)
+        client.restart()
```
fjetter, line 36: This restart should not be necessary considering that this is a cluster we only use once.

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,139 @@
...
+    ) as cluster:
+        yield cluster
+
+
+@pytest.fixture
+def from_csv_to_parquet_client(
+    from_csv_to_parquet_cluster,
+    cluster_kwargs,
+    upload_cluster_dump,
+    benchmark_all,
```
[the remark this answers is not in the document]
milesgranger: Same comment above, will adjust. Thanks.

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,139 @@
...
+
+    df = df.map_partitions(
+        lambda xdf: xdf.drop_duplicates(subset=["SOURCEURL"], keep="first")
+    )
+    df["national_paper"] = df.SOURCEURL.str.contains(
+        "washingtonpost|nytimes", regex=True
+    )
+    df = df[df["national_paper"]]
+    df = df.persist()
+    assert len(df)
```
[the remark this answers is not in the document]
milesgranger: Okay, it was recommended to approach this in a naive way I think, and asserting there was a length in my mind was ensuring we weren't accidentally not processing anything. I'd be happy to refactor it to be more efficient if that's the type of workflow we want to benchmark for.

phofl (commented):
small comment

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,139 @@
...
+            yield client
+
+
+SCHEMA = OrderedDict(
+    [
+        ("GlobalEventID", "Int64"),
+        ("Day", "Int64"),
+        ("MonthYear", "Int64"),
+        ("Year", "Int64"),
+        ("FractionDate", "Float64"),
```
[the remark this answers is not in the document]
milesgranger: No reason other than that's what was used in the previous PR.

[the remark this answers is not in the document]
milesgranger: Is it broken enough that we ought to change it for this workflow then have an issue to update it pinning some upstream issue?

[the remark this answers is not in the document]
phofl: https://github.com/pandas-dev/pandas/issues/32265

This one is the most relevant, but there are a bunch more. High level the issue is: You can have np.nan and pd.NA, but functions like fillna don't capture pd.NA, so not really usable. At least we shouldn't "recommend" it.

tests/workflows/test_from_csv_to_parquet.py
```diff
@@ -0,0 +1,102 @@
...
+
+@pytest.mark.client("from_csv_to_parquet")
+def test_from_csv_to_parquet(client, s3_factory, s3_url):
+    s3 = s3_factory(anon=True)
+    files = s3.ls("s3://gdelt-open-data/events/")[:1000]
+    files = [f"s3://{f}" for f in files]
+
+    df = dd.read_csv(
+        files,
+        sep="\t",
```
phofl, line 78: This makes read_csv fall back to the python engine. 2 things:

- if this is intended, we should be explicit
- the python engine is really slow. So if this is not intended we should choose a file that has a 1 character separator

milesgranger: I suppose it's quite likely the average user (including me) would not know this. 

Therefore, IMO, I'd think it ought to stay this way so when it changes in the future, the bump in performance would be evident historically because it's what the average user might expect.

The second option I think would be a slightly hacky work-around that the benchmarks aren't designed for. My understanding is these workflows should represent somewhat realistic use cases and not as a means for maximizing performance itself. Could be mistaken though. cc @fjetter @jrbourbeau

tests/workflows/test_from_csv_to_parquet.py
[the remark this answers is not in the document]
milesgranger: Sorry, I'm still not sure I understand. It seems somewhat related to our other discussion on the purpose of these workflows, if they should be designed from start to maximizing performance vs what a typical user would do.

If not, are you advocating we use "Int64" but _not_ "Float64"? If so, should I reference the issue you listed above and make a new issue in this repo that tracks the closing of that issue (and others) which would imply we follow up and switch this back to "Float64"?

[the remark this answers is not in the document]
phofl: Personally, I would be comfortable with using Int64  but not Float64, yes. This isn't really related to performance. Yes referencing the issue sounds good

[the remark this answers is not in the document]
milesgranger: Done and done, thanks for the clarification. :) 
https://github.com/coiled/benchmarks/pull/841/commits/530d35c7c8c2fec080cb91abbbe2a316fe3726d7

tests/workflows/test_from_csv_to_parquet.py
hendrikmakait: In a follow-up ticket, we could parametrize this test with comma-separated and tab-separated files if we care about the performance of the different engines.

PR #848, opened 2023-05-22
Polish PyTorch test

crusaderky:
- Allow test suite to run with a minimal A/B environment
- Don't pollute the environment used by all test with a hard to spot dependency tree introduced in `environment.yml`
- Test timeout
- Add notes on various hacks

### DIFF
--- a/tests/workflows/test_pytorch_optuna.py
+++ b/tests/workflows/test_pytorch_optuna.py
@@ -241,6 +241,6 @@ def test_hpo(client):
         client.submit(study.optimize, objective, n_trials=1, pure=False)
         for _ in range(n_trials)
     ]
-    await wait(futures, client, 300)
+    wait(futures, client, 600)
 
     assert len(study.trials) == n_trials
--- a/tests/workflows/test_pytorch_optuna.py
+++ b/tests/workflows/test_pytorch_optuna.py
@@ -15,11 +15,11 @@ optuna = pytest.importorskip("optuna")
     "pytorch_optuna",
     worker_plugin=PipInstall(
         [
-            # FIXME package_sync doesn't seem to deduce GPU specific installs of
+            # FIXME https://github.com/coiled/platform/issues/1249
+            #       package_sync doesn't seem to deduce GPU specific installs of
             #       libraries like torch
             "torch==2.0.0",
-            # FIXME https://github.com/coiled/benchmarks/pull/787#issuecomment-1532882045
-            #       Windows package_sync doesn't like torchvision
+            # FIXME Windows package_sync doesn't like torchvision
             "torchvision==0.15.1",
             # FIXME https://github.com/boto/botocore/issues/2926
             #       urllib3 v2 removed openssl / ciphers which causes an error in
--- a/tests/workflows/test_pytorch_optuna.py
+++ b/tests/workflows/test_pytorch_optuna.py
@@ -241,6 +241,6 @@ def test_hpo(client):
         client.submit(study.optimize, objective, n_trials=1, pure=False)
         for _ in range(n_trials)
     ]
-    wait(futures, client, 600)
+    wait(futures, client, 900)
 
     assert len(study.trials) == n_trials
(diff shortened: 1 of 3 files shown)


tests/workflows/test_pytorch_optuna.py
```diff
@@ -3,25 +3,30 @@
...
+
+optuna = pytest.importorskip("optuna")
 
 
 @pytest.mark.client(
     "pytorch_optuna",
     worker_plugin=PipInstall(
         [
+            # FIXME package_sync doesn't seem to deduce GPU specific installs of
+            #       libraries like torch
```
crusaderky, line 19: @milesgranger @ntabris @shughes-uk is there a coiled issue for this we can reference here?

tests/workflows/test_pytorch_optuna.py
```diff
@@ -3,25 +3,30 @@
...
             "torch==2.0.0",
-            "torchvision",
-            # FIXME: https://github.com/boto/botocore/issues/2926
-            # urllib3 v2 removed openssl / ciphers which causes an error in botocore httpsession
+            # FIXME https://github.com/coiled/benchmarks/pull/787#issuecomment-1532882045
+            #       Windows package_sync doesn't like torchvision
```
crusaderky, line 22: idem; coiled issue?

tests/workflows/test_pytorch_optuna.py
```diff
@@ -3,25 +3,30 @@
...
+            "torchvision==0.15.1",
+            # FIXME https://github.com/boto/botocore/issues/2926
+            #       urllib3 v2 removed openssl / ciphers which causes an error in
+            #       botocore httpsession
             "urllib3<2.0.0",
+            # FIXME https://github.com/optuna/optuna/pull/4589
+            #       Need optuna >3.1.1
```
crusaderky, line 29: @milesgranger here you are installing a version of optuna on the client and a different one on the workers. I guess that the client-side part is luckily compatible with the worker-side one?

tests/workflows/test_pytorch_optuna.py
milesgranger: https://github.com/coiled/platform/issues/1249

tests/workflows/test_pytorch_optuna.py
milesgranger: At the time it was the same IIRC, this change was due to https://github.com/coiled/benchmarks/pull/787/files#r1196779222

tests/workflows/test_pytorch_optuna.py
milesgranger: I suppose it may fall under the same (https://github.com/coiled/platform/issues/1249) albeit since this from CI/Windows specific maybe there is or worthy of a separate one. @ntabris

tests/workflows/test_pytorch_optuna.py
```diff
@@ -34,6 +39,8 @@ def test_hpo(client):
     NZ = 100  # Size of z latent vector (i.e. size of generator input)
 
     def weights_init(m):
+        import torch.nn as nn
```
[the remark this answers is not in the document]
crusaderky: Because pytorch is no longer installed on the client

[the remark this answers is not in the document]
jrbourbeau: Maybe we should add it to the environment for the client too. Otherwise it will make setup for running this test locally more difficult.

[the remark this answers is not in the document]
crusaderky: The thing is - package_sync doesn't work with it; see FIXMEs. So I feel that having those packages on the client would give the false impression that pytorch works like all other libraries, which is false.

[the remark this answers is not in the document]
milesgranger: [Wish logs would stay around longer (or I should have copied the error)](https://github.com/coiled/benchmarks/pull/787#issuecomment-1532882045) but that is how I remember it, yes.

crusaderky:
@milesgranger this is ready for review and merge

milesgranger (approved):
Thank you.

PR #849, opened 2023-05-22
Prometheus is on by default, remove old setting to enable

ntabris:


jrbourbeau (approved):
Nice, thanks @ntabris 

For other readers, just wanted to note that `prometheus_client` is a dependency of `coiled` now, so it will always be installed

PR #857, opened 2023-05-26
Clean up optuna and snowflake imports

crusaderky:


hendrikmakait (approved):
Thanks, @crusaderky!

AB_environments/AB_baseline.conda.yaml
```diff
@@ -34,22 +36,27 @@ dependencies:
     - toolz ==0.12.0
     - zict ==3.0.0
     - xgboost ==1.7.4
-    - dask-ml ==2023.3.24
-    - optuna ==3.1.0
+    - optuna ==3.1.1  # FIXME Actually need >3.1.1; see git override below
     - scipy ==1.10.1
-    - sqlalchemy ==2.0.12
+    - snowflake-connector-python ==3.0.4
+    - snowflake-sqlalchemy ==1.4.7
```
jrbourbeau, line 42: This seems like overkill. Do we need to pin these `dask-snowflake` dependencies? (my guess is "yes" and that I'm just missing something )

crusaderky: I would not pin these dependencies normally. But they introduce pins on other stuff (see comments in the file) so I got the impression that they can be quite delicate.

PR #865, opened 2023-05-30
We should not run snowflake CI job unless certain conditions apply

j-bennet:
See https://github.com/coiled/benchmarks/issues/864.


hendrikmakait:
It looks like https://github.com/coiled/benchmarks/pull/857 might supersede this one.

j-bennet:
> It looks like https://github.com/coiled/benchmarks/pull/857 might supersede this one.

Thank you. Closing.

PR #903, opened 2023-07-05
Test overhead of large amounts of non-rootish tasks

crusaderky:
After conversation on https://github.com/dask/distributed/pull/7961
CC @fjetter 

image
image


fjetter (approved):
This looks really bad because of the regression you already fixed or is there something else going on?

Thanks for doing this!

fjetter:
Ah right, there is a piling up on the one worker happening. I'll need to open a (public) issue about this

PR #911, opened 2023-07-21
Simple deltalake benchmark

j-bennet:
Compare `read_deltalake` vs `read_parquet`.


j-bennet:
Right now, this fails because deltalake needs explicit credentials to access s3 bucket - `anon: true` doesn't work. Issue here: https://github.com/delta-io/delta-rs/issues/1554.

PR #914, opened 2023-07-25
[DO NOT MERGE] see if this fixes the regression

j-bennet:
See if this fixes the regression in `make_timeseries` https://github.com/coiled/benchmarks/issues/910.

Xref https://github.com/dask/dask/pull/10428.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -75,6 +75,11 @@ jobs:
       - name: Add test dependencies
         run: mamba env update --file ci/environment-test.yml
 
+      - name: Install Dask from branch, no dependencies
+        run: |
+          pip uninstall -y dask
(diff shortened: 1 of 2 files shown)


j-bennet:
Looks like this didn't help. :(

image

Might be best to revert https://github.com/dask/dask/pull/10392, and have another version that doesn't touch `make_timeseries` at all.

phofl:
Hm that's unfortunate. Maybe try dask pyspy?

j-bennet:
> Hm that's unfortunate. Maybe try dask pyspy?

I have, but with the multiple threads, the picture is still pretty complicated. I'll try simplifying it + limiting threads.

j-bennet:
Yes! It did it.

image

PR #915, opened 2023-07-25
[DNM] Try two worker processes per host

ntabris:
I'd like to see how this impacts perf of various benchmarks.

### DIFF
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -15,6 +15,7 @@ default:
   package_sync: true
   wait_for_workers: true
   scheduler_vm_types: [m6i.large]
+  _n_worker_specs_per_host: 2
   backend_options:
     spot: true
     spot_on_demand_fallback: true


cluster_kwargs.yaml
```diff
@@ -15,6 +15,7 @@ default:
   package_sync: true
   wait_for_workers: true
   scheduler_vm_types: [m6i.large]
+  _n_worker_specs_per_host: 2
```
fjetter, line 18: is this on prod now?

fjetter: IIUC a _fair_ assessment would require us to double the machine sizes and reduce cluster sizes by half. Otherwise the same-host workers have much less memory to work with and are much more likely to run into spilling and OOM

fjetter: I kicked off an A/B test with this https://github.com/coiled/benchmarks/actions/runs/5667683049

fjetter: Results of this A/B test are interesting but likely require a bit further analysis

image

We can see a couple of tests that regress significantly while others are getting much worse.

My best read without digging deeper is that most/almost all of our tests require a bit of spilling and that all test cases that are spilling behave really poorly, likely because disk is just busy?

ntabris:
@fjetter you want to try run with 1/2 the number of `m6i.xlarge` machines? want me to push that change to this branch?

fjetter:
@ntabris https://github.com/coiled/benchmarks/pull/917 is what I ran for the A/B tests. Clusters were using only half the number of VMs, VMs were twice as large, i.e. same total number of CPUs, same total RAM, half the number of VMs with two workers each

PR #926, opened 2023-08-04
[DNM] Try m7i instead of m6i

ntabris:
Let's see if they're faster!

> They deliver up to 19% better price performance compared to M6i instances

https://aws.amazon.com/ec2/instance-types/m7i/

### DIFF
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -14,7 +14,7 @@
 default:
   package_sync: true
   wait_for_workers: true
-  scheduler_vm_types: [m6i.large]
+  scheduler_vm_types: [m7i.large]
   backend_options:
     spot: true
     spot_on_demand_fallback: true
@@ -23,7 +23,7 @@ default:
 # For all tests using the small_client fixture
 small_cluster:
   n_workers: 10
-  worker_vm_types: [m6i.large]  # 2CPU, 8GiB
+  worker_vm_types: [m7i.large]  # 2CPU, 8GiB
 
 # For tests/benchmarks/test_parquet.py
 parquet_cluster:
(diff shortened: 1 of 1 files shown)


mrocklin:
Ooh, exciting.  Blog post material?

ntabris:
So far looks like it does improve runtimes...

[previous m6i array tests](https://cloud.coiled.io/clusters/251623/information?account=dask-benchmarks) vs [m7i](https://cloud.coiled.io/clusters/251966/information?account=dask-benchmarks)

[previous m6i dataframe tests](https://cloud.coiled.io/clusters/251274/information?account=dask-benchmarks) vs [m7i](https://cloud.coiled.io/clusters/251970/information?account=dask-benchmarks)

scharlottej13:
> So far looks like it does improve runtimes...
> 
> [previous m6i array tests](https://cloud.coiled.io/clusters/251623/information?account=dask-benchmarks) vs [m7i](https://cloud.coiled.io/clusters/251966/information?account=dask-benchmarks)
> 
> [previous m6i dataframe tests](https://cloud.coiled.io/clusters/251274/information?account=dask-benchmarks) vs [m7i](https://cloud.coiled.io/clusters/251970/information?account=dask-benchmarks)

Yeah! And enough faster to make it cheaper (according to the _very_ approximate cloud cost column).

Do you know why network rate isn't showing up for the m7i tests?

ntabris:
Looks like it was 89 cpu-hours:

image

I think typical run is ~190 cpu-hours:

image

https://cloud.coiled.io/settings/billing?account=dask-benchmarks&groupby=tag%3AGITHUB_RUN_ID&start=1690934400&end=1691282414

Final point in the plots is for this run, and I'm seeing a lot of plots that look something like this:

image

Or this:

image

Or this:

image

ntabris:
> Do you know why network rate isn't showing up for the m7i tests?

Not sure, I'll look next week. My guess is that network device has a different name and I'll need to tweak the config to capture it.

ntabris:
Comparing cost, it looks like `m7i` run was fewer cpu-hours but had higher cost of $1.76 vs $1.33 for `m6i` run:

image

Part of the explanation is that we got all spot for `m6i` but had to use some on-demand for `m7i`:

image

`m7i.large` spot price is about **27%** higher than `m6i.large` currently: $0.0415/hr vs $0.0325/hr. This is because there's 5% higher base price and less of a discount currently on spot.

# Upshot

When making heavy use of spot (which we are in benchmarks), `m6i` still gives the best overall price.

If you were already using mostly on-demand instances, `m7i` can (in workloads sufficiently similar to our benchmarks) deliver enough performance increase to justify the slight cost increase.

fjetter:
I see you are using the flex instances. do you know what's the difference to the normal ones? I can't seem to find anything in the description

ntabris:
"flex" is interesting. As far as I can tell, it's like burstable in that you don't have guaranteed access to full CPU, but unlike burstable in that you seem to get better sustained CPU util without paying a huge amount for extra "cpu credits".

Total time was slightly worse than `m7i.large` and better than `m6i.large`, total cost was also in between:

image

As with `m7i.large`, this looks like a good option if you're paying on-demand prices, but spot discount isn't as good so total cost is still better with `m6i`.

Total cost:
- `m7i`: $1.76
- `m7i-flex`: $1.57
- `m6i`: $1.33

PR #933, opened 2023-08-07
QoL dashboard ranges

fjetter:
These are a couple of QoL changes to how we're setting ranges

- Default view only selects 30days. All data is still there if scrolled
- Set y domain to mean +- stddev
- Scrolling now locks in Y axis. It happens very rarely that we're interested in changing y axis
- Make dots smaller
- Make legend less verbose by not spelling out the test name again, just the parameter

This is still far from ideal but I find it a little easier to work with.

Note: The code is a bit of a mess and I didn't put in a lot of effort to make this very clean. I think `make_test_report` is a bit of an anti pattern since it just redirects to different implementations making the kwarg forwarding very awkward. Topic for another day

Before
image

After
image

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -292,7 +292,10 @@ def details_report_fname(runtime: str, fullname: str) -> str:
 
 
 def make_timeseries(
-    df: pandas.DataFrame, spec: ChartSpec, title: str
+    df: pandas.DataFrame,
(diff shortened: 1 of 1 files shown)


hendrikmakait (approved):
LGTM, thanks @fjetter!

PR #936, opened 2023-08-08
Upgrade `pandas` version to 2.0.3

phofl:
we will cut the rc for 2.1 this week, so lets test on 2.0.3 before that happens


jrbourbeau (approved):
Thanks @phofl

PR #937, opened 2023-08-08
Avoid removing shuffle in shuffle benchmark

phofl:
We are removing shuffles now when the length is computed on a DataFrame, so this test is a but pointless for dask_expr. That's the fastest op that should not remove the shuffle.


fjetter (approved):
I'm fine with this. However, I guess a more realistic version of this would be

`df.groupby("0").apply(lambda x:x)` which is effectively the same but realistically, a user will rather call groupby.apply than shuffle directly

PR #938, opened 2023-08-08
Unskip large_map_first_work

fjetter:
Closes https://github.com/coiled/benchmarks/issues/521

The restart deadlock has not been fully resolved but I believe the benchmarking suite is no longer susceptible to it.


jrbourbeau (commented):
Thanks @fjetter. This looks fine to me, but I'm not sure what's going on with the dashboard build. I restarted that CI job, and it's consistently failing

PR #971, opened 2023-09-01
Add tpch benchmark queries

phofl:


tests/benchmarks/test_tpch.py
```diff
@@ -0,0 +1,315 @@
...
+            "sum_disc_price": "sum",
+            "sum_charge": "sum",
+            "avg_qty": "mean",
+            "avg_price": "mean",
+            "avg_disc": "mean",
+            "count_order": "count",
+        }
+    )
+
+    total.reset_index().compute().sort_values(["l_returnflag", "l_linestatus"])
```
fjetter, line 59: nit: feels a little like cheating to compute first before sorting. In a perfect world, dask would know the data is small enough to collect and sort locally

tests/benchmarks/test_tpch.py
```diff
@@ -0,0 +1,315 @@
...
+            "s_acctbal",
+            "s_name",
+            "n_name",
+            "p_partkey",
+            "p_mfgr",
+            "s_address",
+            "s_phone",
+            "s_comment",
+        ]
+    ].persist().sort_values("s_acctbal", ascending=False).head(100)
```
fjetter, line 113: why the persist? 

I wonder how common smth like this is. That could be easily improved with an optimized topk-like operation

tests/benchmarks/test_tpch.py
```diff
@@ -0,0 +1,315 @@
...
+    total2 = total2.drop(columns=["o_orderkey", "l_orderkey"])
+
+    # concat results
+    total = dd.concat([total1, total2])
+    result_df = (
+        total.groupby(["supp_nation", "cust_nation", "l_year"])
+        .revenue.agg("sum")
+        .reset_index()
+    )
+    result_df = result_df.compute()
```
fjetter, line 305: again, a premature compute that feels like cheating.

tests/benchmarks/test_tpch.py
```diff
@@ -0,0 +1,315 @@
...
+    path = DATASETS[enabled_dataset] + filename + "/"
+    return dd.read_parquet(path, engine="pyarrow")
+
+
+@pytest.mark.client("tpch")
+def test_query_1(client):
+    VAR1 = datetime(1998, 9, 2)
+    lineitem_ds = read_data("lineitem")
+
+    lineitem_filtered = lineitem_ds[lineitem_ds.l_shipdate <= VAR1]
```
fjetter, line 29: I think this is a little bit biased towards dask-expr. Ordinarily, we'd point users to using filters in the read_parquet call.
That's fine, I think, just wanting to point it out

tests/benchmarks/test_tpch.py
fjetter, line 113: The TPCH query 2 defines the sorting on multiple columns (see https://www.tpc.org/TPC_Documents_Current_Versions/pdf/TPC-H_v3.0.1.pdf)

image

tests/benchmarks/test_tpch.py
```diff
@@ -0,0 +1,315 @@
...
+    fcustomer = cutomer_ds[csel]
+    jn1 = fcustomer.merge(forders, left_on="c_custkey", right_on="o_custkey")
+    jn2 = jn1.merge(flineitem, left_on="o_orderkey", right_on="l_orderkey")
+    jn2["revenue"] = jn2.l_extendedprice * (1 - jn2.l_discount)
+    total = jn2.groupby(["l_orderkey", "o_orderdate", "o_shippriority"])[
+        "revenue"
+    ].sum()
+    total.reset_index().head(10)[
+        ["l_orderkey", "revenue", "o_orderdate", "o_shippriority"]
+    ]
```
fjetter, line 139: I believe this is missing an orderby

tests/benchmarks/test_tpch.py
phofl: That's a good idea, I'll add this to my todo list for dask-expr. I'd like to avoid measuring the intermediate compute though, that's why I added it in this order.

tests/benchmarks/test_tpch.py
mrocklin: > That could be easily improved with an optimized topk-like operation

sort+head -> TopK seems like a classic optimization we should probably have at some point.  Should be easy to write too (if we have nlargest around already).  If we wanted to onboard someone onto the project this might be an interesting good first issue (for an experienced dask dev)

tests/benchmarks/test_tpch.py
mrocklin: > Ordinarily, we'd point users to using filters in the read_parquet call.

I'm pretty comfortable with what's here.  I rarely see users use the read_parquet kwargs in practice, even though they should.  

More generally, I think that we should write things the way we think a naive users would write them, rather than the way they should write them.

tests/benchmarks/test_tpch.py
mrocklin: Also, just to convey optimization thoughts, this is why we do optimizations top-down.  We would not want to touch the sort/shuffle/p2p lowering phase at all until the expressions above it (the head) have an opportunity to make all of that obsolete.

tests/benchmarks/test_tpch.py
fjetter: sorry, to be more explicit: dask should not sort _locally_ but reduce to a single worker, sort there and return your result

tests/benchmarks/test_tpch.py
fjetter: > sort+head -> TopK seems like a classic optimization we should probably have at some point.

apart from the optimization, I think we currently don't even have an API for this. We have topk for arrays. That's good. But in this case we want to return entire DF rows based on the TopK. That's not harder to implement, it's just not there, I think.

My question was not targeted for the optimization but rather whether we have (or should offer) this API

mrocklin: Maybe you're looking for https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.nlargest.html ?

fjetter: ahhh, thanks 🤦

tests/benchmarks/test_tpch.py
phofl: Actually, I don't think they should. I've done some measurements with filters in read_parquet and they are terribly slow if you don't have metadata available for the columns you are filtering. 

I've explicitly avoided pushing the filters into read_parquet in my benchmarks because that was at least 3 times as slow as doing this with dask/pandas

fjetter: Well, that's sad and we should investigate. If this is indeed slower, `read_parquet` is trying to be smart where it shouldn't be or smth is just implemented poorly.

tests/benchmarks/test_tpch.py
phofl: Yep thx

tests/benchmarks/test_tpch.py
phofl: The slowdown is in arrow itself. We aren't doing anything fancy except passing the filters in.

phofl:
ok to merge?

PR #980, opened 2023-09-07
Collect prometheus metrics

fjetter:
This collects a couple of prometheus metrics from coiled after the run to expand our measurements to scheduler metrics (CPU + memory) and event loop health metrics (scheduler and workers)

cc @ntabris for a sanity check of the queries

cc @hendrikmakait @crusaderky @phofl


tests/conftest.py
```diff
@@ -231,6 +231,62 @@ def _benchmark_time():
...
+                start_ts=start,
+                end_ts=end,
+            )
+            test_run_benchmark.scheduler_cpu_avg = cluster.get_aggregated_metric(
+                query="cpu&scheduler",
+                over_time="avg",
+                start_ts=start,
+                end_ts=end,
+            )
+            # TODO: Do we want to have plain max or 80th percentile?
```
fjetter, line 273: I'm not entirely sure what we're interested in, yet. My sense is that we do not want to be alarmed if there is a single peak (e.g. single gc.collect takes longer) but if a high quantile goes up we should be concerened

fjetter:
A follow up is obviously to integrate this to the dashboards

tests/conftest.py
```diff
@@ -231,6 +232,62 @@ def _benchmark_time():
...
+    def _benchmark_coiled_prometheus(client):
+        if not test_run_benchmark:
+            yield
+        else:
+            start = math.floor(time.time())
+            yield
+            end = math.ceil(time.time())
+            cluster = client.cluster
+            test_run_benchmark.scheduler_memory_max = cluster.get_aggregated_metric(
+                query="(host_memory_total-host_memory_available)&scheduler",
```
ntabris, line 263:
```suggestion
                query="scheduler:(host_memory_total-host_memory_available)&scheduler",
```

`&scheduler` is a filter (like `&workers-up-30s`), `scheduler:` says that we want scheduler rather than worker metrics.

Example at https://github.com/coiled/platform/blob/e0913930dd7f04e618027176d28dbf9a4b9b394e/declarative/services/prometheus/tags.py#L38

tests/conftest.py
```diff
@@ -231,6 +232,62 @@ def _benchmark_time():
...
+            )
+            test_run_benchmark.scheduler_cpu_avg = cluster.get_aggregated_metric(
+                query="cpu_rate&scheduler",
+                over_time="avg",
+                start_ts=start,
+                end_ts=end,
+            )
+            # TODO: Do we want to have plain max or 80th percentile?
+            test_run_benchmark.worker_max_tick = cluster.get_aggregated_metric(
+                query="worker_max_tick|pct80",
```
ntabris, line 276: just to be clear, `pct80` here means that at each time you ignore the worst 20% of workers.

aggregations in `query` are applied at each time, aggregations in `over_time` are then (next step) applied to the values over time.

ntabris: if you want the worst worker, but ignoring the worst 20% of the time, then you'd have

```
query="worker_max_tick|max",
over_time="quantile(.8)",
```

tests/conftest.py
```diff
@@ -231,6 +232,62 @@ def _benchmark_time():
...
+                over_time="max",
+                start_ts=start,
+                end_ts=end,
+            )
+            test_run_benchmark.scheduler_max_tick = cluster.get_aggregated_metric(
+                query="scheduler_max_tick|pct80",
```
ntabris, line 282: so `pct80` doesn't do much here, just one value at any given time

tests/conftest.py
fjetter: So I don't need the suffix `&scheduler` anymore, do I?

ntabris: I'd leave it in, it might not be needed in some specific cases (not sure about this one), but it's a way of addressing https://github.com/coiled/platform/issues/1358 in the cases where that applies.

tests/conftest.py
crusaderky: 95% quantile on some very short tests seems very aggressive to me; 80% sounds more reasonable IMHO

dashboard.py
```diff
@@ -34,9 +34,13 @@ class ChartSpec(NamedTuple):
 
 
 SPECS = [
-    ChartSpec("duration", "Wall Clock", "(s)", 1),
-    ChartSpec("average_memory", "Average Memory", "(GiB)", 2**30),
-    ChartSpec("peak_memory", "Peak Memory", "(GiB)", 2**30),
+    ChartSpec("duration", "Wall Clock", "[s]", 1),
+    ChartSpec("average_memory", "Average Memory", "[GiB]", 2**30),
+    ChartSpec("peak_memory", "Peak Memory", "[GiB]", 2**30),
```
crusaderky, line 39:
```suggestion
    ChartSpec("average_memory", "Average Memory (W)", "[GiB]", 2**30),
    ChartSpec("peak_memory", "Peak Memory (W)", "[GiB]", 2**30),
```

crusaderky (changes requested):
Sorry about the delayed review.
There are a few issues:

1. it seems like the new metrics are being collected exclusively for `tpch/test_dask.py`:

image

2. avg CPU seems to be 0.55% for query1, which is suspiciously low, and is exactly zero for the other measures:
image

If I look at grafana for query7, I read 8~15%:
image

3. max tick (both scheduler and worker) according to the measures are in the nanoseconds, which can't be right:
image

fjetter:
> it seems like the new metrics are being collected exclusively for tpch/test_dask.py:

see https://github.com/coiled/benchmarks/pull/1101

fjetter:
Regarding the data itself, I guess we'll have to wait for @ntabris I'm having a little bit of a hard time with the query syntax

fjetter:
I had another look at the queries and responding to @crusaderky comments above in https://github.com/coiled/benchmarks/pull/980#pullrequestreview-1680827281

1. We were only running TPCH. Has been fixed a long time ago
2. CPU query was wrong and this was basically showing randomly the different CPU types (see https://github.com/coiled/platform/issues/3665)
3. The tick duration returned by coiled is in seconds and the normalization is wrong. I edited it now to show in `ms` which is the _desired_ scale although realistically we'll probably still see seconds. I don't have a strong preference but the chart should now show correct numbers

crusaderky:
> 3. The tick duration returned by coiled is in seconds and the normalization is wrong. I edited it now to show in `ms` which is the _desired_ scale although realistically we'll probably still see seconds. I don't have a strong preference but the chart should now show correct numbers

I don't have a strong preference either as long as the numbers match the visualized scale :)

fjetter:
Yeah, I think I like those metrics. The scheduler CPU charts, for instance already point to a couple of scheduler CPU bound workloads

image

which I can indeed confirm are using way too much scheduler CPU

image

fjetter:
There also seems to be something off with the  read_hive_partitions test. The event loop on the worker is blocked for more than a minute

image

fjetter:
With this we could also consider dropping the memory sampler but I suggest to do so in a follow up

crusaderky (approved):

crusaderky:
> There also seems to be something off with the read_hive_partitions test. The event loop on the worker is blocked for more than a minute

I'm not seeing anything remotely like that in the latest CI run. I'm reading 20ms.
Could it be an instance that just didn't receive any CPU time from the hypervisor for one minute? @ntabris

ntabris:
> Could it be an instance that just didn't receive any CPU time from the hypervisor for one minute? 

That sounds extremely unlikely.

PR #991, opened 2023-09-15
[DNM] Test worker plugin fix

fjetter:
This should confirm that https://github.com/dask/distributed/pull/8189 restores functionality

Closes https://github.com/coiled/benchmarks/issues/986


fjetter:
The failing builds are due to a too small version of pyarrow https://github.com/coiled/benchmarks/issues/990

I'll push another commit that should reflect this

PR #1012, opened 2023-09-22
Bump all dependencies in conda recipe

crusaderky:
General upgrade of all dependencies.
A/B test [https://github.com/coiled/benchmarks/actions/runs/6262315429] highlights a reduction in runtime in several H2O tests (q1-q7) and no regression anywhere.

- Blocked by and incorporates #1004 
download
download (1)


phofl (approved):
good to merge when CI is through, the snowflake connector pinning pandas is very unfortunate

PR #1019, opened 2023-09-25
[DNM] Use (feature-flagged) AWS creds endpoint

ntabris:
I don't expect any issues, but this is easy way to confirm that this works with all the ways we rely on AWS creds here.


ntabris:
just remember that I probably need to wait for [conda-forge pr](https://github.com/conda-forge/coiled-feedstock/pull/236) before running this

PR #1023, opened 2023-09-26
Centralize client fixture

crusaderky:
Replace small_client and parquet_client with the generic 'client' fixture already used by workflows.

I left spill_client there for now as it has the additional complication of having custom config.


crusaderky:
Abandoning this, as it's nowhere near working and the cost/benefit ratio is way too low.

The original benefit, to have a single point that starts a cluster which can be polished and filled with hacks and workarounds, is much less important than before thanks to upstream changes in coiled.

PR #1034, opened 2023-09-29
Faster skip for H2O tests

fjetter:
The current implementation generates a test case for all possible defined Datasets. This is not necessary since we can directly just generate those that are enabled.

Either way, if we decide to skip, the skip should be implemented in a fixture further up the DAG. By having `ddf` depend on `small_client`, it will execute `small_client` every time, i.e. it will connect a new client and restart the cluster just to enter `ddf` and decide to skip.
Considering that 6 out of 9 datasets are skipped and there are 9 test cases, this gives us 54 unnecessary client connections + 2 restarts.

This fixes both...

### DIFF
--- a/tests/benchmarks/test_h2o.py
+++ b/tests/benchmarks/test_h2o.py
@@ -37,11 +37,13 @@ else:
     }
 
 
-@pytest.fixture(params=list(DATASETS))
-def ddf(request, small_client):
-    if request.param not in enabled_datasets:
-        raise pytest.skip("Disabled by default config or H2O_DATASETS env variable")
+@pytest.mark.fixture(autouse=True)
+def client(small_client):
+    yield small_client
 
+
+@pytest.fixture(params=list(enabled_datasets), scope="module")
+def ddf(request):
     n_gib = float(request.param.split(" GB ")[0])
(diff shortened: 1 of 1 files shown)


tests/benchmarks/test_h2o.py
```diff
@@ -37,11 +37,13 @@
...
-@pytest.fixture(params=list(DATASETS))
-def ddf(request, small_client):
-    if request.param not in enabled_datasets:
-        raise pytest.skip("Disabled by default config or H2O_DATASETS env variable")
+@pytest.mark.fixture(autouse=True)
+def client(small_client):
+    yield small_client
 
+
+@pytest.fixture(params=list(enabled_datasets), scope="module")
```
crusaderky, line 45: This causes the order of the tests to become pseudo-random - which from memory is something that pytest has problems with.

Could you change both definitions of `enabled_datasets` above to be lists?

fjetter: I went with a sorted instead since `DATASETS` is a dict not a set

crusaderky: > I don't mind changing it to lists but how is my version any different to what was here before?

previously it was iterating on dict keys, which retain insertion order

PR #1044, opened 2023-10-03
Tpch: Dask vs PySpark and PySpark, Polars and DuckDB single node

milesgranger:
Merged #971  and #1027 together and added tests comparing each per tpch query; right now the first 7 queries as implemented in #971. 

First run: https://cloud.coiled.io/clusters/284888/information?account=dask-engineering&tab=Metrics
Note that dask runs first, then pyspark for the same query. 
One can pretty easily determine start/stop of each based on tasks being present in the first chart.

Initial impression is PySpark spills to disk less / uses less memory, and Dask is generally faster.

Dashboard artifact:
https://github.com/coiled/benchmarks/suites/16815018862/artifacts/960398346


milesgranger:
Ran again with 20 nodes:
https://cloud.coiled.io/clusters/284918/information?account=dask-engineering&tab=Metrics

Dashboard link:
https://github.com/coiled/benchmarks/suites/16817723847/artifacts/960516668

---

For 8 nodes:
https://cloud.coiled.io/clusters/284888/information?account=dask-engineering&tab=Metrics

Dashboard link:
https://github.com/coiled/benchmarks/suites/16815018862/artifacts/960398346

mrocklin:
> Initial impression is PySpark spills to disk less / uses less memory, and Dask is generally faster.

Surprising, but encouraging!  I'm intersted in digging in

> One can pretty easily determine start/stop of each based on tasks being present in the first chart.

I wonder if we can improve on this.  I'm not very familiar with how A/B tests work, but if we had two branches with the same test names like `test_tpch.py::test_query_1` but where one implementation was with Dask and the other was with Spark, then would the normal A/B test charts work?

Right now looking at the charts I think I'm seeing that Dask is a lot slower on more queries than Spark is.  It's a little hard to tell though (I'm looking at the x-axis range on charts like these).  Is that the right way to interpret these?

Screenshot 2023-10-03 at 7 21 32 AM

milesgranger:
> I think I'm seeing that Dask is a lot slower on more queries than Spark is.
Ah, you are indeed correct, I found from the 8 node cluster that dask was faster 3 out of the 7 queries.

mrocklin:
Also, while I'm asking for things.  I'll also be very curious how to changes at different scale levels.  Common wisdom today says "Dask is ok at medium scales, but for large scale you really have to use Spark".  I'm curious how true this is today.

milesgranger:
Have it running for tpch scale 1000:

Cluster: https://cloud.coiled.io/clusters/286647/information?account=dask-benchmarks&tab=Metrics&filterPattern=
Dashboard: https://github.com/coiled/benchmarks/suites/16906252166/artifacts/965862625d

---

What now?

I think keeping the parameterized dask/pyspark tests (`tests/benchmarks/test_tpch_pyspark.py::TestTpchDaskVsPySpark::test_query_1[dask/pyspark] `) is useful to both compare against pyspark and keep track of potential regressions in dask-expr (now that #1046 is in)

To do so we alter the tests in `test_tpch` and `test_tpch_pyspark` to only be entry points to tpch query implementations for dask/pyspark, then keep this PR's `TestTpchDaskVsPySpark` class tests which are parameterized by dask/pyspark for each tpch query. This would basically be just renaming those tests and keeping the parameterized comparison tests the same.

Then in a separate PR I can make proper A/B tests between the tpch query implementations to get those nicer graphs as requested earlier by @mrocklin 

cc @fjetter  @phofl

mrocklin:
> What now?

I think that this is worth a larger synchronous conversation, probably between @fjetter @milesgranger @phofl and I.  There are, I think, some good possible directions:

1. 10TB
2. Figure out why we're slower than Spark at large scale and fix it
3. Add Polars and DuckDB
4. Pull out Spark work here into a more solid thing that others can use reliably

phofl:
I would prefer to run spark and dask-expr on different clusters. Not sure if they might influence each other. this is a general thing, not only for A/B tests.

Next step is to figure out what spark is doing differently for the merge heavy queries

I would also like to see same queries on smaller clusters, I was running them on 50 workers and some of the queries are barely faster in this run with 75 workers (see https://github.com/coiled/benchmarks/actions/runs/6275077913)

tests/benchmarks/test_tpch_pyspark.py
```diff
@@ -0,0 +1,248 @@
...
+            spark = get_or_create_spark(f"query{name}")
+            module.setup(spark)  # read spark tables query will select from
+
+            if hasattr(module, "ddl"):
+                spark.sql(module.ddl)  # might create temp view
+
+            q_final = spark.sql(module.query)  # benchmark query
+            try:
+                # trigger materialization of df
+                _retry_s3_forbidden(lambda: print(q_final.show(10)))
```
fjetter, line 81: by just "showing ten rows", this could trigger yet another optimization step. Naively, I would assume that this just modifies the query to have something like a `LIMIT 10` which, depending on the query, can change the result substantially.

From this perspective, to ensure the benchmarks align, we should also make sure that the dask and spark queries return the same kind of data. For instance, if the dask cluster actually submits the results back to the client, the spark cluster should do the same.
If the result is just computed and forgotten (e.g. persist + wait in dask), we should do the equivalend in spark land. Otherwise, we'll have a systematic error due to the additional client<->scheduler transmission.

milesgranger: Very good point. 

In my opinion, especially if we're moving forward with benchmarking things like Polars and DuckDB, is they should at minimum compute something native to themselves, I think this is also your opinion. PySpark converting to pandas is not really encouraged, stranger still if we did that with Polars and DuckDB.

**Best approach** in my current understanding is to benchmark all in their ability to process the queries into parquet files. This is pretty unbiased, ensures each engine doesn't need to rely on some specific conversion implementation to some other engine, and means no magical lazy-non-computation occurs.

(Side note, after some experimentation, it doesn't appear `.show(...)` does anything fancy related to optimization)

fjetter: I'm not saying we should compare the results, just make sure that we're treating the final data transmission equally. I do not think we should convert anything to parquet since this just adds another IO layer that may be unreliable for the benchmark

mrocklin:
Looking at this I'm seeing some possible directions for future work (no particular order):

1.  Make sure that the queries are optimal (we'll probably have to engage some external reviewer for this)
2. Make sure that the Spark configuration is optimal (same as above)
3. Make sure that the queries roughly match what we do in the Dask side (we're doing this)
4. Extend to Polars and DuckDB (single large machine)
5. Increase scale
6. Pull out Spark stuff to platform lab.  See https://github.com/coiled/platform/issues/3478
    - Figure out Spark Connect
    - Clean up implementation

milesgranger:
https://github.com/coiled/benchmarks/pull/1044/commits/1dd1f948486a48e624facfd962804a6b87ad4a72 Has the initial implementation of duckdb/pyspark/dask running on a single VM.

Set up to read directly from S3 and appears DuckDB and Spark are good at this. 

For the first query:
DuckDB: ~161s
Dask: ~ timeout after 1200s
Spark: ~69s
https://cloud.coiled.io/clusters/288892/information?account=dask-benchmarks&tab=Metrics


Do we want to add a fixture/preload step to download the data to disk first though? I don't want to create any bias here, but Dask appears to be much slower than DuckDB and Spark when reading from S3 from a single node, does much better from a cluster as we've seen.

phofl:
This doesn’t reflect my experience at all. I’ll take a closer look. What
type of VM did you use?

Miles ***@***.***> schrieb am Di. 10. Okt. 2023 um 14:33:

phofl:
Sorry should have checked the cluster link

phofl:
This runs in 43 seconds for me:

```
import os
from datetime import datetime

import dask_expr as dd

DATASETS = {
    "scale 100": "s3://coiled-runtime-ci/tpch_scale_100/",
    "scale 1000": "s3://coiled-runtime-ci/tpch-scale-1000/",
}

enabled_dataset = os.getenv("TPCH_SCALE")
if enabled_dataset is not None:
    if enabled_dataset not in DATASETS:
        raise ValueError("Unknown tpch dataset: ", enabled_dataset)
else:
    enabled_dataset = "scale 100"


def read_data(filename):
    path = DATASETS[enabled_dataset] + filename + "/"
    return dd.read_parquet(path, engine="pyarrow")

from distributed import LocalCluster
cluster = LocalCluster()
client = cluster.get_client()



VAR1 = datetime(1998, 9, 2)
lineitem_ds = read_data("lineitem")

lineitem_filtered = lineitem_ds[lineitem_ds.l_shipdate <= VAR1]
lineitem_filtered["sum_qty"] = lineitem_filtered.l_quantity
lineitem_filtered["sum_base_price"] = lineitem_filtered.l_extendedprice
lineitem_filtered["avg_qty"] = lineitem_filtered.l_quantity
lineitem_filtered["avg_price"] = lineitem_filtered.l_extendedprice
lineitem_filtered["sum_disc_price"] = lineitem_filtered.l_extendedprice * (
    1 - lineitem_filtered.l_discount
)
lineitem_filtered["sum_charge"] = (
    lineitem_filtered.l_extendedprice
    * (1 - lineitem_filtered.l_discount)
    * (1 + lineitem_filtered.l_tax)
)
lineitem_filtered["avg_disc"] = lineitem_filtered.l_discount
lineitem_filtered["count_order"] = lineitem_filtered.l_orderkey
gb = lineitem_filtered.groupby(["l_returnflag", "l_linestatus"])

total = gb.agg(
    {
        "sum_qty": "sum",
        "sum_base_price": "sum",
        "sum_disc_price": "sum",
        "sum_charge": "sum",
        "avg_qty": "mean",
        "avg_price": "mean",
        "avg_disc": "mean",
        "count_order": "count",
    }
)

total.reset_index().compute().sort_values(["l_returnflag", "l_linestatus"])
```

tests/benchmarks/test_tpch_pyspark.py
```diff
@@ -0,0 +1,338 @@
...
+
+            func.__name__ = f"test_query_{query_num}"
+            return func
+
+        for query_num in range(1, 23):
+            # Limit to queries implemented in dask
+            test_tpch_dask = getattr(test_tpch, f"test_query_{query_num}", None)
+            if test_tpch_dask is not None:
+                test = make_test(test_tpch_dask, test_tpch_pyspark, query_num)
+                setattr(cls, test.__name__, test)
```
mrocklin, line 239: I'll admit that I find these tests somewhat confusing.  I feel like it's maybe too DRY, and so hard to reason about maybe.  I wonder if a different approach would make sense where we had some fixtures to do some of the preparation, and then had simple tests written like `test_tpch/duckdb.py::test_q1` `test_tpch/duckdb::test_q2` and so on.  This structure seems a little more flat / unmagical to me and maybe easier for others (or at least me) to understand.

milesgranger:
> I think though that you are not creating a LocalCluster in your workload which would explain the difference

My first try locally used LocalCluster with the same result; maybe something awkward is happening since I'm starting a cluster within a cluster? (Since coiled functions has scheduler and worker on the same node) I'm not sure, but will poke at it some more, thanks.

tests/benchmarks/test_tpch_pyspark.py
mrocklin: > It was suggested to just get something working and not concern oneself too much with the implementation.

Understood, and please don't take my comment here as a directive from on-high, but rather as a perspective of another engineer.   Personally I default to flat architectures, especially when other people have to interact with them.  I find that things that do automation like this tend to not get updated because other people have difficulty understanding them.

phofl:
Could you try to get timing for DuckDB and Spark? Even if Dask doesn't work, I want to look into a few things tomorrow and some numbers would be helpful

milesgranger:
Here are some timings using a m6i.24xlarge: (will update with Dask once I have that going)

| Query Number | DuckDB Total Time | PySpark Total Time | DuckDB Query Time | PySpark Query Time |
|--------------|-------------------|--------------------|-------------------|--------------------|
| 1            | 77.563586         | 49.960665          | 77.097343         | 16.664233          |
| 2            | 21.386901         | 18.402115          | 21.360888         | 13.950061          |
| 3            | 85.677673         | 60.846358          | 85.646625         | 47.057319          |
| 4            | -                 | 49.237413          | -                 | 45.502427          |
| 5            | -                 | 41.079659          | -                 | 36.684029          |
| 6            | -                 | 8.677918           | -                 | 4.802619           |
| 7            | -                 | 23.965761          | -                 | 18.127046          |

The ones with duckdb missing are either failed with [RuntimeError: Resource temporarily unavailable](https://github.com/duckdb/duckdb/issues/3391), or says it ran out of memory, which is almost certainly not true [given the metrics on the cluster](https://cloud.coiled.io/clusters/289389/information?account=dask-benchmarks&tab=Metrics&filterPattern=) and the low (<0.5GB) it was trying to allocate in the message)

Split it into total time and query time, as PySpark / DuckDB has a bit of setup code before running the query itself. ex. PySpark downloading the jars etc.

phofl:
DuckDB beats Polars on smaller scale, so it is worth an investigation whether they scale less good or if we are doing something wrong.

mrocklin:
If I wanted to get numbers from these runs from main (once this is merged), where would I find that data?  People have mentioned a sqlite database somewhere.  Where does that live?

milesgranger:
> If I wanted to get numbers from these runs from main (once this is merged), where would I find that data?

Yep, as you probably saw in the other issue, [passing `--benchmark` creates the sqlite db. ](https://github.com/coiled/benchmarks#running-the-benchmarks-locally). The numbers I've been posting here are just using `timeit` at various places for quick output, and to ignore pyspark setup time (downloading JARs and what not).

mrocklin:
> The numbers I've been posting here are just using timeit at various places for quick output, and to ignore pyspark setup time (downloading JARs and what not)

Does this mean that accurate PySpark numbers aren't in the database?

If so, the way I would resolve this is by making test fixtures that set up everything you need to run a Spark query, and then I would have the test function just run the query.  My guess/hope is that timings in the database will reflect just the test execution time, and not the fixture setup/teardown time.

mrocklin:
> If so, the way I would resolve this is by making test fixtures that set up everything you need to run a Spark query, and then I would have the test function just run the query. My guess/hope is that timings in the database will reflect just the test execution time, and not the fixture setup/teardown time

Hrm, looking at the code it looks like that's what you're doing with the `tpch_pyspark_client`.  I'm curious, also, where is this created?  I'm not able to find this even after grepping through the codebase.

milesgranger:
> Does this mean that accurate PySpark numbers aren't in the database?

I'd say they are mostly accurate, after reviewing it seems they are only downloaded once at the cost of approx 5-10sec; after that the setup time seems trivial.


---

I think this PR is ready for review now, and apologies for the many files.

There is a flag on here now to ignore all these new benchmarks until we figure out how we want A/B normal runs; probably updating the dashboard A/B script to combine across `test_tpch` (dask) and `test_tpch_pyspark`. As `test_tpch_comparison_cluster.py` might need to go altogether. That was placed just to run dask/pyspark on same cluster to get a proper A/B test graph which we deemed as not needed.

The single node benchmarking works for all engines except Dask, b/c there is some issue starting a new LocalCluster on the scheduler already running dask. Will check in with James later on for this.

To run pyspark cluster benchmarks:
`python -m pytest -k test_tpch_pyspark --tpch-non-dask` 

> tpch_pyspark_client. I'm curious, also, where is this created?

That's in `tests/conftest.py` line 556.

mrocklin:
> That's in tests/conftest.py line 556

Ah, indeed.  My apologies.

mrocklin:
It looks like you maybe need to add polars and python-duckdb to the environment.yml.

mrocklin:
I'd like us to reconsider how we lay out these tests.  Right now the organization seems to be the following:

-  test_tpch.py: Dask queries
- test_tpch_comparison_cluster: Dask and Spark queries
- test_tpch_single_machine: Dask and Spark and Polars and DuckDB queries

This organization doesn't seem natural to me.  It seems to have a lot more to do with the history of how we came up with these things to run, rather than our implementation than with how we'd think about things conceptually.  This is totally fine.  It makes sense for things to grow organically and then rethink in the future.

I propose that we aim towards the following instead:

We have a `tests/tpch/` directory (Or maybe just a `tpch` directory at root level?  Any reason not to flatten things?)

We then have the following files:

- dask.py
- spark.py
- polars.py
- duckdb.py

These files contain the implementations for each query, they're also parametrized by scale and insfrastructure.  For example in dask.py we would have:

```python
@pytest.fixture
def client_and_dataset(...):
    ... # goes through the different hardware we want to test on

def test_query_1(client_and_dataset):
    client, dataset = client_and_dataset

    df = dd.read_parquet(dataset)
    ...
```

We would make different hardware for Dask, Spark, Polars, and DuckDB (no need to mix these I think).  

Maybe this makes things more complex for single-machine Dask. That probably warrants more thought.  I think I care more about things be organized though than I care about that specific case.  (others please speak up if you disagree.)

mrocklin:
When running this locally I get these errors:

```python-traceback
tests/benchmarks/test_tpch_comparison_cluster.py:89: in _run_test
    results = pyspark_test(client)
tests/benchmarks/test_tpch_pyspark.py:63: in test_query_7
    return run_tpch_pyspark(tpch_pyspark_client, queries.q7)
tests/benchmarks/test_tpch_pyspark.py:136: in run_tpch_pyspark
    timing, rows = tpch_pyspark_client.run_on_scheduler(_run_tpch)
../../mambaforge/envs/test-env/lib/python3.11/site-packages/distributed/client.py:2863: in run_on_scheduler
    return self.sync(self._run_on_scheduler, function, *args, **kwargs)
../../mambaforge/envs/test-env/lib/python3.11/site-packages/distributed/utils.py:357: in sync
    return sync(
../../mambaforge/envs/test-env/lib/python3.11/site-packages/distributed/utils.py:424: in sync
    raise exc.with_traceback(tb)
../../mambaforge/envs/test-env/lib/python3.11/site-packages/distributed/utils.py:397: in f
    result = yield future
../../mambaforge/envs/test-env/lib/python3.11/site-packages/tornado/gen.py:767: in run
    value = future.result()
../../mambaforge/envs/test-env/lib/python3.11/site-packages/distributed/client.py:2822: in _run_on_scheduler
    raise exc.with_traceback(tb)
tests/benchmarks/test_tpch_pyspark.py:133: in _run_tpch
    return await asyncio.to_thread(_)
/opt/coiled/env/lib/python3.11/asyncio/threads.py:25: in to_thread
    ???
/opt/coiled/env/lib/python3.11/concurrent/futures/thread.py:58: in run
    ???
tests/benchmarks/test_tpch_pyspark.py:118: in _
    module.setup(spark)  # read spark tables query will select from
/opt/coiled/env/lib/python3.11/site-packages/tests/benchmarks/tpch_pyspark_queries/q7.py:48: in setup
    ???
/opt/coiled/env/lib/python3.11/site-packages/tests/benchmarks/tpch_pyspark_queries/utils.py:44: in read_parquet_spark
    ???
/opt/coiled/env/lib/python3.11/site-packages/pyspark/sql/readwriter.py:531: in parquet
    ???
/opt/coiled/env/lib/python3.11/site-packages/py4j/java_gateway.py:1322: in __call__
    ???
/opt/coiled/env/lib/python3.11/site-packages/pyspark/errors/exceptions/captured.py:169: in deco
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   Exception: Py4JJavaError('An error occurred while calling o89.parquet.\n', JavaObject id=o92)
```

milesgranger:
> When running this locally I get these errors:

Interesting, I got that as well a while ago and now though it seems to be fine. I wonder if the maven repo jars are downloaded from was down for a bit?

[Ran all 7 queries just now](https://cloud.coiled.io/clusters/290323/information?account=dask-benchmarks&tab=Metrics)

I agree on the file layout for the most part. I'll adjust and probably just remove the `test_tpch_comparison_cluster.py` file as we can update the dashboard script to pull together `test_tpch_pyspark/dask.py` files and their tests later on.

ci/environment.yml
```diff
@@ -1,7 +1,12 @@
 channels:
   - conda-forge
 dependencies:
+  - conda
   - python >=3.9
+  - pyspark ==3.4.1
+  - openjdk ==20.0.2
+  - python-duckdb ==0.9.0
```
mrocklin, line 8: I think that you want 0.9.1 here.  It was released literally minutes ago, and it stops a conflict with pandas 2.1.0

phofl:
Can we remove the queries that we don't need at the moment? No use in having them here until we add them ourselves.

mrocklin:
> Can we remove the queries that we don't need at the moment? No use in having them here until we add them ourselves.

I think that my preference is to keep them, but disable the tests for the moment.  There might be value here in avoiding future work.

mrocklin:
@milesgranger if you agree with the Polars changes then is it easy for you to do something similar with DuckDB?

milesgranger:
Just updated DuckDB portion, but removed the queries before I saw your comment. Should I add the remainder in the same fashion and then skip them in normal testing?

phofl:
> There might be value here in avoiding future work.

I was looking at this from the other direction. We will have to update them when we eventually want to use them (and more importantly we have to actually remember this when adding new queries), so better to copy as needed

mrocklin:
I don't have a strong preference

milesgranger:
> We will have to update them when we eventually want to use them (and more importantly we have to actually remember this when adding new queries), so better to copy as needed

Just remove the `pytest.mark.skip`? This was my original approach in implementing all the queries then limited to however many have been implemented in Dask. But here, 1) the work is done. 2) only need to remove the skip marker. 3) maybe I'm missing something. :)

mrocklin:
I think that Patrick is saying that by the time we implement these queries it's decently likely that the upstream implementations will also have changed, so we'll have to go back and look at them anyway.  

If so, then I guess this depends on how much work there is to modify the upstream queries, and how frequently they change.

tests/benchmarks/tpch/test_comparison_single_vm.py
```diff
@@ -0,0 +1,142 @@
+import dask
```
mrocklin, line 1: Is this file still useful, or are we able to remove it?

mrocklin:
I'm curious if anyone has thoughts on moving the entire `tpch/` directory to root level, rather than having it in `tests/benchmarks/tpch/`.  I think that this is actually two questions:

1. Will this screw up any particular infrastructure that we have today, like benchmark plots?
2. Do people agree with this aesthetic?

milesgranger:
> These queries change regularly over In the polars repo, so I wanted to avoid keeping stuff in synch that we are not using anyway

I don't understand, then don't update them? They're being skipped anyway. Anyhow, it appears you feel strongly about this so I'll remove them.

tests/benchmarks/tpch/test_comparison_single_vm.py
milesgranger: Probably, Polars/DuckDB are already running single node, I suppose we'll have something similar for spark/dask to get them on single. Will remove it.

phofl:
It's likely that folks would just add our implementation without looking at the upstream changes. I want to avoid the cognitive load here since copy-paste should be quick at a later point.

> Do people agree with this aesthetic?

No preference one way or the other

mrocklin:
Anything stopping us from merging?  I'm sensitive to having this PR stay around over the weekend.

tests/conftest.py
```diff
@@ -65,6 +70,15 @@ def pytest_collection_modifyitems(config, items):
...
             item.add_marker(skip_workflows)
 
+    skip_benchmarks = pytest.mark.skip(reason="need --tpch-non-dask option to run")
+    for item in items:
+        if not config.getoption("--tpch-non-dask") and not (
+            str(item.path).startswith(
+                str(TEST_DIR / "benchmarks" / "tpch" / "test_dask")
+            )
+        ):
+            item.add_marker(skip_benchmarks)
```
fjetter, line 80: Please open a follow up ticket for this. We don't need this for this PR but I want this to run somewhat regularly (every commit, once a day, etc.)

milesgranger:
Nothing from my end stopping a merge. 👍

tests/benchmarks/tpch/test_pyspark.py
```diff
@@ -0,0 +1,229 @@
...
+        return await asyncio.to_thread(_)
+
+    if not is_local:
+        rows = tpch_pyspark_client.run_on_scheduler(_run_tpch)
+    else:
+        rows = asyncio.run(_run_tpch(None))  # running locally
+    print(f"Received {len(rows)} rows")
+
+
+class SparkMaster(SchedulerPlugin):
```
fjetter, line 136: I would also suggest as a follow up to

- Move the plugins into a spark.py submodule or something like this. That makes it easier to import if one runs a couple of ad-hoc tests. At least until we have this in coiled
- I would refactor the `run_tpch_pyspark` function slightly to allow it being used as `run_spark_query(dask_client, query)` for similar reasons.

tests/conftest.py
```diff
@@ -526,8 +585,9 @@ def tpch_cluster(request, dask_env_variables, cluster_kwargs, github_cluster_tag
         **cluster_kwargs["tpch"],
     )
     dump_cluster_kwargs(kwargs, f"tpch.{module}")
-    with Cluster(**kwargs) as cluster:
-        yield cluster
+    with dask.config.set({"distributed.scheduler.worker-saturation": "inf"}):
```
fjetter, line 588: I'm a bit confused. This should be on main already, isn't it?

fjetter (commented):
I have some suggestions for follow up work but I would like to not do this on this PR. There is one question regarding configuration of the dask TCPH benchmarks and queuing that we should address. This PR should not change any other benchmarks IMO

If that is a no-op, we're good to go from my POV

tests/conftest.py
fjetter: Ok, I think this is a rendering issue https://github.com/coiled/benchmarks/blob/b8f0f0ee44fcd4d92c1e244a8b4c6c709ae3079f/tests/conftest.py#L518-L531

tests/benchmarks/tpch/test_pyspark.py
mrocklin: I think that for pyspark we merge it as-is, fix up spark outside of this repo, and then make this file like the others, where the query is just defined in the test.

mrocklin: Future work regardless though.

tests/benchmarks/tpch/test_pyspark.py
```diff
@@ -0,0 +1,229 @@
...
+def fix_timestamp_ns_columns(query):
+    """
+    scale100 stores l_shipdate/o_orderdate as timestamp[us]
+    scale1000 stores l_shipdate/o_orderdate as timestamp[ns] which gives:
+        Illegal Parquet type: INT64 (TIMESTAMP(NANOS,true))
+    so we set spark.sql.legacy.parquet.nanosAsLong and then convert to timestamp.
+    """
+    for name in ("l_shipdate", "o_orderdate"):
+        query = re.sub(rf"\b{name}\b", f"to_timestamp(cast({name} as string))", query)
+    return query
```
fjetter, line 83: @milesgranger please also open an issue for this. I don't feel comfortable with this string casting when we're comparing us to spark. If that means we have to regenerate the dataset, that's unfortunate but I wouldn't want us to bias spark too severely. At the very least we should confirm on smaller / other data what the impact here is and whether it can be ignored.

fjetter (approved):

PR #1045, opened 2023-10-03
this is a test - no changes

dchudz:


fjetter:
FYI You don't need to open a PR. I believe it is sufficient to open a branch to run CI

ntabris:
Does that mean you don't want us opening DNM PRs? (I find PR simpler because it links to the CI runs)

fjetter:
> Does that mean you don't want us opening DNM PRs? (I find PR simpler because it links to the CI runs)


I don't really care. Just wanted to point out this isn't necessary. I agree, it's easier to track (and ideally a bot would just post the results to the PR)

PR #1046, opened 2023-10-05
Enable dask-expr for tpch queries

phofl:
Not sure if we should rather pin dask-expr in the ab tests


fjetter:
why do you think we should pin it?

fjetter (approved):

phofl:
I am not aware of the rational behind pinning everything else, I don't see any strong reasons

Some things might become tricky since we push the requirement of dask/dask regularly

PR #1058, opened 2023-10-12
Flatten Polars

mrocklin:
This brings the polars work into a single file.  I think that this is simpler and easier for outsiders to understand.

This also leverages `coiled.functions` a bit more explicitly in tests, and pulls things like restrating and warm starts out to fixtures so that they don't get confused with execution times. 

Looking at results, it looks like we're pretty severly bound by bandwidth (I think so anyway, not sure https://github.com/coiled/platform/issues/3521)

@milesgranger @fjetter  thoughts on this approach?


mrocklin:
If folks are ok with this approach then additional work:
- Do this again with DuckDB
- I looked into doing this with Spark, but I haven't figured out yet how to make that system clean.  I think I'm going to pull out the Spark bits first and then see what happens
- Dask stuff seems ok
- I'd like to pull `tests/benchmarks/tpch/` down to `tpch/`.  To me it doesn't seem like the nested directory structure buys us anything here

PR #1076, opened 2023-10-16
Add IAMInstanceCredentialsProvider for Spark config

milesgranger:
Will fix #1074 

As explained there, prior to v1 it seems we could reliably use env vars, but after https://github.com/coiled/platform/pull/3396(?) using IAMInstanceCredentialsProvider is preferred.

PR #1077, opened 2023-10-16
Add session token for duckdb AWS use

### LINKED ISSUE #1068: TPCH - DuckDB Failures
I'm getting segfaults when running DuckDB locally.  

When I run as-is I get the following:

```
E       duckdb.duckdb.HTTPException: HTTP Error: HTTP GET error on '/?encoding-type=url&list-type=2&prefix=tpch_scale_100%2Flineitem%2F' (HTTP 403)
```

This is after going through a clean install as outlined in the README, followed by the following:

```
py.test tests/benchmarks/tpch/test_duckdb.py::test_query_1 --tpch-non-dask
```

@milesgranger I'm guessing that you're probably the best positioned to help here

milesgranger:
Fix #1068

After coiled https://github.com/coiled/platform/pull/3396 switch in credential handling, and similar to #1076 this updates duckdb failing by adding a missing session token.


phofl:
The extension https://duckdb.org/docs/extensions/aws should handle the variables automatically with the newest duckdb release (0.9.x). At least that was the promise when I wrote my blog post but that was before 0.9.x was out

milesgranger:
I tried that.

`CALL load_aws_credentials()` but it does not work, and is experimental anyway so I didn't chase it much further beyond that. (tried setting and not setting region as well in this exploration with no additional benefit)

phofl:
Ok that's a pity, you can ignore my comment then

crusaderky:
query_1 fails in CI neither before nor after this PR, even if the coiled version is 1.1.0.
Not sure what's the expected change?

mrocklin (approved):
Things used to fail for me before and now they run.  I'm +1 on this.

mrocklin:
Although I don't know about CI failures.  Were things already failing on main?

milesgranger:
@mrocklin, that's being looked into now by @phofl 
https://github.com/coiled/benchmarks/actions/runs/6532939100/job/17737096347#step:10:750

PR #1080, opened 2023-10-16
Pin dask-expr; realign A/B envs

crusaderky:


ci/environment.yml
```diff
@@ -1,19 +1,14 @@
 channels:
   - conda-forge
 dependencies:
-  - conda
```
crusaderky, line 4: redundant with environment-test.yml

ci/environment.yml
```diff
@@ -1,19 +1,14 @@
...
-  - python-duckdb ==0.9.1
-  - polars >=0.19.3,<=0.19.8  # 0.19.3 on Windows, 0.19.8 on Linux and MacOS
   - pip
   - coiled >=0.2.54
   - numpy ==1.24.4
   - pandas ==2.1.1
   - dask ==2023.9.2
   - distributed ==2023.9.2
-  - dask-expr
+  - dask-expr ==0.1.8
```
crusaderky, line 11: Note that both before and after this is overridden by environment-git-tip.yml. This change is for the benefit of A/B testing.

milesgranger (approved):
Thank you!

ci/environment.yml
phofl: So nightly will consistently run with main?

crusaderky: Yes, both nightly and PRs use the overlay of
1. python={3.9|3.10|3.11} depending on matrix in `.github/workflows/tests.yml`
2. `ci/environment.yml`
3. in a single matrix configuration, `ci/environment-snowflake.yml`
4. `ci/environment-git-tip.yml`
5. `ci/environment-test.yml`

Whereas A/B tests use
1. one of `AB_environments/AB_*.conda.yml`
2. `ci/environment-test.yml`

Issue #1097, opened 2023-10-17
Report only shows data for TPC-H tests since 2023-10-14


hendrikmakait:
The reporting system only includes data for the TPC-H related tests since 2023-10-14. We stopped showing (persisting/recording?) data for all other tests.

<img width="1674" alt="Screenshot 2023-10-17 at 10 56 26" src="https://github.com/coiled/benchmarks/assets/2699097/a45cea21-b071-4a55-a411-33b7106a2bdd">

phofl:
There are 2 different plots no, one before renaming the directory and one after

hendrikmakait:
> There are 2 different plots no, one before renaming the directory and one after

That's #1098.

hendrikmakait:
Closed as completed by #1101

Issue #1099, opened 2023-10-17
Reorganize Benchmarks


mrocklin:
I was just chatting with @fjetter about #1094 and the general struture of this repository.  

Currently things look like this:

-  tests
    -   benchmarks
        - tpch
    - runtime
    - stability
    - workflows

In my ideal world we'd have something like the following:

- tpch
- arrays
- stability
- ...

Maybe that's not possible though because Python doesn't like importing things from root level, and we'll likely have some shared utilities.  If that's true then maybe we need the following?

- benchmarks
    - tpch
    - arrays
    - stability
    - ...

Also, are there things that we should remove?  For example, `runtime`?

mrocklin:
I'd love to get things down to root level, but that requires being careful in a way that I think is unlikely in the next couple of days.  With that in mind I'll suggest the following changes:

-  Rename `tests` to `benchmarks`
-  Rename `tests/benchmarks` to `benchmarks/general`
- Remove `tests/runtime` (unless people actively look at this
- I'm not sure what to do with `tests/workflows`.  Do we use these today?

These seem inoffensive to me and like a good small step in this direction.  I may be understimating the costs of small steps though.  In particular I'm not sure how expensive it is to construct alembic migrations.

fjetter:
> I'm not sure what to do with tests/workflows. Do we use these today?

I consider those rather integration tests than anything else. For instance, we run a snowflake thing on coiled. If this lights up for whatever reason, we should get notified.
From that POV, other items in that folder are less interesting and I'd be fine with nuking them and move the snowflake test to `stability` instead (or `integration`)

fjetter:
> In particular I'm not sure how expensive it is to construct alembic migrations.

It's cheap and not difficult, just a little annoying. We also don't have any tests for the migration but if we mess this up, we can "restore" the database from an earlier GH actions run (assuming we notice in time; FWIW we never had to do this)

```python
alembic revision -m"Rename foo to bar"
```

generates a new file in `alembic/versions` with the appropriate revisions already set. We then just have to write some simple sql that describes the changes.

For example, we moved the H2O tests once from one directory to another. The SQL for that was
```sql
        update test_run
        set path = 'benchmarks/test_h2o.py'
        where path = 'benchmarks/h2o/test_h2o_benchmarks.py';
```


https://github.com/coiled/benchmarks/blob/main/alembic/versions/9d6f8ea24ee1_move_h2o_tests.py

hendrikmakait:
[...]

PR #1101, opened 2023-10-18
Don't skip all tests

fjetter:
This toggle caused us to skip all tests and just run the TCPH dask benchmarks.

This toggle will either be obsolete or has to change after https://github.com/coiled/benchmarks/issues/1099 is done


milesgranger (approved):
Ouch, sorry! 
Thanks and you might also be interested in https://github.com/coiled/benchmarks/pull/1083, adding this flag to the schedule / label conditional

hendrikmakait:
Closes #1097

PR #1102, opened 2023-10-18
Add importorskip for duckdb, polars, and pyspark

milesgranger:
On request of @phofl. Beneficial if one runs A/B tests and removes these to lighten the load during package sync and environment setup time.


phofl:
More context: It's not high priority, but it feels like a string code smell that you need Spark to run dask-expr/polars/duckdb queries. Especially if we want to broadcast the repository more.

fjetter (approved):

fjetter:
see https://github.com/coiled/benchmarks/pull/1108

phofl:
Yeah I think that's my main concern, there is some weird cross importing going on that we should avoid

PR #1104, opened 2023-10-18
Update coiled_function decorator

mrocklin:
Two changes:

1.  Remove the old functools.wraps functionality, now that upstream handles this for us.
2.  Add the module name into the cluster name, so that we can run duckdb and polars simultaneously more easily

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -1,5 +1,6 @@
-import functools
 import os
+import sys
+import threading
 import uuid
 
 import coiled
@@ -10,7 +11,7 @@ from dask.distributed import LocalCluster
 
 @pytest.fixture(scope="session")
 def scale():
-    return 100
+    return 1000
 
 
 _local = False
@@ -26,7 +27,7 @@ def dataset_path(local, scale):
     remote_paths = {
         10: "s3://coiled-runtime-ci/tpch_scale_10/",
         100: "s3://coiled-runtime-ci/tpch_scale_100/",
-        1000: "s3://coiled-runtime-ci/tpch-scale-1000/",
+        1000: "s3://coiled-runtime-ci/tpch-scale-1000-date/",
     }
     local_paths = {
         10: "./tpch-data/scale10/",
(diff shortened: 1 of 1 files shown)


mrocklin:
Still working on this.

tests/tpch/conftest.py
```diff
@@ -134,14 +135,17 @@ def spark(spark_setup, benchmark_time):
 
 def coiled_function(**kwargs):
     def _(function):
+        frame = sys._current_frames()[threading.get_ident()]
+        filename = frame.f_back.f_code.co_filename
+        module = filename.split("_")[-1].split(".")[0]  # dask, spark, ...
```
jrbourbeau, line 140: Just a driveby comment but maybe `inspect.getmodule` or `function.__module__` is useful here

mrocklin: Thanks!  That's much nicer.

mrocklin:
OK, this is feeling good now.

milesgranger (approved):
A go from me, assuming tests are happy. :+1:

mrocklin:
Test failure points to this run: https://cloud.coiled.io/clusters/294695/information?account=dask-benchmarks&computation=d9e9bd9d-4ef9-4455-8ff9-7c8206bb4344&sinceMs=1697656440506&untilMs=1697656440764&tab=Code&filterPattern=

I'm not getting any information from the platform, even logs.  Things just hang.  @dchudz I'm curious if you all are able to see anything more deeply here on this run.

dchudz:
@mrocklin the cluster has plenty of logs.

is your link intentionally only for a specific time range?

my interpretation would be that nothing happened in that time range. 

I can't say more without knowing more about how the benchmarks work or how that link was put together.

dchudz:
oh, maybe that's a link for an *empty* time range?

Screen Shot 2023-10-18 at 4 26 10 PM

mrocklin:
Oh odd.  I clicked on the code block for `test_query_2`, the query that failed, and that fits in that 20 minute block of emptiness (I think this is also the timeout period with pytest) assuming that it was taking up that 20 minute block.  It looks like I was mistaken though.  

cc'ing also @fjetter in case he can help to understand it.  Maybe flakiness on Dask's part?

dchudz:
probably this isn't very interesting, but:

* Matt's original link is a zero-minute block
* in the 20-minute block of emptiness ([new link](https://cloud.coiled.io/clusters/294695/information?account=dask-benchmarks&computation=&tab=Metrics&filterPattern=&sinceMs=1697656446977&untilMs=1697657636977)) there are metrics, but indeed no tasks, and no logs

PR #1094, opened 2023-10-17
WIP - Pull TPC-H out of tests/benchmarks

mrocklin:
This moves the TPC-H benchmarks outside of the `tests/benchmarks` directory, and treats it as more of a standalone system.  

Then, this also rewrites some of the relevant `conftest.py` file in a way that, I think, makes it simpler to manipulate (that might be subjective though).  

Some notable changes:

1.  There are `local` and `scale` fixtures that control datasets and should eventually control clusters that are used
2. pyspark uses the dask infrastructure (it's exactly the same)
3. Cleaned up some very old config settings in cluster_kwargs.yaml

However, I suspect that this approach screws with existing tests because they import things from conftest, which is now at root level.  Things will still need to be moved around to make this work.  Mostly right now I'm looking for if this makes sense to do, or if there are major reasons why not to do it.

To give more motivation, if TPC-H benchmarks work, people will want to come here and look at them.  If they're highly intertwined with the existing benchmark machinery then I think that they will be hard to understand and run.  I would like to make the tpc-h benchmarks more accessible to people who are unfamiliar with this repository.

PR #1105, opened 2023-10-18
Add migration for tpch tests restructure from #1094 and #1044

milesgranger:
Manually checked this out from the benchmark db on S3; after migration is applied:

image


hendrikmakait:
IIUC, we still lack a migration for the changes in #1044, which should arguably be applied before the migration for #1094. Is that correct?

hendrikmakait (approved):
Dashboard finally looks good again, thanks!

PR #1106, opened 2023-10-18
Generate TPC-H data script

milesgranger:
Rough script, easy enough way to generate data to up to around scale 100 seems okay. Can probably go beyond by not doing it in-memory.  Can configure the desired output size of files

It's at least easier than dbgen directly.


phofl (commented):
This is to download from s3?

milesgranger:
No, it generates tpch in duckdb [using the tpch extension](https://duckdb.org/docs/extensions/tpch.html) and uploads to s3 (or locally if a local path)

phofl:
Interesting,

more docs would be appreciated.

Does this work properly?

tpch_gen.py
```diff
@@ -0,0 +1,93 @@
...
+@click.command()
+@click.option("--sf", default=10, help="Scale factor to use")
+@click.option("--target-mb", default=50, help="Target output parquet file size")
+@click.option(
+    "--path",
+    default="./tpch-data",
+    help="Local or S3 base path, will affix 'scale-<scale>' subdirectory to this path",
+)
+@coiled.function(region="us-east-2", memory="512 GiB", disk_size=500)
+def main(sf, target_mb, path):
```
mrocklin, line 19: My preference would be to have this also accessible as a function that people can call from Python.  I think that this will make this code more flexibly useful.  

Maybe one way to do this would be to pull out most of the body of the `main` code to a standalone function, and then to lightly wrap that function in a main function with all the click decorators.  That way we could do both. Thoughts?

Also, the names `sf`, `target_mb` are not immediately obvious to me.  I suggest instead `scale_factor` and `target_size` (but other names are probably good as long as they're understandable without having to refer to docs or comments).

mrocklin:
I haven't played with this yet, so don't have very good feedback. I'll try it this afternoon though.

tpch_gen.py
milesgranger: Ya it's rough.. feel free to push any changes you like when playing with it. I'll make some further updates tomorrow.

mrocklin:
In principle I'm excited about the direction here though.  Thank you.

mrocklin (approved):

PR #790, opened 2023-04-13
Use `da.random.default_rng` API for random array creation

jrbourbeau:

PR #1109, opened 2023-10-18
Rewrite tests for rechunking

hendrikmakait:


hendrikmakait:
Follow-up: Add tests for median over time from https://discourse.pangeo.io/t/rechunking-large-data-at-constant-memory-in-dask-experimental/3266

mrocklin:
This will cut in half test_array.py, which is our current longest running test file at about 33 minutes, so in general I encourage this work

tests/benchmarks/test_rechunk.py
```diff
@@ -0,0 +1,36 @@
...
+@pytest.fixture
+def configure_chunksize(chunksize):
+    with dask.config.set({"array.chunk-size": chunksize}):
+        yield
+
+def test_tiles_to_rows(small_client, memory_multiplier, configure_chunksize, configure_rechunking):
+    memory = cluster_memory(small_client)
+    shape = scaled_array_shape(memory * memory_multiplier, ("x", "x"))
+
+    rng = da.random.default_rng()
```
mrocklin, line 25: I'm curious, why this rather than the normal `da.random.random` function?

tests/benchmarks/test_rechunk.py
```diff
@@ -0,0 +1,36 @@
...
+    arr = rng.random(shape, chunks="auto")
+    arr.rechunk((-1, "auto")).sum().compute()
+
+
+def test_swap_axes(small_client, memory_multiplier, configure_chunksize, configure_rechunking):
```
mrocklin, line 30: There's a lot of matrixed fixtures here.  This makes me concerned about runtime.

Have you looked into what this takes to run?

mrocklin (commented):
Hrm, this is now the longest running test, it also is five of the top 20 or so.  It might make sense to think hard about if we want all of these, or if we do then maybe we block until we have an ability to run them daily or weekly rather than on every commit.

```
1015.31s call     tests/benchmarks/test_rechunk.py::test_swap_axes[1-8 MiB-p2p]
620.09s call     tests/benchmarks/test_join.py::test_join_big[1-tasks]
507.39s call     tests/benchmarks/test_parquet.py::test_read_spark_generated_data
444.51s call     tests/benchmarks/test_join.py::test_set_index[1-tasks-True]
409.57s call     tests/benchmarks/test_spill.py::test_dot_product_spill[compressible]
386.49s call     tests/benchmarks/test_spill.py::test_dot_product_spill[uncompressible]
385.21s call     tests/benchmarks/test_rechunk.py::test_swap_axes[1-8 MiB-tasks]
382.70s call     tests/benchmarks/test_join.py::test_join_big[1-p2p]
348.86s setup    tests/benchmarks/test_rechunk.py::test_tiles_to_rows[0.1-8 MiB-tasks]
316.16s call     tests/benchmarks/test_join.py::test_set_index[1-p2p-True]
312.03s call     tests/stability/test_deadlock.py::test_repeated_merge_spill
300.42s call     tests/benchmarks/test_dataframe.py::test_shuffle[1-tasks]
281.17s call     tests/benchmarks/test_spill.py::test_spilling[uncompressible-keep]
266.92s call     tests/benchmarks/test_join.py::test_set_index[1-tasks-False]
218.54s call     tests/benchmarks/test_work_stealing.py::test_work_stealing_on_scaling_up
216.87s call     tests/benchmarks/test_join.py::test_set_index[1-p2p-False]
209.62s call     tests/benchmarks/test_dataframe.py::test_shuffle[1-p2p]
194.57s call     tests/benchmarks/test_parquet.py::test_write_wide_data
184.95s call     tests/benchmarks/test_work_stealing.py::test_work_stealing_on_inhomogeneous_workload
178.03s call     tests/benchmarks/test_h2o.py::test_q8[5 GB (parquet)-tasks]
177.07s call     tests/benchmarks/test_rechunk.py::test_tiles_to_rows[1-8 MiB-p2p]
175.89s call     tests/benchmarks/test_rechunk.py::test_tiles_to_rows[1-8 MiB-tasks]
```

tests/benchmarks/test_rechunk.py
hendrikmakait: Originally because of coiled/benchmarks#790, but it looks like dask/dask#10247 may have fixed this.

hendrikmakait:
> Hrm, this is now the longest running test, it also is five of the top 20 or so. It might make sense to think hard about if we want all of these, or if we do then maybe we block until we have an ability to run them daily or weekly rather than on every commit.

I'm fine with skipping the `8 MiB` test variants for now and running them on demand/a longer interval. IMO, they are valuable when actively working on rechunking performance, but less so as general benchmarks that run every time.

mrocklin (approved):

PR #1121, opened 2023-10-20
Deactivate H2O tests on 0.5 GB

hendrikmakait:


fjetter (approved):

tests/benchmarks/test_h2o.py
```diff
@@ -31,8 +31,6 @@
         raise ValueError("Unknown h2o dataset(s): ", unknown_datasets)
 else:
     enabled_datasets = {
-        "0.5 GB (csv)",
```
fjetter, line 34: Do we have any other CSV tests? While I don't care deeply about this, I would care if this was 100% slower

fjetter (changes requested):

tests/benchmarks/test_h2o.py
hendrikmakait: There's `test_csv.py::test_csv_basic`. If we want something more integrated, I'd rather include `5 GB (csv)`. `0.5 GB` just doesn't make any sense for `dask`.

fjetter: `test_csv_basic` works for me. I just would like to get notified if we're significantly slower all of a sudden

fjetter (approved):

PR #1129, opened 2023-10-22
Reorganize fixtures for TPC-H

mrocklin:
Previously we had a split system to accommodate the single-machine systems with coiled.function.  Now we make those slightly less convenient in order to clean up the fixture infrastructure.


mrocklin:
Merging tomorrow morning my time if no one gets to it first.

PR #1134, opened 2023-10-24
Update TPCH Infrastructure

mrocklin:
This does a few things:

1.  Generate performance reports for every query if you add the `--performance-report` flag
2. Generate altair comparison charts
3. Change the infrastructure size up and down based on scale
4. Use Arrow filesystem (although I'm not sure how big a deal this is yet)
5. Improve charts to use consistent coloring, and drop ibis for pandas

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -1,10 +1,14 @@
+import contextlib
+import datetime
 import os
+import time
 import uuid
 
 import coiled
 import dask
+import filelock
 import pytest
-from dask.distributed import LocalCluster
+from dask.distributed import LocalCluster, performance_report
 
 ##################
 # Global Options #
@@ -13,14 +17,24 @@ from dask.distributed import LocalCluster
 
 def pytest_addoption(parser):
     parser.addoption("--local", action="store_true", default=False, help="")
+    parser.addoption("--cloud", action="store_false", dest="local", help="")
     parser.addoption("--restart", action="store_true", default=True, help="")
     parser.addoption("--no-restart", action="store_false", dest="restart", help="")
+    parser.addoption(
+        "--performance-report", action="store_true", default=False, help=""
+    )
 
     parser.addoption(
         "--scale",
         action="store",
         default=10,
         help="Scale to run, 10, 100, 1000, or 10000",
+    )
+    parser.addoption(
+        "--name",
+        action="store",
+        default="",
+        help="Name to use for run",
     )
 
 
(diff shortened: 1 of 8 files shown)


mrocklin:
Merging tomorrow morning my time if there are no comments

tests/tpch/conftest.py
```diff
@@ -44,10 +58,13 @@ def dataset_path(local, scale):
     remote_paths = {
         10: "s3://coiled-runtime-ci/tpch/scale-10/",
         100: "s3://coiled-runtime-ci/tpch/scale-100/",
+        # 1000: "s3://coiled-runtime-ci/tpc-h/scale-1000/",
         1000: "s3://coiled-runtime-ci/tpch/scale-1000/",
-        10000: "s3://coiled-runtime-ci/tpch/scale-10000/",
+        10000: "s3://coiled-runtime-ci/tpc-h/scale-10000/",
+        # 10000: "s3://coiled-runtime-ci/tpch/scale-10000/",
     }
```
milesgranger, line 65: This can now be:

```python
    remote_paths = {
        10: "s3://coiled-runtime-ci/tpc-h/scale-10/",
        100: "s3://coiled-runtime-ci/tpc-h/scale-100/",
        1000: "s3://coiled-runtime-ci/tpc-h/scale-1000/",
        10000: "s3://coiled-runtime-ci/tpc-h/scale-10000/",
    }
```
I'm generating the "strict" schema versions now (no casting of date or decimal types), maybe later you'll be interested in adding those as alternative options to run on.

tests/tpch/conftest.py
```diff
@@ -136,20 +288,54 @@ def spark(spark_setup, benchmark_time):
...
-}
-
-_uuid = uuid.uuid4().hex[:8]
+@pytest.fixture(scope="session")
+def machine_spec(scale):
+    if scale == 10:
+        return {
+            "vm_type": "m6i.8xlarge",
+        }
+    elif scale <= 100:
```
milesgranger, line 322:
```suggestion
    elif scale == 100:
```

Not that it actually matters. ;)

milesgranger (approved):
Just a couple comments, otherwise looks great!

tests/tpch/conftest.py
mrocklin: Thanks!

tests/tpch/conftest.py
mrocklin: Done.  Thanks.

phofl:
@mrockling arrow file system doesn't work with dask-expr main. I am guessing that you are still using my hacky branch that avoid the error? Should we invest effort into making this work or roll back for now?

fjetter:
@phofl why not just merge your fixes in dask-expr?

phofl:
My fix removes the filesystem from our hashing logic, resulting in ugly patches in a few places. This was never meant to get merged, rather to try it out.

We have to implement a tokenize logic for Arrow Filesystems. This might be trivial, I don't know that

hendrikmakait:
@phofl: What branch are you talking about here? There's https://github.com/dask/distributed/issues/8301, for which I'd need the TPC-H tests to work to dig deeper into it.

phofl:
https://github.com/phofl/dask-expr/tree/testing

mrocklin:
> @mrocklin arrow file system doesn't work with dask-expr main. I am guessing that you are still using my hacky branch that avoid the error? Should we invest effort into making this work or roll back for now?

I think I'm no longer using arrow filesystem anyway (although I would like to)

> We have to implement a tokenize logic for Arrow Filesystems

We don't need to do anything smart here I think.  It's ok for us to be dumb I think?  We might not get good reuse, but that's not a big deal.  Or maybe I don't understand the situation.  Is there somewhere I can learn more?

mrocklin:
@hendrikmakait you can apply this diff (I'll fix this shortly)

```diff
diff --git a/tests/tpch/conftest.py b/tests/tpch/conftest.py
index d174a78..07a9237 100644
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -291,6 +291,7 @@ def fs(local):
     if local:
         return None
     else:
+        return None
         import boto3
         from pyarrow.fs import S3FileSystem
```

PR #1152, opened 2023-10-26
Only test certain modules

mrocklin:
Others like tpch/{spark,polars,duckdb} don't benefit from being regularly tested or like runtime, we don't care about.

Originally I was going to add a `pytest.mark.skip` but then I realized that that would make it hard to explicitly run certain tests.  I think that instead we want to make CI more specific about what it cares about.

### DIFF
--- a/ci/scripts/run_tests.sh
+++ b/ci/scripts/run_tests.sh
@@ -10,4 +10,4 @@ then
   EXTRA_OPTIONS="$EXTRA_OPTIONS --benchmark"
 fi
 
-python -m pytest $EXTRA_OPTIONS $@
+python -m pytest $EXTRA_OPTIONS $@ tests/{benchmarks,stability,workflows,tpch/test_dask.py}


ci/scripts/run_tests.sh
```diff
@@ -10,4 +10,4 @@ then
   EXTRA_OPTIONS="$EXTRA_OPTIONS --benchmark"
 fi
 
-python -m pytest $EXTRA_OPTIONS $@
+python -m pytest $EXTRA_OPTIONS $@ tests/{benchmarks,stability,workflows,tpch/test_dask.py}
```
milesgranger, line 13: Won't this always ignore the scheduled runs of `--tpch-non-dask` then?
https://github.com/coiled/benchmarks/blob/158b07cb0dd2e85d86d31896125351ef9b75cc90/.github/workflows/tests.yml#L110

mrocklin: I don't have any desire to run non-Dask TPC-H benchmarks on a schedule.  Is there some motivation for this?  Or is this just something we're doing because historically we tend to run things on a daily basis?

If so, maybe that makes sense for projects that are included in `git-tip` because the code used in the benchmarks changes.  However, for these projects their software is pinned.  I see no reason to run them on any schedule except manually.  

Open to disagreement though if people have other thoughts.

mrocklin: Understood.  I disagree with that comment.  

@fjetter any objection to not running these benchmarks on a regular basis?

PR #1161, opened 2023-10-27
Align TPC-H benchmarks better with their definition

phofl:
Count is the wrong function for these operations.

According to the spec, we want to get the length of every group. Count does count the rows in every group, but it ignores missing values. The results are the same because the column that we apply count to doesn't have missing values, but this requires knowledge of the data which is not in scope for TPC-H. The thing we actually want is size, which returns the size of every group irrespective of missing values


milesgranger (approved):
LGTM, assuming tests are happy. :+1:

PR #1166, opened 2023-10-28
TPCH - Working Branch

mrocklin:
This probably shouldn't be merged, I'm just pushing up what I'm working on for visibility

### DIFF
--- a/tests/tpch/generate-data.py
+++ b/tests/tpch/generate-data.py
@@ -1,4 +1,5 @@
 import functools
+import multiprocessing
 import pathlib
 import tempfile
 import warnings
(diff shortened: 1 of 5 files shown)


tests/tpch/generate-data.py
```diff
@@ -117,7 +125,6 @@ def _tpch_data_gen(
             f"""
             SET memory_limit='{psutil.virtual_memory().available // 2**30 }G';
             SET preserve_insertion_order=false;
-            SET threads TO 1;
```
milesgranger, line 120: Removing this upsets the target partition size, as the page size calculated to meet the partition size will be written out per thread.

mrocklin: Thanks. Resolved.

PR #1184, opened 2023-11-02
Fix local paths for TPC-H data

hendrikmakait:
The `dataset_path` fixture didn't fit the paths created by the data generation script. This PR fixes that.


milesgranger (approved):
Thank you!

PR #1198, opened 2023-11-13
Use Cluster.get_spark for pyspark TPC-H tests

### LINKED ISSUE #1193: PySpark tpch fail to connect
[Workflow Run URL](https://github.com/coiled/benchmarks/actions/runs/6818965990)

milesgranger:
Will fix #1193

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -155,14 +155,6 @@ def cluster_spec(scale):
         idle_timeout="1h",
         wait_for_workers=True,
         scheduler_vm_types=["m6i.xlarge"],
-        backend_options={
-            "ingress": [
-                {
-                    "ports": [443, 8786, 8787, 7077, 8080, 4040, 15002],
-                    "cidr": "0.0.0.0/0",
-                },
-            ],
-        },
     )
     if scale == 10:
         return {
@@ -260,10 +252,7 @@ def spark_setup(cluster, local):
 
         spark = SparkSession.builder.master("local[*]").getOrCreate()
     else:
-        from coiled.spark import get_spark
-
-        with cluster.get_client() as client:
-            spark = get_spark(client)
+        spark = cluster.get_spark()
 
     # warm start
     from pyspark.sql import Row


milesgranger:
Hmm, @ntabris trying to just use `get_spark_cluster` in https://github.com/coiled/benchmarks/pull/1198/commits/fe93b437ddf6e74f668b2f835ee3a7737fe19b61 gave [similar failures](https://github.com/coiled/benchmarks/actions/runs/6846508469/job/18613233367#step:12:1468)

Then for sanity, I tried opening the ports you suggested in https://github.com/coiled/benchmarks/pull/1198/commits/cd26d5b0bad7c91c6562b44ee99565bb652e0266 and that [is also failing.](https://github.com/coiled/benchmarks/actions/runs/6849007565/job/18620324696?pr=1198#step:12:1734)

(Apologies if I'm doing something dumb here.. )

ntabris:
@milesgranger the error previously (before this PR) was a connection error. The errors in this PR look like it's complaining about getting disconnected. Do we know that running spark code failed? Or could this be the known issue with very messy shutdown?

ntabris:
Hm, I'm trying (outside of test) and it does seem to be broken. I'll try some things.

ntabris:
> Hm, I'm trying (outside of test) and it does seem to be broken. I'll try some things.

Oh, okay, so I tried `df = spark.read.parquet("s3a://coiled-data/uber")` very soon after calling `cluster.get_spark()` and I got the error. But then after a short wait I tried again and it worked.

It's possible that before the change to make the connection secure, it would just wait till spark was up when trying to connect, while now it gets an error if spark isn't up yet.

How big of a concern is that?

milesgranger:
Ah, indeed. Okay, ya I was confused as I was staring at a notebook I had just used with the same logic and that worked.. I must have waited just long enough then.

I've added a simple loop to try/wait and seemed to work from local. May some similar logic should be in the coiled implementation?

ntabris:
sorry for the hassle, we're both figuring this out as we go, but here's a PR that (once deployed) would simplify what you need to do in tests: https://github.com/coiled/platform/pull/3848

specifically, we'd open required ports by default, so you'd use `coiled.Cluster(...)` like you normally do, and when you call `cluster.get_spark()` it would wait until spark is ready.

milesgranger:
Great, let's wait for that then! :+1:

ntabris (approved):
LGTM, thanks for patience as the UX around this has been changing.

PR #1220, opened 2023-12-11
General dependencies bump

crusaderky:

PR #1223, opened 2023-12-12
Move aside TPCH non-dask dependencies

crusaderky:
Non-dask TPCH tests have substantial dependencies that can be very fiddly, and are typically excluded from PRs and A/B tests. Move them out of the way in order to slim down the general environment file and possibly for faster deployment (a crude spot observation shows cluster startup time has gone down from 300s to 230s).

- Blocked by and incorporates #1220


milesgranger (approved):
Good stuff, but maybe we ought to update the README as well, people wanting to run TPC-H?

crusaderky:
Updated README

PR #1236, opened 2023-12-14
More rechunking tests

crusaderky:
Add important use cases to the array rechunk test suite.
- XREF https://github.com/dask/distributed/pull/8330


hendrikmakait (approved):
Thanks, @crusaderky!

PR #1259, opened 2024-01-23
[TPC-H] Dask query 11

milesgranger:

Works with `dask.dataframe`  and dask-expr

### DIFF
--- a/tests/tpch/test_dask.py
+++ b/tests/tpch/test_dask.py
@@ -488,9 +488,6 @@ def test_query_11(client, dataset_path, fs):
         order by
             value desc
     """
-    # Note this query is a bit strange, the HAVING clause is pointless.
-    # It's comparing a sum to the same sum multiplied by 0.0001 so will always be true.
-    # Assuming it's there as part of the benchmark test we'll keep it.
     partsupp = dd.read_parquet(dataset_path + "partsupp", filesystem=fs)
     supplier = dd.read_parquet(dataset_path + "supplier", filesystem=fs)
     nation = dd.read_parquet(dataset_path + "nation", filesystem=fs)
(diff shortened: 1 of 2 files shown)


tests/tpch/test_dask.py
```diff
@@ -306,3 +306,61 @@ def test_query_7(client, dataset_path, fs):
...
+                ps_suppkey = s_suppkey
+                and s_nationkey = n_nationkey
+                and n_name = 'GERMANY'
+            )
+        order by
+            value desc
+    """
+    # Note this query is a bit strange, the HAVING clause is pointless.
+    # It's comparing a sum to the same sum multiplied by 0.0001 so will always be true.
+    # Assuming it's there as part of the benchmark test we'll keep it.
```
hendrikmakait, line 344: IIUC, the `HAVING` clause is not pointless. 

The business question of Q11 is:
> The Important Stock Identification Query finds, from scanning the available stock of suppliers in a given nation, all the parts that represent a significant percentage of the total value of all available parts. The query displays the part number and the value of those parts in descending order of value.
> 
> (https://www.tpc.org/TPC_Documents_Current_Versions/pdf/TPC-H_v3.0.1.pdf)

That is, the left side of the comparison of the HAVING clause is the value calculated per part

```sql
select
    ps_partkey,
    sum(ps_supplycost * ps_availqty) AS comparison_value
from
    partsupp,
    supplier,
    nation
where
    ps_suppkey = s_suppkey
    and s_nationkey = n_nationkey
    and n_name = 'GERMANY'
group by
    ps_partkey  
```

and the right side is a small fraction of the total value.

```sql
select
    sum(ps_supplycost * ps_availqty) * 0.0001 AS comparison_value
from
    partsupp,
    supplier,
    nation
where
    ps_suppkey = s_suppkey
    and s_nationkey = n_nationkey
    and n_name = 'GERMANY'
```

hendrikmakait (changes requested):
The current implementation is wrong and needs another pass.

tests/tpch/test_dask.py
milesgranger: Ah, quite right. I missed the obvious grouping, thanks! https://github.com/coiled/benchmarks/pull/1259/commits/e1bf55795aebd6df56ae6e7ff8c151953428a7d0

milesgranger:
@hendrikmakait I think this is good to go, failures are not related.

hendrikmakait (approved):
Thanks, @milesgranger! CI doesn't appear to be very happy with the benchmark query, but it's generally unhappy right now and the returned result is correct, so I'm inclined to merge and address this in a follow-up.

PR #1262, opened 2024-01-24
[TPC-H] Dask query 12

milesgranger:

This works for both `dask.dataframe` and dask-expr.


hendrikmakait (approved):
Thanks, @milesgranger!

PR #1264, opened 2024-01-24
[TPC-H] Dask query 14

milesgranger:

This works for both dask.dataframe and dask-expr.

### DIFF
--- a/tests/tpch/test_dask.py
+++ b/tests/tpch/test_dask.py
@@ -567,3 +567,51 @@ def test_query_13(client, dataset_path, fs):
         .sort_values(by=["custdist", "c_count"], ascending=[False, False])
         .compute()
     )
+
+
+@pytest.mark.shuffle_p2p
+def test_query_14(client, dataset_path, fs):
+    """
+    select
+        round(100.00 * sum(case
+            when p_type like 'PROMO%'
+                then l_extendedprice * (1 - l_discount)
+            else 0
+        end) / sum(l_extendedprice * (1 - l_discount)), 2) as promo_revenue
+    from
+        lineitem,
+        part
+    where
+        l_partkey = p_partkey
+        and l_shipdate >= date '1995-09-01'
+        and l_shipdate < date '1995-09-01' + interval '1' month
+    """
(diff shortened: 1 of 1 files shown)


tests/tpch/test_dask.py
```diff
@@ -306,3 +306,49 @@ def test_query_7(client, dataset_path, fs):
...
+    where
+        l_partkey = p_partkey
+        and l_shipdate >= date '1995-09-01'
+        and l_shipdate < date '1995-09-01' + interval '1' month
+    """
+    lineitem = dd.read_parquet(dataset_path + "lineitem", filesystem=fs)
+    part = dd.read_parquet(dataset_path + "part", filesystem=fs)
+
+    shipdate_from = datetime.strptime("1995-09-01", "%Y-%m-%d")
+    shipdate_to = shipdate_from + timedelta(days=365 / 12)
```
hendrikmakait, line 332:
```suggestion
    shipdate_to = datetime.strptime("1995-10-01", "%Y-%m-%d")
```

tests/tpch/test_dask.py
```diff
@@ -306,3 +306,49 @@ def test_query_7(client, dataset_path, fs):
...
+
+    # Promo revenue by line; CASE clause
+    table["promo_revenue"] = 0.00
+    mask = table.p_type.str.match("PROMO*")
+    table["promo_revenue"] = table.promo_revenue.where(
+        mask, table.l_extendedprice * (1 - table.l_discount)
+    )
+
+    # aggregate promo revenue calculation
+    _ = round(
```
hendrikmakait, line 347: nit: Can we move the `round` into the computation?

hendrikmakait (changes requested):
One fix-up needed for datetime arithmetics.

tests/tpch/test_dask.py
milesgranger: the `dd.Scalar<...>` has doesn't have .round method, but could do with using to_delayed the computing. :+1:

hendrikmakait (approved):
Thanks, @milesgranger!

PR #1265, opened 2024-01-24
[TPC-H] Dask query 15

milesgranger:

This works for both dask.dataframe and dask-expr.


tests/tpch/test_dask.py
```diff
@@ -306,3 +306,72 @@ def test_query_7(client, dataset_path, fs):
...
+        lineitem.groupby("l_suppkey")
+        .revenue.sum()
+        .to_frame()
+        .reset_index()
+        .rename(columns={"revenue": "total_revenue", "l_suppkey": "supplier_no"})
+    )
+
+    # Query
+    table = supplier.merge(
+        revenue, left_on="s_suppkey", right_on="supplier_no", how="left"
```
fjetter, line 369: I believe

```sql
from
        supplier,
        revenue
    where
        s_suppkey = supplier_no
```
is supposed to be an inner join

milesgranger: Fixed! Thanks! https://github.com/coiled/benchmarks/pull/1265/commits/9d984070a65d259087e70397ed07fb7c1bc23645

hendrikmakait (changes requested):

hendrikmakait:
CI failure for `test_query_15`: `ValueError: Not all divisions are known, can't align partitions. Please use `set_index` to set the index.`

Interestingly, this seems to work for a smaller scale factor (e.g., in `test_dask_results`)

milesgranger:
@hendrikmakait, propose we xfail it then. Opened an issue https://github.com/dask-contrib/dask-expr/issues/812 in the meantime.

phofl:
dask-expr should work now, the error is gone at least

hendrikmakait (approved):
Thanks, @milesgranger for implementing! @phofl, thanks for the fix, I've removed the `xfail` and the query seems to work.

PR #1266, opened 2024-01-24
[TPC-H] Dask query 16

milesgranger:

Works with dask.dataframe and dask-expr


tests/tpch/test_dask.py
```diff
@@ -519,3 +519,81 @@ def test_query_12(client, dataset_path, fs):
...
+        & (~table.p_type.str.match("MEDIUM POLISHED*"))
+        & (table.p_size.isin((49, 14, 23, 45, 19, 3, 36, 9)))
+        & (
+            # Need to compute, otherwise get NotImplementedErrror
+            # this works for dask.dataframe, fails for dask-expr w/
+            # RuntimeError: Sequence [('dtype_backend', None)] cannot be deterministically hashed.
+            ~table.ps_suppkey.isin(
+                supplier[
+                    supplier.s_comment.str.match("*Customer*Complaints*")
+                ].s_suppkey.compute()
```
hendrikmakait, line 584: If this works for dask.dataframe without calling compute, I'd suggest to remove the compute, XFAIL the test and file an issue for `dask-expr`.

milesgranger: Already xfailed.. it needs a compute, otherwise it fails in both with NotImplementedError. 
https://github.com/dask-contrib/dask-expr/issues/805

milesgranger: actually, did a small refactor and it works for both dask.dataframe and dask-expr again.

hendrikmakait:
The result is probably incorrect because of https://github.com/dask-contrib/dask-expr/pull/811.

milesgranger:
> The result is probably incorrect because of https://github.com/dask-contrib/dask-expr/pull/811.

Ah nope, it was wrong b/c of two incorrect things, misuse of `str.match` and doing `count` instead of `nunique`. Should be good to go now.

tests/tpch/test_dask.py
```diff
@@ -785,6 +785,71 @@ def test_query_15(client, dataset_path, fs):
...
+        p_type,
+        p_size
+    """
+    partsupp = dd.read_parquet(dataset_path + "partsupp", filesystem=fs)
+    part = dd.read_parquet(dataset_path + "part", filesystem=fs)
+    supplier = dd.read_parquet(dataset_path + "supplier", filesystem=fs)
+
+    supplier["is_complaint"] = supplier.s_comment.str.contains(
+        "Customer"
+    ) & supplier.s_comment.str.contains("Complaints")
```
hendrikmakait, line 828: This is slightly incorrect, it could return `True` for `Complaints Customer`, while the SQL would be `False` 

```suggestion
    supplier["is_complaint"] = supplier.s_comment.str.contains(
        "Customer.*Complaints"
    )
```

hendrikmakait (approved):
Thanks, @milesgranger! I'll open a follow-up for the slight incorrectness.

PR #1270, opened 2024-01-24
Add remaining DuckDB queries

hendrikmakait:


milesgranger (commented):
Looks good to me, except tests aren't happy with q15: https://github.com/coiled/benchmarks/actions/runs/7639050133/job/20811215626?pr=1270#step:12:288

Maybe just add that as another table in the `with` statement?

tests/tpch/test_duckdb.py
```diff
@@ -312,3 +312,673 @@ def _():
...
+                    l_shipdate >= date '1996-01-01'
+                    and l_shipdate < date '1996-01-01' + interval '3' month
+                group by
+                    l_suppkey
+            """
+        )
+
+        connection.execute(
+            f"""
+            with supplier as (select * from read_parquet('{dataset_path}supplier/*.parquet'))
```
milesgranger, line 632: untested, but think something like this will work

```suggestion
            with lineitem as (select * from read_parquet('{dataset_path}lineitem/*.parquet')),
                     revenue as (
                         select
                             l_suppkey as supplier_no,
                             sum(l_extendedprice * (1 - l_discount)) as total_revenue
                         from
                             lineitem
                         where
                             l_shipdate >= date '1996-01-01'
                             and l_shipdate < date '1996-01-01' + interval '3' month
                         group by
                             l_suppkey
                     ),
                    supplier as (select * from read_parquet('{dataset_path}supplier/*.parquet'))
```

hendrikmakait: The spec wants to see a view, so I've just removed the with statement.

milesgranger (approved):
Awesome, thanks! :+1:

PR #1275, opened 2024-01-25
[TPC-H] xfail dask query 9

milesgranger:
Follow up to #1257, and part of https://github.com/coiled/benchmarks/issues/1274


milesgranger:
Closing in favor of https://github.com/coiled/benchmarks/pull/1277 which refactors to avoid the recursion error; still interesting that this first iteration caused it in dask-expr and not in dask.dataframe though.

PR #1277, opened 2024-01-25
[TPC-H] Refactor query 9 to avoid recursion error in dask-expr

milesgranger:
Follow up to https://github.com/coiled/benchmarks/pull/1257, and part of https://github.com/coiled/benchmarks/issues/1274

Refactors query 9 to avoid the recursion error encountered w/ dask-expr.


hendrikmakait (approved):
Thanks, @milesgranger!

tests/tpch/test_dask.py
```diff
@@ -382,23 +381,26 @@ def test_query_9(client, dataset_path, fs):
         )
         .merge(orders, left_on="l_orderkey", right_on="o_orderkey", how="inner")
         .merge(nation, left_on="s_nationkey", right_on="n_nationkey", how="inner")
+        .assign(part_contains_green=lambda df: df.p_name.str.match("*green*"))
```
phofl, line 384: Can we not do this? lambdas are terrible for optimisation

Same for query, let's rather use an explicit check

phofl: You can just create new objects, no need to chain everything together

hendrikmakait: From what I understand, we have to address https://github.com/dask-contrib/dask-expr/issues/807 to avoid the lambda.

milesgranger: > You can just create new objects, no need to chain everything together

Ya.. but, I like it. :sweat_smile: Will avoid doing so if that's the group's preference of course. :heavy_check_mark:

hendrikmakait: FWIW, I also like chaining :)

phofl: I agree that chaining is nice, but not if your performance sucks with it. We can chain when we have the Col expression.

PR #1278, opened 2024-01-25
Verify TPC-H results for Dask

### LINKED ISSUE #1276: [TPC-H] Query results are never validated
While reviewing the implementation of the remaining TPC-H queries, I've identified multiple subtle mistakes. This is a cumbersome process that doesn't guarantee that every subtlety is caught. We should add tests that assert the correctness of our queries. DuckDB contains all the data we need for that: https://duckdb.org/docs/extensions/tpch.html#listing-expected-answers

hendrikmakait:
Closes #1276


tests/tpch/test_correctness.py
```diff
@@ -0,0 +1,76 @@
...
+def answers_path(tmp_path_factory):
+    path = tmp_path_factory.mktemp("answers")
+    return generate_answers(base_path=path)
+
+
+@pytest.fixture(scope="module")
+def data_path(tmp_path_factory):
+    path = tmp_path_factory.mktemp("data")
+    scale = VERIFICATION_SCALE
+    return pathlib.Path(generate_data(scale=scale, path=str(path), relaxed_schema=True))
```
hendrikmakait, line 25: Currently, we generate answers and data on the fly. We can adjust this if it strains the CI job, but right now it only adds ~15s overhead locally.

tests/tpch/test_dask.py
```diff
@@ -209,7 +224,9 @@ def test_query_6(client, dataset_path, fs):
     )
 
     flineitem = line_item_ds[sel]
-    (flineitem.l_extendedprice * flineitem.l_discount).sum().compute()
+    result = (flineitem.l_extendedprice * flineitem.l_discount).sum().compute()
+
+    return pd.DataFrame({"revenue": [result]})
```
hendrikmakait, line 229: The dask/pandas API returns a scalar as a result. The TPC-H specs expect a table, so wrapping this in a DataFrame adheres to that spec and makes testing easier while it shouldn't impact the test much.

tests/tpch/test_dask.py
```diff
@@ -226,7 +243,7 @@ def test_query_7(client, dataset_path, fs):
     lineitem_filtered = line_item_ds[
         (line_item_ds["l_shipdate"] >= var1) & (line_item_ds["l_shipdate"] < var2)
     ]
-    lineitem_filtered["l_year"] = 1  # lineitem_filtered["l_shipdate"].dt.year
```
hendrikmakait, line 229: I assume this had been commented out because it had not been implemented yet.

fjetter (approved):
LGTM assuming CI is OK

hendrikmakait:
CI is red, but the issues are already present on main (https://github.com/coiled/benchmarks/actions/runs/7662098986/job/20882751097#step:12:1713).

PR #1289, opened 2024-01-26
Avoid shuffle error in Q10

hendrikmakait:
Part of #1274

The result is still wrong but that's a separate issue.


milesgranger (approved):
LGTM!

PR #1299, opened 2024-01-29
Revert "[TPC-H] Refactor query 9 to avoid recursion error in dask-expr"

phofl:
Reverts coiled/benchmarks#1277

can not reproduce locally

### DIFF
--- a/tests/tpch/test_correctness.py
+++ b/tests/tpch/test_correctness.py
@@ -78,7 +78,7 @@ def verify_result(result: pd.DataFrame, query: int, answer_dir: pathlib.Path):
         6,
         7,
         pytest.param(8, marks=pytest.mark.skip(reason="Not implemented")),
-        9,
+        pytest.param(9, marks=pytest.mark.xfail(reason="Result is wrong")),
         pytest.param(10, marks=pytest.mark.xfail(reason="Result is wrong")),
         11,
         12,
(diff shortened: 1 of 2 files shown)


milesgranger:
> can not reproduce locally

Weird, I can still produce it with the latest release (0.3.5) on this PR (dask 2024.1.1)

phofl:
Using lambdas is inherently bad in systems that can be optimised, so we should start warning users about this at some point when we have column expressions. There is not much point in benchmarking this and shooting ourselves in the foot on purpose while doing so. You are shuffling and reading the whole dataset which is just incredibly slow, which explains the timeouts

I was doing something dumb, which hid the error from me, sorry.

phofl:
@milesgranger Could you push this over the finish line? it looks like the query is schematically off

milesgranger:
@phofl ready now, failures are unrelated.

>FAILED tests/tpch/test_dask.py::test_query_10 - RuntimeError: shuffle_barrier failed during shuffle 9a0b09c0c0f40cd95e7fa7560513bb54
ERROR tests/tpch/test_dask.py::test_query_11 - asyncio.exceptions.TimeoutError: 1/16 nanny worker(s) did not shut down within 120s

hendrikmakait (approved):
Thanks!

PR #1309, opened 2024-01-30
[TPC-H] Avoid shuffle barrier RuntimeError in q10

milesgranger:


milesgranger:
@hendrikmakait, seems q9 is still a problematic one. I'll poke at it and worse case xfail. This seems to alleviate the barrier error though for q10.

hendrikmakait (approved):
Thanks, @milesgranger!

PR #1318, opened 2024-01-31
xfail PyTorch Optuna workflow on TimeoutError

milesgranger:
Will close https://github.com/coiled/benchmarks/issues/1261

_"90% of the time, it works every time."_ :wink: 

Mostly works on CI, but occasionally fails because the model doesn't manage to detect/use CUDA. IMO I don't think it's worthwhile to devote much time debugging why this is the case.


hendrikmakait (approved):
I'm mildly concerned that it fails to interact with CUDA but this works for me for now.

PR #1319, opened 2024-01-31
Fix Q16

### LINKED ISSUE #1303: [TPC-H] Query 16  is slightly incorrect
https://github.com/coiled/benchmarks/blob/8c9db008242ce9dff52d805558d2482cbbf7b913/tests/tpch/test_dask.py#L826-L828

is slightly incorrect, it could return `True` for `Complaints Customer`, while the SQL would be `False`. The following should be correct:

```suggestion
    supplier["is_complaint"] = supplier.s_comment.str.contains(
        "Customer.*Complaints"
    )
```

_Originally posted by @hendrikmakait in https://github.com/coiled/benchmarks/pull/1266#discussion_r1470678186_

hendrikmakait:
Closes #1303

### DIFF
--- a/tests/tpch/test_dask.py
+++ b/tests/tpch/test_dask.py
@@ -887,7 +887,7 @@ def test_query_16(client, dataset_path, fs):
     table = partsupp.merge(part, left_on="ps_partkey", right_on="p_partkey")
     table = table[
         (table.p_brand != "Brand#45")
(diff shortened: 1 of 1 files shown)


milesgranger (approved):
Assuming nothing surprising in CI, it looks good to me!

PR #1324, opened 2024-01-31
[TPC-H] Move Dask queries to their own file

hendrikmakait:
This PR refactors the TPC-H tests so that we have functions that provide us with the queries in another file. 

This is 99.9% verbatim copy+paste, I don't touch the queries themselves.


hendrikmakait:
This is a step toward #1325

milesgranger (approved):
Thanks for the explanation, seems reasonable to me. :+1:

PR #1326, opened 2024-01-31
Benchmark plan generation and optimization on TPC-H queries

### LINKED ISSUE #1325: Benchmark plan generation and optimization in `dask-expr`
We've recently hit a regression in the optimization runtime in `dask-expr` (https://github.com/dask-contrib/dask-expr/issues/796). We should build a benchmark for the runtime of plan generation and optimization. The TPC-H queries look like a good subject for that test since they are somewhat complex.

hendrikmakait:
Closes #1325 

Blocked by and includes #1324

### DIFF
--- a/tests/tpch/test_optimization.py
+++ b/tests/tpch/test_optimization.py
@@ -1,0 +1,56 @@
+import pytest
+from distributed import LocalCluster
+
+pytestmark = pytest.mark.tpch_dask
+
+
+@pytest.fixture(scope="module")
+def cluster():
+    with LocalCluster() as cluster:
+        yield cluster
+
+
+@pytest.fixture(scope="module")
+def client(cluster, restart, benchmark_time):
+    with cluster.get_client() as client:
+        if restart:
+            client.restart()
+        client.run(lambda: None)
+
+        with benchmark_time:
+            yield client
+
+
+@pytest.mark.parametrize(
+    "query",
+    [
+        1,
+        2,
+        3,
+        4,
+        5,
+        6,
+        7,
+        8,
+        9,
+        10,
+        11,
+        12,
+        13,
+        14,
+        15,
+        16,
+        17,
+        18,
+        19,
+        20,
+        21,
+        22,
+    ],
+)
+def test_optimization(query, dataset_path, fs, client):
+    from . import dask_queries
+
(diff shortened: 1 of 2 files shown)


hendrikmakait:
Benchmark results look odd given that Q18 takes 2.0 - 2.5 seconds to complete and 80 seconds to optimize.

Screenshot 2024-02-01 at 10 44 31

hendrikmakait:
This looks better:

Screenshot 2024-02-01 at 12 26 08

tests/tpch/test_optimization.py
```diff
@@ -0,0 +1,39 @@
...
+        20,
+        21,
+        22,
+    ],
+)
+def test_optimization(query, dataset_path, fs, client):
+    func = getattr(dask_queries, f"query_{query}")
+    result = func(dataset_path, fs)
+    # We need to inject .repartition(npartitions=1) which .compute() does under the hood
+    result.repartition(npartitions=1).optimize()
```
fjetter, line 39: by default, this is using a coiled Cluster. Is this intentional? This makes running this test in isolation really expensive

hendrikmakait: > by default, this is using a coiled Cluster. Is this intentional? This makes running this test in isolation really expensive

This strictly follows the deployment pattern of the actual TPC-H benchmarks to account for necessary pre-computations. This allows us to match the measured performance to the numbers we retrieve for the total computation. You should be able to run this in isolation with `--local` without any Coiled infrastructure. However, this would then also read local data which will skew the measurements (but should be fine for CPU-based stuff).

hendrikmakait: > Benchmark results look odd given that Q18 takes 22 - 25 seconds to complete and 80 seconds to optimize.
> 
> Screenshot 2024-02-01 at 10 44 31
> EDIT: This may have been caused by me using a `LocalCluster`, the query sets an index and requires some pre-computation.

This shows what happens when running this without a Coiled cluster but using cloud-based data. For now, I'm inclined to keep it this way and investigate the benchmarked queries first. There are already suggestions on how to reduce the amount of pre-computation, so we adjusting this might introduce less skew in the future.

phofl (approved):
lgtm, merge when ready

PR #1332, opened 2024-02-01
Don't overwrite `benchmark.db` in Action matrix

hendrikmakait:
This PR fixes a bug where matrix elements with the same `os` and `python-version` would generate the same `benchmark.db` with the last uploaded version winning. This explains why Snowflake data is mostly missing.


milesgranger (approved):
Nice, thanks @hendrikmakait

PR #1333, opened 2024-02-01
Limit execution of TPC-H non-Dask benchmarks

hendrikmakait:
We thought that the existing logic worked, but it turns out that we were just running Dask-based TPC-H queries once as part of  `Tests / tests` and then running _all_ TPC-H queries as part of the `Tests / tests/tpch -m tpch_nondask` run and discarding the results of the latter run due to #1332.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -111,10 +111,12 @@ jobs:
 
       - name: Disable non-Dask TPCH benchmarks on most PRs and on daily schedule (except Sundays)
         if: |
-          ${{ !(
-            contains(matrix.pytest_args, 'tpch_nondask')
-            || (github.event_name == 'schedule' || github.event.schedule == "1 0 * * 0")
-            || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'tpch'))
+          ${{ 
+            !contains(matrix.pytest_args, 'tpch_nondask')
+            || !(
+              (github.event_name == 'schedule' || github.event.schedule == "1 0 * * 0")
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -108,10 +109,15 @@ jobs:
         run: |
           echo PYTEST_MARKERS=" and not workflows" >> $GITHUB_ENV
 
-      - name: Disable non-Dask TPCH benchmarks on most PRs
+      - name: Disable non-Dask TPCH benchmarks on most PRs and on daily schedule (except Sundays)
         if: |
-          github.event_name != 'schedule'
-          && !(github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'tpch'))
+          ${{ 
+            !contains(matrix.pytest_args, 'tpch_nondask')
```
hendrikmakait, line 115: Skip if it's not `tpch_nondask`

.github/workflows/tests.yml
```diff
@@ -108,10 +109,15 @@ jobs:
...
+            || !(
+              (github.event_name == 'schedule' || github.event.schedule == "1 0 * * 0")
+              || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'tpch'))
+            )
```
hendrikmakait, line 119: OR skip if it's neither the scheduled run on Sundays nor requested via label.

hendrikmakait:
This should probably be merged after #1332

milesgranger (approved):
Okay for me, thanks!

PR #1335, opened 2024-02-02
[TPC-H] Simplify Dask queries and avoid `.query` and `.apply` 

hendrikmakait:


tests/tpch/dask_queries.py
```diff
@@ -839,6 +841,7 @@ def query_16(dataset_path, fs):
     supplier = dd.read_parquet(dataset_path + "supplier", filesystem=fs)
 
     supplier["is_complaint"] = supplier.s_comment.str.contains("Customer.*Complaints")
+    # FIXME: We have to compute this early because passing a `dask_expr.Series` to `isin` is not supported
```
hendrikmakait, line 844: XREF: https://github.com/dask-contrib/dask-expr/issues/834

hendrikmakait:
Honestly, I'm pretty confused by CI, but I don't touch any of these tests in here, so this appears to be unrelated.

tests/tpch/dask_queries.py
```diff
@@ -570,19 +571,20 @@ def query_11(dataset_path, fs):
...
+    threshold = (joined.ps_supplycost * joined.ps_availqty).sum() * 0.0001
 
-    def calc_value(df):
-        return (df.ps_supplycost * df.ps_availqty).sum().round(2)
+    joined["value"] = joined.ps_supplycost * joined.ps_availqty
 
-    return (
-        joined.groupby("ps_partkey")
-        .apply(calc_value, meta=("value", "f8"))
+    res = joined.groupby("ps_partkey").value.sum()
```
phofl, line 578: .value is dangerous, because pandas has attributes that are very very similar which would take precedence, I would prefer square brackets here or a different name

milesgranger (approved):
Ya noticed the int errors in the other PR. Agree unreleated.

phofl (approved):
small comments, feel free to tackle in a follow up

PR #1338, opened 2024-02-02
[TPC-H] Don't return from tests

hendrikmakait:
Should address https://github.com/coiled/benchmarks/issues/1337#issuecomment-1923757050


milesgranger (approved):
Thanks!

PR #1339, opened 2024-02-02
Fix CI failure

hendrikmakait:
Fixes `ERROR tests/tpch/test_dask.py::test_query_22 - ValueError: invalid literal for int() with base 10: 'optimization[1]'`


milesgranger (approved):
Thanks @hendrikmakait

PR #1349, opened 2024-02-05
Fix running non-Dask TPC-H tests

hendrikmakait:


.github/workflows/tests.yml
```diff
@@ -8,7 +8,7 @@ on:
   pull_request:
   schedule:
     # Runs "At 00:01" (see https://crontab.guru)
-    - cron: "1 0 * * 0" # every Sunday (relevant for non-Dask TPC-H benchmarks) 
+    - cron: "1 * * * *" # every Sunday (relevant for non-Dask TPC-H benchmarks)
```
hendrikmakait, line 11: This will be reverted in a follow-up, I just want to be able to observe this on `main`.

phofl (approved):

PR #1356, opened 2024-02-05
TPC-H data generation script fixup

jrbourbeau:
I was getting errors running `python tests/tpch/generate_data.py --scale 1` locally (one from `click.Choice` not liking a generator as input, and another due to compression name uppercase/lowercase mismatch). This PR fixes things for me locally.


milesgranger (approved):
Thanks, @jrbourbeau!

PR #1359, opened 2024-02-06
xfail TPC-H Dask query 21 for shuffle_barrier

milesgranger:
Part of #1316


phofl:
I have a pr https://github.com/dask-contrib/dask-expr/pull/852 that should fix this, so please hold off on merging

PR #1385, opened 2024-02-13
Fix python-version argument for benchmarks

### LINKED ISSUE #1383: Dashboard generation is broken
We don't show any data after Feb 05, 2024. From what I see, it's still included in the generated `benchmark.db`, but somehow not picked up.

hendrikmakait:
Closes #1383 
Follow-up to https://github.com/coiled/benchmarks/pull/1349 which accidentally caused all tests to be run on Python 3.11. 

FWIW, we didn't lose data but accidentally ran tests on 3.11 and even generated dashboards for that: https://benchmarks.coiled.io/Python%203.11.html

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,7 +77,7 @@ jobs:
           miniforge-variant: Mambaforge
           use-mamba: true
           condarc-file: ci/condarc
-          python_version: ${{ matrix.python_version }}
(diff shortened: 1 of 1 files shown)


hendrikmakait:
We're back to having data for Python 3.9: 
Screenshot 2024-02-13 at 17 46 50

phofl (approved):

phofl:
thx

PR #1392, opened 2024-02-14
Refactor TPC-H generation script - consolidate small table partitions

### LINKED ISSUE #1386: Adjust TPC-H data generation script to create a single `.parquet` files for `nation` and `region`
As discussed in #1380, partitioning these two datasets is unrealistic for this particular workload.

As a workaround, I have manually replaced the partitioned data with unpartitioned data for the Arrow-based datasets. We need to adjust the data generation scripts to make the benchmarks reproducible for others (including our future selves).

milesgranger:
Will close #1386

It's made slightly more complex b/c duckdb doesn't support generating data for a single table. So we can keep things mostly the same but just not write out the partitions but collect and combine them at the end of the normal partition generation.

Also made some other adjustments small refactoring, including using a localcluster to generate the data locally to more closely replicate the behavior when using Coiled. Plus it goes faster generating larger scales. :)

### DIFF
--- a/tests/tpch/generate_data.py
+++ b/tests/tpch/generate_data.py
@@ -2,8 +2,11 @@ import enum
 import functools
 import pathlib
 import tempfile
+import time
 import uuid
 import warnings
+from collections import namedtuple
+from typing import Union
 
 import boto3
 import botocore.session
@@ -12,10 +15,13 @@ import coiled
 import dask
 import duckdb
 import psutil
+import pyarrow as pa
 import pyarrow.compute as pc
 import pyarrow.parquet as pq
 
 REGION = None
+SMALL_TABLES = {"region", "nation"}
+SmallTable = namedtuple("SmallTable", ("name", "out_dir", "data"))
 
 
 class CompressionCodec(enum.Enum):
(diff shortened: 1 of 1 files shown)


milesgranger:
Run on coiled, saw some workers being killed for memory so adjusted worker sizing a bit.
Confirm it works there too otherwise, scale 100 w/ single partition of nation and region at s3://coiled-runtime-ci/milesg/tpch/scale-100

tests/tpch/generate_data.py
```diff
@@ -60,20 +66,75 @@ def generate(
...
+def collect_small_tables(
+    tables: list[dict[str, SmallTable]], compression: CompressionCodec
+):
+    table_names = {t for d in tables for t in d}
+    for table in table_names:
+        small_tables: list[SmallTable] = [t[table] for t in tables if table in t]
+        if small_tables:
+            name = small_tables[0].name
+            out_dir = small_tables[0].out_dir
+            table = pa.concat_tables(t.data for t in small_tables)
```
hendrikmakait, line 108: Does this approach guarantee the correct ordering? I'm not sure it's relevant here but I'd like the data to look the same as if generated by a single process.

milesgranger: Before it 'worked' but I think it's mostly chance to that the futures returned in order and less likely at larger scales. Updated w/ https://github.com/coiled/benchmarks/pull/1392/commits/40e155563d9d01e587b35ffcc02ab5a7d1f7b6a8 to ensure they're returned in order.

hendrikmakait: Thanks!

hendrikmakait (approved):
Thanks, @milesgranger!

PR #1400, opened 2024-02-15
[TPC-H] Fix restarting for Polars and DuckDB

hendrikmakait:


tests/tpch/conftest.py
```diff
@@ -340,7 +340,7 @@ def warm_start(module_run, local):
 
 
 @pytest.fixture(scope="function")
-def run(module_run, restart, benchmark_time, warm_start, make_chart):
+def run(module_run, local, restart, benchmark_time, warm_start, make_chart):
     if restart and not local:
```
hendrikmakait, line 344: Without adding the fixture, `local` would be a function and always true-ish.

phofl (approved):

PR #1406, opened 2024-02-17
[TPC-H] Adopt Spark join ordering in query 18

hendrikmakait:
The join ordering we chose in query 18 is suboptimal and causes workers run OOM. This PR reimplements the query using Spark's join ordering. Note that Polars join ordering (https://github.com/coiled/benchmarks/commit/8eb24d068401269b665c485b27ee53d8b748da47) appears to be slightly faster (285s vs. 318s), but I do not want to optimize too much manually. Picking Spark's join ordering will provide the best apples-to-apples comparison against Dask's main competitor.

### DIFF
--- a/tests/tpch/dask_queries.py
+++ b/tests/tpch/dask_queries.py
@@ -945,7 +945,6 @@ def query_18(dataset_path, fs):
         o_orderdate
     limit 100
     """
-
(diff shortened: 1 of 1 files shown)


milesgranger (approved):

PR #1409, opened 2024-02-20
Enable dask-expr for benchmarks

### LINKED ISSUE #1260: Toggle dask-expr for all benchmarks
dask/dask CI looks good and we're preparing to switch over to dask-expr as the dataframe backend soon. This removes the need to watch "legacy dask" very closely. I don't expect anything major to slow down dask/dask (that wouldn't break dask-expr in a similar way).

We should set the dask-expr toggle for our benchmark suite to reflect this (possibly run one A/B test before we switch to verify our baseline)

phofl:
closes #1260


phofl:
Please don't merge yet, want to check that ci is green before running an ab test

PR #1416, opened 2024-02-26
Bump versions

hendrikmakait:


hendrikmakait:
CI failure on Windows is #1418

milesgranger:
I'd at least prefer we work out the Windows issue first so we don't have CI failing anymore than it does now.

milesgranger (approved):
Ah shoot, I see duckdb .10 is already in my mistake.

PR #1425, opened 2024-02-28
use pyarrow fs

fjetter:


tests/tpch/conftest.py
```diff
@@ -268,21 +268,19 @@ def fs(local):
...
+
+        session = boto3.session.Session()
+        credentials = session.get_credentials()
+
+        fs = S3FileSystem(
+            secret_key=credentials.secret_key,
+            access_key=credentials.access_key,
+            region="us-east-2",
+            session_token=credentials.token,
+        )
```
mrocklin, line 282: We should probably raise a request for this functionality to happen automatically in PyArrow.  Probably this is something that they should handle in the future. 

In the meantime, maybe we handle it in dask.dataframe?

fjetter: frankly, I haven't checked what the current behavior is but have been using this for testing. I agree that this should be handled automatically.

mrocklin: Yeah, mostly I'm saying it because I suspect that we'll want this and that it'll also probably take a while for it to be released in pyarrow, so sending an issue sooner might be good.

PR #1434, opened 2024-03-02
Avoid column projections in join/set_index tests

phofl:


tests/benchmarks/test_join.py
```diff
@@ -24,6 +24,8 @@ def test_join_big(small_client, memory_multiplier):
     df2_big = df2_big.astype({"predicate": "int"})
 
     join = df1_big.merge(df2_big, on="predicate", how="inner")
+    # Avoid column projections
+    join = join.map_partitions(lambda x: x)
```
fjetter, line 28: I think you should provide more information in code. Without context this doesn't make any sense

fjetter: Technically, this isn't even about column projection but rather about the size/len optimization, isn't it?

phofl: Yes that's true, needs more investigations before it's ready to merge, will hopefully get to this today

phofl:
depends on https://github.com/dask-contrib/dask-expr/pull/938

fjetter (approved):

PR #1436, opened 2024-03-05
A/B tests - calculate confidence interval against all combinations

fjetter:
When running multiple configurations it is sometimes useful to caculate the differences between the various configurations and not just the baseline.

I think this is useful and cheap enough for us to do every time.


phofl (approved):

phofl:
thanks

PR #1438, opened 2024-03-05
Measure computation submission delay

fjetter:
On top of the plain optimization, this measures the time it takes to submit the graph. This is typically the time it takes after hitting compute until the dashboard shows something.

This is likely skewed by caching as introduced in https://github.com/dask-contrib/dask-expr/pull/917 (the fsspec implementation also uses various caching layers) but I still believe this is added value


phofl (approved):

phofl:
thx

PR #1441, opened 2024-03-06
Add CI action to trigger TPC-H benchmarks

hendrikmakait:


.github/workflows/tpch.yml
```diff
@@ -0,0 +1,68 @@
...
+        DASK_COILED__TOKEN: ${{ secrets.COILED_BENCHMARK_BOT_TOKEN }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.RUNTIME_CI_BOT_AWS_ACCESS_KEY_ID }}
+        AWS_DEFAULT_REGION: us-east-2 # this is needed for boto for some reason
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.RUNTIME_CI_BOT_AWS_SECRET_ACCESS_KEY }}
+        COILED_RUNTIME_VERSION: ${{ matrix.runtime-version }}
+        DB_NAME: tpch_1000.db
+        CLUSTER_DUMP: always
+        DASK_DATAFRAME__QUERY_PLANNING: True
+      run: |
+        pytest --benchmark tests/tpch/test_dask.py tests/tpch/test_duckdb.py tests/tpch/test_pyspark.py -n 4 --dist loadscope --scale 1000
```
fjetter, line 59:
```suggestion
        pytest --benchmark \
            tests/tpch/test_dask.py \
            tests/tpch/test_duckdb.py \
            tests/tpch/test_pyspark.py \
            -n 4 --dist loadscope \
            --scale 1000
```

fjetter:
I guess your script is still a little messy but I suggest to commit that as well. This way we can all easily generate the plots and iterate on them

.github/workflows/tpch.yml
```diff
@@ -0,0 +1,68 @@
+name: TPC-H Scale 1000
+on:
+  pull_request:
```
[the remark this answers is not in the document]
fjetter: does it work as expected?

.github/workflows/tpch.yml
fjetter: makes things easier to modify/comment out sections in case we're not interested in e.g. spark or duckdb or want to change the scale

.github/workflows/tpch.yml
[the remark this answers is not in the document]
hendrikmakait: It does, I'm using it right now to generate the latest benchmark results. The only thing we need to deal with in a follow-up is the occasional flake by DuckDB.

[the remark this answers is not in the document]
hendrikmakait: Removed this.

hendrikmakait:
> I guess your script is still a little messy but I suggest to commit that as well. This way we can all easily generate the plots and iterate on them

Will do in a follow-up

.github/workflows/tpch.yml
```diff
@@ -0,0 +1,84 @@
+name: TPC-H Benchmarks
+on:
+  workflow_dispatch:
+    inputs:
+    scale:
```
fjetter, line 5: is this supposed to be indented?

hendrikmakait: good catch!

PR #1447, opened 2024-03-11
Milesgranger/1437 migrate ab test to pip for dependabot

### LINKED ISSUE #1437: migrate AB tests to smth compatible with depandabot
It happens rather frequently that I am running on rather old versions of dask/distributed in AB_environments. We're not doing a good job of updating those environments. I wanted to setup depandabot for this job but noticed that conda is not even supported.

I suggest to move off of conda for the A/B tests to smth else, e.g. pip / pip-compile that supports dependabot and let it update the dependencies at a regular cadence

milesgranger:
Will close #1437

- #1437 

Moves dependencies to new `AB_<name>.requirements.txt` file, which is dependabot friendly for automated upgrades. Alone w/ the associated `dependabot.yml` file configured to look in `AB_environments` directory.


milesgranger:
Successful A/B run here: https://github.com/coiled/benchmarks/actions/runs/8246636881

AB_environments/AB_baseline.requirements.txt
```diff
@@ -0,0 +1,44 @@
...
+scipy==1.11.4
+sqlalchemy==2.0.23
+pynvml==11.5.0
+bokeh==3.3.2
+gilknocker==0.4.1
+# End copy-paste
+
+# Make sure you install dask and distributed either both from pip or both from
+# conda. You may alternatively point to your own git fork (but make sure you
+# sync'ed tags!)
```
crusaderky, line 37: This comment looks obsolete?

AB_environments/AB_sample.requirements.txt
```diff
@@ -0,0 +1,44 @@
...
+scipy==1.11.4
+sqlalchemy==2.0.23
+pynvml==11.5.0
+bokeh==3.3.2
+gilknocker==0.4.1
+# End copy-paste
+
+# Make sure you install dask and distributed either both from pip or both from
+# conda. You may alternatively point to your own git fork (but make sure you
+# sync'ed tags!)
```
crusaderky, line 37: This comment looks obsolete?

AB_environments/AB_baseline.requirements.txt
milesgranger: https://github.com/coiled/benchmarks/pull/1447/commits/1d9a784f673b3b56a042fcbe92ff04db1df08569

crusaderky (changes requested):
As per https://github.com/coiled/benchmarks/issues/1437#issuecomment-1991548225, could you migrate the environment used for nighlty runs too please?

**Note:** expect pain with `environment-snowflake.yml` and `environment-tpch-nondask.yml`. The upstream maintainers tend to remain behind with releases and are very likely to break when auto-updated. You may want to consider excluding these two envs from dependabot.

milesgranger:
@crusaderky, it seemed to get unwieldly needing to copy around a core set of requirements to so many files when moving to pip. Therefore I've converted to pip-compile so we can include these in other environments, by just adding `-r requirements.in` or similar both in CI and AB environments. Then just two conda env files, one for the base to determine python version and stuff like openssl and openjdk which is easier w/ conda, then one for updating to git-tip.

Done in https://github.com/coiled/benchmarks/pull/1447/commits/d628ccee9253acb9f3bff8f332bf2643761a6c40, let me know what you think.

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+  - directory: ci/
+    package-ecosystem: pip
+    schedule:
+      interval: weekly
+    ignore:
+      # https://github.com/coiled/benchmarks/issues/1221
+      - dependency-name: pyspark
+        versions: ["1.5.x"]
```
milesgranger, line 10: I think this ought to be this, no?
```suggestion
      - dependency-name: pyspark
        versions: ["3.4.x"]
```

crusaderky: 3.5.X, but yes

crusaderky:
@milesgranger I've gone through all the documentation; please review.
If you're happy with it I think we can merge?

milesgranger (commented):
@crusaderky Just took one more pass, and I have nothing to add or change. Thanks for your help here!

crusaderky (approved):

PR #1448, opened 2024-03-11
Increase instance disk size for TPC-H SF 1000

hendrikmakait:
Spark recently ran out of space (https://github.com/coiled/benchmarks/actions/runs/8231277279/job/22506263406#step:8:745). We should bump the disk size.


milesgranger (approved):
Good for me, thanks!

Issue #1449, opened 2024-03-11
[TPC-H] PySpark workloads are vulnerable to undetected worker loss


hendrikmakait:
In this run, the PySpark cluster lost three workers over time without us ever detecting let alone healing this. Things went wrong for PySpark and likely skewed test results afterward.

At the very least, we should be able to assert that a PySpark cluster still has the requested # of workers when beginning a new query.

Cluster: https://cloud.coiled.io/clusters/410048/
CI Run: https://github.com/coiled/benchmarks/actions/runs/8231277279/job/22506263406#step:8:745

hendrikmakait:
@milesgranger: Any ideas if it's possible for us to assert the correct number of Spark workers?

milesgranger:
Hrmm, I'm almost inclined to think that this is kinda on Spark for losing and not recovering workers. But otherwise, I'd think one can check with `sc.getConf().get("spark.executor.instances")` (as we looked into this config setting) but I don't know if that's dynamic?

milesgranger:
I can poke around at this tomorrow if that'd be helpful.

hendrikmakait:
> Hrmm, I'm almost inclined to think that this is kinda on Spark for losing and not recovering workers.

I'd generally side with you, but I'm not entirely sure if it's on us (our cluster configuration/setup) that those workers drop without healing.

> I can poke around at this tomorrow if that'd be helpful.

If it's quick to do, that would be nice!

ntabris:
From what I see, when the Spark worker exits, dask nanny/worker is also getting killed, so the VM shuts itself down. Is that what other folks are seeing?

(For example, https://cloud.coiled.io/clusters/410048/account/foobar/information?account=&scopes=%7B%22type%22%3A%22account%22%2C%22id%22%3A7568%2C%22name%22%3A%22foobar%22%2C%22organizationId%22%3A9991%7D&tab=Logs&filterPattern=10.0.21.218&daskLogs=1)

milesgranger:
> From what I see, when the Spark worker exits, dask nanny/worker is also getting killed, so the VM shuts itself down. Is that what other folks are seeing?

[...]

PR #1450, opened 2024-03-12
Track changes in Spark executors between queries

milesgranger:
Part of #1449

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -145,24 +145,6 @@ def cluster_spec(scale):
         idle_timeout="1h",
         wait_for_workers=True,
         scheduler_vm_types=["m6i.xlarge"],
-        backend_options={
-            "ingress": [
-                {
-                    "ports": [
-                        443,
-                        8786,
-                        15003,
-                        22,
-                        7077,
-                        8787,
-                        4040,
-                        8080,
-                        15002,
-                    ],
-                    "cidr": "0.0.0.0/0",
-                },
-            ],
-        },
     )
     if scale == 10:
         return {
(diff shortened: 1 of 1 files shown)


tests/tpch/conftest.py
```diff
@@ -143,6 +145,24 @@ def cluster_spec(scale):
...
+                    "ports": [
+                        443,
+                        8786,
+                        15003,
+                        22,
+                        7077,
+                        8787,
+                        4040,
+                        8080,
+                        15002,
```
fjetter, line 160: It would be very helpful for future work to document what these ports are for (once you're done)

milesgranger: Ya, agreed. I'm kinda hoping platform can adjust ports to add 4040 to default spark ports and this can all go away. Had to specify all of them here since I'm overriding the backend options in order to get 4040 in. 

@ntabris What do you think?

ntabris: I'd want to understand what's on 4040, if we should/can put auth in front of it, etc.

FYI you can use `get_spark_cluster(_open_debug_ports=True)` instead of what you're doing here.

ntabris:
I don't love that to get this endpoint, you're basically turning off all the security.

Maybe that's fine for now, not sure. People port scan known AWS IP addresses, not sure if anyone cares about exposed Spark clusters.

If 4040 is just a simple HTTP thing, might be worth just adding token auth in front of it.

tests/tpch/conftest.py
ntabris: > at the point of the code here the cluster is already made

No idea what you mean, you can only set ports to open when making a new cluster.

milesgranger:
> I don't love that to get this endpoint, you're basically turning off all the security.

I know, help me. ;-) 

It's in draft b/c I'm quite confident I don't want to do it this way. But to get the current number of active executors, it appears using the REST API is the most reliable way. This is draft as I show that it functionally can be done, but it's probably not fully correct in its current standing. :)  

ref: [spark REST API](https://spark.apache.org/docs/latest/monitoring.html#rest-api)

mrocklin:
> I'd want to understand what's on 4040, if we should/can put auth in front of it, etc.

My understanding is that it's a pretty simple HTTP thing that gives information around Spark's runtime state.  Putting a token in front of it seems reasonable to me.  I think that we should treat it much like how we treat Spark's server running on 8080

mrocklin:
Giving it a name, like `/spark-master` might also be good (@milesgranger feel free to suggest a better name if you have one)

milesgranger:
Changes in commit https://github.com/coiled/benchmarks/pull/1450/commits/dc5ac067758604ad6a7839394dcc95d78d6f7db6 depends on https://github.com/coiled/platform/pull/5001 (or something equivalent).

4040 is already exposed at `/spark`, so it appears we just need to store the `spark_dashboard` variable after the `get_spark` call. Of course could repeat the generation logic here, but seems nice to store the link anyway since we store dask dashboard link.

hendrikmakait (approved):
Thanks, @milesgranger! I assume you have tested this manually?

milesgranger:
> I assume you have tested this manually?

Yes, I can attest to this having been done, thanks! :+1:

PR #1451, opened 2024-03-12
Rewrite q17

fjetter:
I was reviewing the implementation for Q17 and was a little surprised. This is actually a pretty heavy query that, in it's current form, requires us to shuffle lineitems multiple times.

However, if I take the sql in the doc string literally, I should only be required to compute this on a filtered linetime (namely, the one after the join, i.e. where `l_partkey = p_partkey`).

analyze-to_frame-0f13b17cb06f0df5a27610b462c448eb

There is a caveat with this... it looks like the optimizer is either never converging or is just taking ages. Even for a scale1 dataset this is simplifying forever

image

### DIFF
--- a/tests/tpch/dask_queries.py
+++ b/tests/tpch/dask_queries.py
@@ -889,23 +889,22 @@ def query_17(dataset_path, fs):
     lineitem = dd.read_parquet(dataset_path + "lineitem", filesystem=fs)
     part = dd.read_parquet(dataset_path + "part", filesystem=fs)
 
+    joined = lineitem.merge(
+        part, left_on="l_partkey", right_on="p_partkey", how="inner"
+    )
+    joined = joined[joined.p_brand == "Brand#23"]
+    joined = joined[joined.p_container == "MED BOX"]
     avg_qnty_by_partkey = (
-        lineitem.groupby("l_partkey")
-        # FIXME: https://github.com/dask-contrib/dask-expr/issues/867
-        .l_quantity.mean(split_out=True)
+        joined.groupby("l_partkey")
+        .l_quantity.mean()
         .to_frame()
         .rename(columns={"l_quantity": "l_quantity_avg"})
     )
+    table = joined.merge(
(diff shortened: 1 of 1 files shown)


fjetter:
regardless of whether this is correct or not, this query breaks the optimizer, see https://github.com/dask/dask-expr/issues/965

fjetter:
so, this rewrite does not actually work because the initial join just blows up without the `p_brand` and `p_container` filters. Pushing those down manually is kind of a violation of the spec but would gain us a quite substantial perf improvement. Interestingly, even the "correctness" tests are not catching this so I assume the data was not prepared in a way that catches this edge case.

Considering that this is the only query where duckdb has an uncanny advantage I can't help but wonder if they are doing this even though they are not allowed to.

hendrikmakait:
TL;DR: I've double-checked and pushing the filters into the subquery is correct.

More context:
Logically, you can think of the correlated subquery as a macro that's executed once per row of the outer query. Since we are only interested in the results of the filtered data, it suffices to execute them once per row that passes the filter predicates. This can be rewritten as a join with the filtered table. (See https://duckdb.org/2023/05/26/correlated-subqueries-in-sql.html for a nice introduction.)

> If that's the case, our dataframe code is actually invalid and we can move forward with a more streamlined implementation.

Technically, it's not invalid but just less efficient.

phofl (approved):
yeah our write up was just not great. I think we can detect this automatically in the future, but this will most likely need a bit of work

fjetter:
I'm not sure if we can detect this automatically. I think this rewrite is only valid assuming `l_partkey` is a primary key, isn't it? This is knowledege we just don't have in dask 🤔

phofl:
Yeah you are correct, sorry, I got confused

hendrikmakait:
> I'm not sure if we can detect this automatically. I think this rewrite is only valid assuming l_partkey is a primary key, isn't it? This is knowledege we just don't have in dask 🤔

This is less about PK-FK relations but about the subquery correlation (which we definitely don't have right now).

PR #1452, opened 2024-03-12
Test TPC-H correctness on more scale factors

### LINKED ISSUE #1443: [TPC-H] Generate expected results for SF > 1
At SF==1, our TPC-H queries boil down to "overcomplicated pandas" since we don't really partition anything. At larger scale factors, several different size-based optimizations kick in and significantly alter the query plans. As a result, results may be correct at SF==1, but may be wrong at SF>1 (see https://github.com/dask-contrib/dask-expr/pull/917#issuecomment-1983112150). 

We should generate the expected results at larger scale factors to be able to run the correctness tests at larger scale.

hendrikmakait:
Closes #1443

~For some reason, Q15 at SF 100 and 100 is off. The DuckDB result appears to be empty, which I assume to be wrong.~ The problem with Q15 has been fixed by #1454.

### DIFF
--- a/tests/tpch/test_correctness.py
+++ b/tests/tpch/test_correctness.py
@@ -14,8 +14,8 @@ pytestmark = pytest.mark.tpch_correctness
 @pytest.fixture(params=[1, 10, 100], scope="session")
 def scale(request):
     scale = request.param
-    # if scale != 100:
-    # pytest.skip(reason="Don't test everything by default")
+    if scale != 100:
+        pytest.skip(reason="Don't test everything by default")
     return scale
 
 
--- a/tests/tpch/test_correctness.py
+++ b/tests/tpch/test_correctness.py
@@ -1,5 +1,4 @@
 import os
-import pathlib
 
 import coiled
 import dask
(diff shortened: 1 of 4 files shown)


tests/tpch/test_correctness.py
```diff
@@ -1,50 +1,63 @@
...
 
-@pytest.fixture(scope="module")
-def answers_path(tmp_path_factory):
-    path = tmp_path_factory.mktemp("answers")
-    return generate_answers(base_path=path)
+@pytest.fixture(params=[1, 10, 100], scope="session")
+def scale(request):
+    scale = request.param
+    # if scale != 100:
+    # pytest.skip(reason="Don't test everything by default")
```
hendrikmakait, line 18: TODO: Don't test every scale by default.

hendrikmakait:
Something is still off in Q15: It fails with `2385908.0563 != 2385908.056300`.

tests/tpch/test_correctness.py
```diff
@@ -1,50 +1,63 @@
...
-@pytest.fixture(scope="module")
-def answers_path(tmp_path_factory):
-    path = tmp_path_factory.mktemp("answers")
-    return generate_answers(base_path=path)
+@pytest.fixture(params=[1, 10, 100], scope="session")
+def scale(request):
+    scale = request.param
+    if scale != 100:
+        pytest.skip(reason="Don't test everything by default")
+    return scale
```
hendrikmakait, line 19: Note that this test doesn't adhere to `--scale X` because we likely want to test correctness at a smaller scale most of the time than the actual benchmarks we run.

milesgranger (approved):

PR #1453, opened 2024-03-12
Fix TPC-H trigger

hendrikmakait:
Fixes https://github.com/coiled/benchmarks/actions/runs/8235051302/job/22518272712#step:12:122


fjetter (approved):
ahhh

PR #1454, opened 2024-03-12
Fix numerical instability in DuckDB Q15

hendrikmakait:
The revenue calculation was numerically unstable and Q15 could (incorrectly) return no results.

### DIFF
--- a/tests/tpch/test_duckdb.py
+++ b/tests/tpch/test_duckdb.py
@@ -615,7 +615,7 @@ def test_query_15(run, connection, dataset_path):
             create or replace temporary view revenue (supplier_no, total_revenue) as
                 select
                     l_suppkey,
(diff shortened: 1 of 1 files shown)


tests/tpch/test_duckdb.py
```diff
@@ -743,7 +743,7 @@ def _():
                 c_name,
                 c_custkey,
                 o_orderkey,
-                o_orderdate as o_orderdat,
```
hendrikmakait, line 746: Driveby

hendrikmakait:
FWIW, I guess that the other libraries have a problem with this as well. @phofl: This kind of instability is something that Dask will become prone to once we disable CSE. If you have a good intuition on how to detect this on query generation, I'd love to flag/handle this.

milesgranger (approved):

PR #1455, opened 2024-03-12
[TPC-H] Require `scale` argument for Dask queries

hendrikmakait:
Having a default for Q11 could lead to the scale used for the data and the query diverging. Requiring it for each query avoids special casing.

### DIFF
--- a/tests/tpch/test_dask.py
+++ b/tests/tpch/test_dask.py
@@ -1,57 +1,57 @@
 import pytest
 
-from . import dask_queries
-
 pytestmark = pytest.mark.tpch_dask
 
 dd = pytest.importorskip("dask.dataframe")
 
-
-def test_query_1(client, dataset_path, fs):
-    dask_queries.query_1(dataset_path, fs).compute()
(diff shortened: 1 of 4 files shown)


tests/tpch/test_dask.py
```diff
@@ -1,57 +1,57 @@
 import pytest
 
-from . import dask_queries
-
 pytestmark = pytest.mark.tpch_dask
 
 dd = pytest.importorskip("dask.dataframe")
 
+from . import dask_queries  # noqa: E402
```
hendrikmakait, line 7: This requires `dask.dataframe` to be importable.

milesgranger (commented):
TPC-H tests seem pretty unhappy. Did you mean to move a scale fixture into this PR?

hendrikmakait:
> TPC-H tests seem pretty unhappy. Did you mean to move a scale fixture into this PR?

Whoops, looks like I forgot to commit a file.

milesgranger (approved):
LGTM, thanks!

PR #1457, opened 2024-03-13
Improve merge ordering for query 18

phofl:
This adopts the polars order and gives us a nice improvement

### DIFF
--- a/tests/tpch/dask_queries.py
+++ b/tests/tpch/dask_queries.py
@@ -958,10 +958,10 @@ def query_18(dataset_path, fs):
     qnt_over_300 = qnt_over_300[qnt_over_300.l_quantity > 300]
 
     table = (
-        qnt_over_300.merge(
-            orders, left_on="l_orderkey", right_on="o_orderkey", how="inner"
+        orders.merge(
+            qnt_over_300, left_on="o_orderkey", right_on="l_orderkey", how="leftsemi"
         )
-        .merge(lineitem, left_on="l_orderkey", right_on="l_orderkey", how="leftsemi")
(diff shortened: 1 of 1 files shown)


tests/tpch/dask_queries.py
```diff
@@ -957,11 +957,13 @@ def query_18(dataset_path, fs):
     )
     qnt_over_300 = qnt_over_300[qnt_over_300.l_quantity > 300]
 
-    L = lineitem.merge(
-        qnt_over_300, left_on="l_orderkey", right_on="l_orderkey", how="leftsemi"
+    table = (
+        qnt_over_300.merge(
+            orders, left_on="l_orderkey", right_on="o_orderkey", how="inner"
+        )
+        .merge(lineitem, left_on="l_orderkey", right_on="l_orderkey", how="leftsemi")
```
hendrikmakait, line 964: nit:
```suggestion
        orders.merge(
            qnt_over_300, left_on="o_orderkey", right_on="l_orderkey", how="leftsemi"
        )
        .merge(lineitem, left_on="o_orderkey", right_on="l_orderkey", how="inner")
```

This sticks more to the SQL specification and shouldn't have a large performance difference.

phofl: Nope that's a lot slower, we should probably investigate why (I suspect because we have to shuffle more data if the left semi is the first join, moving the left semi back is fine), but that's not in scope for this reordering.

phofl: Is the current implementation incorrect? I checked the results and they are the same

hendrikmakait (approved):
Thanks, @phofl!

PR #1464, opened 2024-03-14
Avoid object cast

phofl:


tests/tpch/dask_queries.py
```diff
@@ -66,7 +66,7 @@ def query_2(dataset_path, fs, scale):
     )
     part_filtered = part_filtered[
         (part_filtered["p_size"] == var1)
-        & (part_filtered["p_type"].astype(str).str.endswith(var2))
+        & (part_filtered["p_type"].astype("string[pyarrow]").str.endswith(var2))
```
fjetter, line 69: why is this cast necessary in the first place?

image

```suggestion
        & (part_filtered["p_type"].str.endswith(var2))
```

tests/tpch/dask_queries.py
```diff
@@ -1106,7 +1106,7 @@ def query_20(dataset_path, fs, scale):
     ]
     res_1 = (
         res_1.groupby(["l_partkey", "l_suppkey"])["l_quantity"]
-        .sum()
+        .sum(split_out=True)
```
fjetter, line 1109: Is this something we could do automatically? Why is this necessary?

tests/tpch/dask_queries.py
phofl: Very good point, I was just passing by here. The cast doesn't seem necessary

tests/tpch/dask_queries.py
phofl: We are already shuffling but partitions are still too big after the groupby, which slows us down quite a bit. We might want to reevaluate the factor, but this doesn't really aggregate, so True is the best value here

phofl: Yeah there is always the case where we don't reduce much at all, we could use split_out=True by default but this creates also tiny partitions in other scenarios, I don't think that we can hit every case without cardinality information

tests/tpch/dask_queries.py
```diff
@@ -1108,7 +1107,7 @@ def query_20(dataset_path, fs, scale):
     ]
     res_1 = (
         res_1.groupby(["l_suppkey", "l_partkey"])["l_quantity"]
-        .sum()
+        .sum(split_out=True)
```
phofl, line 1110:
```suggestion
        .sum()
```

tests/tpch/dask_queries.py
phofl: nope this can go away

fjetter (approved):

PR #1465, opened 2024-03-14
Rewrite Q8

hendrikmakait:


phofl:
Any estimate how large the difference is?

hendrikmakait:
Preliminary testing says that there might be none. From what I see, both versions saturate disk I/O though, so maybe that's our bottleneck.

hendrikmakait:
FWIW, we reduce the # of tasks by 1/3. 😅

hendrikmakait:
Closing as stale

PR #1471, opened 2024-03-16
Change join order in query 21

phofl:
this helps together with the PR that uses trivial merges in those cases


tests/tpch/dask_queries.py
```diff
@@ -1107,7 +1107,7 @@ def query_20(dataset_path, fs, scale):
         & (lineitem["l_shipdate"] < shipdate_to)
     ]
     res_1 = (
-        res_1.groupby(["l_partkey", "l_suppkey"])["l_quantity"]
```
fjetter, line 1110: I'm not too happy about manual fixes like this. I believe this is something the optimizer should figure out itself

hendrikmakait: I'd like to understand how this impacts the query runtime. I suspect it's based on cardinality? In that case, we'd have to put this on the list of things we'd like to do once we have cardinality estimates.

fjetter: ah, right. I thought this was about the stuff in https://github.com/dask/dask-expr/pull/946

hendrikmakait: I've reviewed the query, and it looks like you might be right, there's
```python
).merge(
        res_1,
        left_on=["ps_suppkey", "ps_partkey"],
        right_on=["l_suppkey", "l_partkey"],
    )
```
In that case, I agree with you. I'm a bit confused though, this does not appear in the current `main` yet it's not flagged as a change?

hendrikmakait: Never mind, this change is not actually part of Q21 but Q20.

crusaderky:
> the PR that uses trivial merges in those cases

@phofl for my understanding, could you link the PR?

hendrikmakait:
@crusaderky: https://github.com/dask/dask-expr/pull/946

PR #1473, opened 2024-03-18
[TPC-H] Default raise on Spark executor count change

### LINKED ISSUE #1449: [TPC-H] PySpark workloads are vulnerable to undetected worker loss

milesgranger:
Close #1449 

Default raise when Spark executor count changes during fixture use, with option to downgrade to a warning: `--ignore-spark-executor-count`.

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -43,6 +43,12 @@ def pytest_addoption(parser):
         help="Name to use for run",
     )
     parser.addoption("--plot", action="store_true", default=False, help="")
+    parser.addoption(
+        "--ignore-spark-executor-count",
+        action="store_true",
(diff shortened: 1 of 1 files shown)


hendrikmakait (approved):
Thanks, @milesgranger, I love the additional option! <3

PR #1474, opened 2024-03-19
Add option to keep TPCH clusters alive after test

crusaderky:


milesgranger (approved):
I see it's still draft status, but if you don't plan on much/any more changes, you've got my green check!
Thanks, @crusaderky.

PR #1478, opened 2024-03-19
Fix memory issues for Spark

hendrikmakait:
Fixes OOM problems that led to inconsistent TPC-H benchmarks.

### DIFF
--- a/ci/requirements-2tpch-non-dask.in
+++ b/ci/requirements-2tpch-non-dask.in
@@ -10,5 +10,8 @@ grpcio==1.62.1
 grpcio-status==1.62.1
 protobuf==4.25.2
 
+# First version to be able to set memory factors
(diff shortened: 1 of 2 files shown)


ntabris:
Maybe it's coincidence that those are the numbers I used as an example, but just to be clear, I have no idea if those are good memory limit numbers.

hendrikmakait:
> Maybe it's coincidence that those are the numbers I used as an example, but just to be clear, I have no idea if those are good memory limit numbers.

Neither do I, but they make intuitive sense to me. Keep a 10% buffer between full memory utilization of a worker and the machine going OOM, and keep a bit more than 10% of that memory for the 10% of Spark overhead that's set per default.

milesgranger (approved):
Looks good for me, thanks @hendrikmakait!

PR #1482, opened 2024-03-19
Tag all tests using spans

crusaderky:
Spans are based on annotations, and annotations don't work with dask-expr.
However this caveat applies only to _nesting_ annotations in the layers - any annotations that are active when you hit compute/persist will be captured.

image
image

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -206,6 +206,7 @@ def client(
     testrun_uid,
     cluster_kwargs,
     benchmark_time,
+    span,
     restart,
     scale,
     local,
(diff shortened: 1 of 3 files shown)


tests/tpch/conftest.py
```diff
@@ -119,40 +116,6 @@ def name(request, tmp_path_factory, worker_id):
...
-        else:
-            start = time.time()
-            yield
-            end = time.time()
-            test_run_benchmark.duration = end - start
-            test_run_benchmark.start = datetime.datetime.utcfromtimestamp(start)
-            test_run_benchmark.end = datetime.datetime.utcfromtimestamp(end)
-            test_run_benchmark.cluster_name = f"tpch-{module}-{scale}-{name}"
-
-    return _benchmark_time()
```
crusaderky, line 153: Needless copy-paste from `../conftest.py`

PR #1486, opened 2024-03-20
pip-compile tweaks for dependabot

crusaderky:
dependabot seems to ignore requirements.in files that contain the `-r` directive. This PR works around it.

Also bump up snowflake.


milesgranger (approved):
Thanks @crusaderky!

PR #1492, opened 2024-03-21
Fix pyspark local execution

milesgranger:
I've tried this with scale 5 and 10 data w/o trouble after verifying the current state is broken.

My machine has 65G so the logic gives  ~23G to executors, and have tried capping that to 5G per executor and it still worked. 

I somewhat arbitrarily assigned `n_executors = 2`, could very well be a single executor or something else. I'm not sure what might be preferred here by others.

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -8,6 +8,7 @@ import warnings
 import coiled
 import dask
 import filelock
+import psutil
 import pytest
 import requests
 from dask.distributed import LocalCluster, performance_report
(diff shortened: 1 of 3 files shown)


fjetter:
I'm running on a M1 mac book and when I close most of my browser tabs and my IDE I have about 9.4GiB of available memory. When I initially executed this script, I essentially had almost no available memory and the `total_executor_memory_g` math produced negative values for me.
d)

fjetter:
Regardless of the settings. I tried a couple of different configs and this doesn't work for me. Even on scale1 I have a job that appears to be stuck

image


I get that spark may be slow but I don't believe that it can't handle 1GB locally.

milesgranger:
Ya, I'm slightly baffled. I was concentrating on scale 100 to get it running for the comparison. But now am experiencing a similar scenario on scales 1, 5, 10. Queries will fail, while on scale 100 all work fine. :face_exhaling: 

> 2024-04-03 10:12:43,053 - distributed.nanny.memory - WARNING - Worker tcp://127.0.0.1:35463 (pid=148634) exceeded 95% memory budget. Restarting...

This occurs on the lower scales for me, but again, not on scale 100.

PR #1499, opened 2024-03-28
Use `get_spark_cluster` to start spark cluster

fjetter:
There's a dedicated coiled API for launching a spark cluster on Coiled. This is mostly just a wrapper around the normal cluster but it is setting some networking config that allows us to access the dashboard.

There is also no reason to have the spark specific stuff in the general top level conftest

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -3,16 +3,13 @@ import datetime
 import os
 import time
 import uuid
-import warnings
 
 import coiled
 import dask
 import filelock
 import pytest
-import requests
 from dask.distributed import LocalCluster, performance_report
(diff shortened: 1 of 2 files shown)


ntabris:
Unless you're using `_open_debug_ports` (which I'd recommend against), there's no difference from using a regular cluster (which I'd recommend).

fjetter:
> from using a regular cluster

I was having difficulties with connecting to the spark dashboard. I thought this was opening additional ports

ntabris:
> I was having difficulties with connecting to the spark dashboard. I thought this was opening additional ports

Not unless you pass `_open_debug_ports=True`.

If you're having trouble connecting to spark dashboard, it's probably some other issue... happy to look at this if you have more info or want to look together.

PR #1504, opened 2024-04-05
Update README for misaligned git tags

crusaderky:
I just wasted another overnight A/B run because of this.

This PR saves 30s when syncing the tags - which is to say, very very frequently. 
Hopefully if the operation is mechanical enough to do it will become automatic for developers.

It would be better to have a pre-commit action but that would take a lot more effort to write.


milesgranger (approved):
Nice, thanks @crusaderky! :pray:

PR #1490, opened 2024-03-21
couple of fixes for local tpch runs

fjetter:
A couple of fixes for getting this running locally again

I'm still struggling with pyspark 

https://github.com/coiled/dask-engineering/issues/248#issuecomment-2012029012

Closes https://github.com/coiled/benchmarks/issues/1496

PR #1505, opened 2024-04-08
Fix local pyspark for osx

fjetter:
This is a follow up from https://github.com/coiled/benchmarks/pull/1490 and supersedes https://github.com/coiled/benchmarks/pull/1492

It also changes the blocking call for spark from `show` to `collect`. Nothing changes really but we won't print the result. I think this is the closest there is to an ordinary `compute`

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -205,12 +205,17 @@ def spark_setup(cluster, local):
 
     spark_dashboard: Url
     if local:
+        cluster.close()
         from pyspark.sql import SparkSession
 
         # Set app name to match that used in Coiled Spark
         spark = (
             SparkSession.builder.appName("SparkConnectServer")
+            .config("spark.driver.host", "127.0.0.1")
             .config("spark.driver.bindAddress", "127.0.0.1")
+            # By default, the driver is only allowed to use up to 1g of memory
+            # which causes it to crash when collecting results
+            .config("spark.driver.memory", "10g")
             .getOrCreate()
         )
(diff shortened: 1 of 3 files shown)


tests/tpch/conftest.py
```diff
@@ -205,12 +205,17 @@ def spark_setup(cluster, local):
...
     spark_dashboard: Url
     if local:
+        cluster.close()
         from pyspark.sql import SparkSession
 
         # Set app name to match that used in Coiled Spark
         spark = (
             SparkSession.builder.appName("SparkConnectServer")
+            .config("spark.driver.host", "127.0.0.1")
             .config("spark.driver.bindAddress", "127.0.0.1")
```
fjetter, line 215: Of course both configs have to be set. Because why not. This appears to be the reason for some jobs to timeout during collect (see https://github.com/coiled/benchmarks/pull/1492#issuecomment-2014849506)

tests/tpch/conftest.py
```diff
@@ -223,19 +228,21 @@ def spark_setup(cluster, local):
...
-    yield spark
+        df = spark.createDataFrame(
+            [
+                Row(a=1, b=2.0, c="string1"),
+            ]
+        )
+        df.show()
+        yield spark
+    finally:
+        spark.stop()
```
fjetter, line 245: just good hygiene. I'm not sure any longer if it is necessary but it feels like it shouldn't hurt

phofl (approved):
lgtm

PR #1510, opened 2024-04-24
[TPC-H] Configure temporary directory in `duckdb` for out-of-core processing

### LINKED ISSUE #1509: Is duckdb out-of-core processing properly enabled?
I just saw the great talk on Dask DataFrames 2.0 at PyData Berlin! I was a bit surprised that duckdb timed out for some of the queries. According to https://duckdb.org/docs/guides/performance/how_to_tune_workloads#larger-than-memory-workloads-out-of-core-processing, if you are not connected to a persistent duckdb database file, which I think is not the case based on the code in https://github.com/coiled/benchmarks/blob/63ca3c20cfd6c8352eebf880211e41a85793be32/tests/tpch/test_duckdb.py, you'd need to set a temporary directory so that duckdb can spill over to disk.

I'm not 100% if this is not set already somewhere else as I didn't dig through all the testing related code but thought you might want to know.

Related issues are #1488, #1214, and #1194.

hendrikmakait:
Closes #1509

### DIFF
--- a/tests/tpch/test_duckdb.py
+++ b/tests/tpch/test_duckdb.py
@@ -1,3 +1,8 @@
+import uuid
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
+
 import botocore.session
 import pytest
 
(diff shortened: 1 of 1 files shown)


ntabris:
This would be a temp directory on the VM? If so, I'd recommend doing something under `/scratch` so it uses NVMe if present.

hendrikmakait:
> This would be a temp directory on the VM? If so, I'd recommend doing something under `/scratch` so it uses NVMe if present.

Good point, I'll adjust the code to check if we're running on the cloud. I suppose we should create the temporary directory within the workspace of the Dask worker that runs the `coiled.function`?

ntabris:
> I suppose we should create the temporary directory within the workspace of the Dask worker that runs the coiled.function?

Yeah, that would work—dask worker workspace is already set to be in `/scratch`.

> Is there a way I can figure out when NVMe is present? Is it only specific instance types?

It's specific instance types on AWS, in particular, instance types with `d` in the family, such as `m6id` or `g4dn` (plus a few others, such as `g5` which does have NVMe).

hendrikmakait:
Closing as stale

PR #1511, opened 2024-04-25
Enable SF 10,000 in Github Actions

hendrikmakait:
Extracted from #1506

### DIFF
--- a/.github/workflows/tpch.yml
+++ b/.github/workflows/tpch.yml
@@ -5,9 +5,10 @@ on:
       scale:
         description: 'Scale Factor'
         required: true
-        default: 1000
+        default: 10000
         type: choice
         options:
+          - 10000
           - 1000
           - 100
           - 10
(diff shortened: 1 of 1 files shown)


.github/workflows/tpch.yml
```diff
@@ -33,11 +34,6 @@ on:
         default: true
         type: boolean
 
-concurrency:
-  # Only allow a single run at a time to rate limiting
-  group: tpch
-  cancel-in-progress: true
```
fjetter, line 39: I think we should have some kind of control but maybe a little more granular. I'm fine either way about the job as long as we're a little careful

fjetter (approved):

.github/workflows/tpch.yml
hendrikmakait: I figured you had some reason to remove it for now so I stuck with that, I'm happy to iterate on this.

PR #1516, opened 2024-05-21
Revert pip compile workflow

fjetter:
I suggested to use a toolchain that supported depandabot, see https://github.com/coiled/benchmarks/issues/1437 
which triggered a migration to pip-compile in https://github.com/coiled/benchmarks/pull/1447

I personally had a hard time working with this new structure and pip compile and believe we should revert back to a pure conda workflow.

I think a different way of achieving a similar result (maybe w/out dependabot idk) would be to use [conda lock](https://github.com/conda/conda-lock) or even [pixi](https://github.com/prefix-dev/pixi)


hendrikmakait (approved):
LGTM

PR #1519, opened 2024-08-01
Add a benchmark for take

phofl:


tests/benchmarks/test_array.py
```diff
@@ -257,3 +258,29 @@ def test_map_overlap_sample(small_client, new_array):
...
+    indexer = list(flatten(_create_indexer(n, chunk_n)))
+    x[:, indexer, :].sum().compute()
+
+
+# def test_shuffle(small_client, new_array):
+#     n = 2000
+#     chunk_n = 250
+#     x = new_array((n, n, n), chunks=(chunk_n, chunk_n, chunk_n))
+#     indexer = _create_indexer(n, chunk_n)
+#     x.shuffle(indexer, axis=1).sum().compute()
```
hendrikmakait, line 286:
```suggestion
```

tests/benchmarks/test_array.py
```diff
@@ -257,3 +258,29 @@ def test_map_overlap_sample(small_client, new_array):
...
+    indexer = []
+    for i in range(0, n, chunk_n):
+        indexer.append(idx[i : i + chunk_n].tolist())
+    return indexer
+
+
+def test_take(small_client, new_array):
+    n = 2000
+    chunk_n = 250
+    x = new_array((n, n, n), chunks=(chunk_n, chunk_n, chunk_n))
```
hendrikmakait, line 276: nit: Annotating some data sizes would be interesting.

tests/benchmarks/test_array.py
```diff
@@ -257,3 +258,29 @@ def test_map_overlap_sample(small_client, new_array):
...
+    idx = np.arange(0, n)
+    np.random.shuffle(idx[: n // 10])
+
+    indexer = []
+    for i in range(0, n, chunk_n):
+        indexer.append(idx[i : i + chunk_n].tolist())
+    return indexer
+
+
+def test_take(small_client, new_array):
```
hendrikmakait, line 273: Is this test based on some real-world use case? If so, some context would be helpful.

hendrikmakait (commented):
Waiting for CI before approving but leaving some comments already.

tests/benchmarks/test_array.py
phofl: should have converted to draft, I don't want to merge this before shuffle is available

tests/benchmarks/test_array.py
phofl: this is the smallest thing that the current implementation can still handle in reasonable time... everything bigger blows up

PR #1521, opened 2024-08-02
Add stability test for private pip install

### LINKED ISSUE #1245: Add tests for installing private package with `PipInstall` plugin
We should test that the `PipInstall` plugin works with setting the token via `client.send_private_envs`.

hendrikmakait:
Closes #1245


phofl (approved):

phofl:
thx

PR #1523, opened 2024-08-02
Truncate `test_run` table in `benchmark.db` to the last 90 days

hendrikmakait:
This PR truncates the `test_run` table in the `benchmark.db` to the runs of the last 90 days. This reduces the size of the database from ~10 GB to ~500 MB and keeps it at that level. Currently, CI workers are unable to process the existing DB causing CI to fail since June 1st.

We don't truncate `tpch_run` since we don't store `start`.


phofl:
thx

PR #1528, opened 2024-08-13
Add example for xarray groupby reduction causing memory pressure

phofl:
each query runs around 50s, we are fine with spot-with-fallback since we mostly care about memory pressure

### DIFF
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -37,7 +37,6 @@ spill_cluster:
 group_reduction_cluster:
   n_workers: 20
   worker_vm_types: [m6i.xlarge]  # 4CPU, 16GiB
-  spot_policy: spot_with_fallback
   region: "us-east-1"  # Same region as dataset
 
 # For tests/workflows/test_embarrassingly_parallel.py
--- a/cluster_kwargs.yaml
+++ b/cluster_kwargs.yaml
@@ -33,6 +33,13 @@ spill_cluster:
   worker_disk_size: 64
   worker_vm_types: [m6i.large]  # 2CPU, 8GiB
 
+# For tests/benchmarks/test_xarray.py
+group_reduction_cluster:
+  n_workers: 20
+  worker_vm_types: [m6i.xlarge]  # 4CPU, 16GiB
+  spot_policy: spot_with_fallback
+  region: "us-east-1"  # Same region as dataset
+
 # For tests/workflows/test_embarrassingly_parallel.py
 embarrassingly_parallel:
   n_workers: 100
--- a/tests/benchmarks/test_array.py
+++ b/tests/benchmarks/test_array.py
@@ -37,10 +37,10 @@ def test_anom_mean(small_client, new_array):
         dims=["time", "x"],
         coords={"day": ("time", np.arange(data.shape[0]) % ngroups)},
     )
-
-    clim = arr.groupby("day").mean(dim="time")
-    anom = arr.groupby("day") - clim
-    anom_mean = anom.mean(dim="time")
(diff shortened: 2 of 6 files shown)


cluster_kwargs.yaml
```diff
@@ -33,6 +33,13 @@ spill_cluster:
   worker_disk_size: 64
   worker_vm_types: [m6i.large]  # 2CPU, 8GiB
 
+# For tests/benchmarks/test_xarray.py
+group_reduction_cluster:
+  n_workers: 20
+  worker_vm_types: [m6i.xlarge]  # 4CPU, 16GiB
+  spot_policy: spot_with_fallback
```
fjetter, line 40: even though we care about memory pressure, this causes the runtime to flake around unnecessarily and our regression detection would trigger unnecessarily.

likewise, if spot instances fail, data replication might also push memory pressure up

```suggestion
```

fjetter: we have tests that run much longer than 50s so I don't think we have to be cost sensitive for this specific addition

tests/benchmarks/test_array.py
```diff
@@ -37,10 +37,10 @@ def test_anom_mean(small_client, new_array):
         dims=["time", "x"],
         coords={"day": ("time", np.arange(data.shape[0]) % ngroups)},
     )
-
-    clim = arr.groupby("day").mean(dim="time")
-    anom = arr.groupby("day") - clim
-    anom_mean = anom.mean(dim="time")
+    with xarray.set_options(use_flox=False):
```
hendrikmakait, line 40: I'd suggest moving the xarray-related tests to `test_xarray.py`.

tests/benchmarks/test_xarray.py
```diff
@@ -0,0 +1,72 @@
...
+        pytest.param(
+            lambda x: x.chunk(time=TimeResampler("ME"))
+            .groupby("time.month")
+            .mean(method="cohorts"),
+            id="chunked-cohorts",
+        ),
+    ],
+)
+def test_xarray_groupby_reduction(group_reduction_client, func):
+    pytest.importorskip("xarray")
```
hendrikmakait, line 57: Let's move these to the top of the module. There's no need to spin up clusters if we don't have xarray installed.

tests/benchmarks/test_array.py
phofl: I'd rather do this as a follow up, but the general idea is good

tests/benchmarks/test_xarray.py
phofl: done

tests/benchmarks/test_xarray.py
```diff
@@ -0,0 +1,70 @@
...
+import fsspec
+import pytest
+from coiled import Cluster
+from distributed import Client
+from xarray.groupers import TimeResampler
+
+from tests.conftest import dump_cluster_kwargs
+from tests.utils_test import wait
+
+xr = pytest.importorskip("xarray")
```
hendrikmakait, line 12: From what it looks like, we can skip the `importorskip` statements, otherwise this shouldn't have worked in the first place. 

IIUC, both `xarray` and `flox` are installed via `ci/environment.yml` which is the basis for all CI jobs and the development environment. If I'm missing something, the suggestion below should fix the order of skips and imports.

```suggestion
from tests.conftest import dump_cluster_kwargs
from tests.utils_test import wait

xr = pytest.importorskip("xarray")
from xarray.groupers import TimeResampler
```

phofl: good point.

the skips are for AB tests if you run on smaller environments

PR #1530, opened 2024-08-15
Adjust TPC-H testing on CI

hendrikmakait:
I don't see a point in testing 10 GB workloads. If we're cost-conscious, I'd rather reduce the frequency of testing.

PR #1531, opened 2024-08-15
Use 50 GB Parquet+PyArrow dataset for H2O tests on CI

hendrikmakait:
Similar to #1530, the 5 GB dataset feels too small to benchmark behavior we care about.

### DIFF
--- a/tests/benchmarks/test_h2o.py
+++ b/tests/benchmarks/test_h2o.py
@@ -31,7 +31,7 @@ if enabled_datasets is not None:
         raise ValueError("Unknown h2o dataset(s): ", unknown_datasets)
 else:
     enabled_datasets = {
-        "5 GB (parquet)",
+        "50 GB (parquet+pyarrow)",
     }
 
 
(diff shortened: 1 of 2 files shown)


fjetter:
how long is one test run (compared to the rest of the suite)?

hendrikmakait:
>  how long is one test run (compared to the rest of the suite)?

I'll have to convert this back to draft, it looks like some of the workloads run OOM.

tests/benchmarks/test_h2o.py
```diff
@@ -89,77 +87,69 @@ def test_q2(ddf):
...
     (
-        ddf[["id2", "id4", "v1", "v2"]]
-        .groupby(["id2", "id4"], dropna=False, observed=True)
+        ddf.groupby(["id2", "id4"], dropna=False, observed=True)
         .apply(
             lambda x: pd.Series({"r2": x.corr(numeric_only=True)["v1"]["v2"] ** 2}),
             meta={"r2": "float64"},
         )
-        .compute()
+        .persist()
```
phofl, line 154: Do we have a wait somewhere?

hendrikmakait: My bad, I've just been adding this.

phofl (approved):

hendrikmakait:
Some of these workloads cause workers to OOM, so I'm shelving this for now and will look into it at a later point.

PR #1534, opened 2024-08-20
Disable automated tests on PRs and run them only once per week

hendrikmakait:


phofl:
i.e. we have to make sure that a test passes locally because it won't be run on CI when we create a PR, correct?

hendrikmakait:
> i.e. we have to make sure that a test passes locally because it won't be run on CI when we create a PR, correct?

We're still able to fire off manual runs via the "Run workflow" button on https://github.com/coiled/benchmarks/actions/workflows/tests.yml

A better way of handling partial runs would be nice, but is out-of-scope.

phofl (approved):

phofl:
thx

PR #1540, opened 2024-09-05
Tag coiled.function in TPC-H

hendrikmakait:
This makes sure that `coiled.function` are tagged appropriately.


jrbourbeau (approved):
LGTM if you've tested this locally works

PR #1546, opened 2024-09-11
Add multi-variable rechunking

jrbourbeau:
This extends our rechunking example to cover multiple dataarrays. This doesn't complete successfully today. 

cc @hendrikmakait  @phofl


hendrikmakait (approved):
Thanks, @jrbourbeau!

PR #1547, opened 2024-09-12
Add NWM zonal averaging workflow

jrbourbeau:


.github/workflows/tests.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python_version: ["3.10"]
-        pytest_args: [tests --ignore=tests/tpch]
+        pytest_args: [tests/geospatial/test_zonal_average.py --ignore=tests/tpch]
```
jrbourbeau, line 34: TODO: Revert before merging

tests/geospatial/test_rechunking.py
```diff
@@ -1,25 +1,39 @@
...
+    client_factory,
+    cluster_kwargs={
+        "workspace": "dask-engineering-gcp",
+        "region": "us-central1",
+        "wait_for_workers": True,
+    },
+    scale_n_workers={
+        "small": 10,
+        "large": 100,
+    },
```
jrbourbeau, line 17: This change came out of needing to now manage having different configs for "small"/"large" versions of tests that run on multiple cloud providers. I thought inlining the spec for what "small"/"large" mean and other config (mostly related to the dataset being used) like "region"/"workspace" directly in the test definition felt nice, and was pretty easy to get working, but am open to other options.

tests/geospatial/test_zonal_average.py
```diff
@@ -0,0 +1,66 @@
...
+            subset,
+            counties_aligned.rename("county"),
+            func="mean",
+            expected_groups=(county_id,),
+        )
+
+        county_mean.load()
+        yearly_mean = county_mean.mean("time")
+        # Save dataset for further analysis
+        yearly_mean.to_netcdf(s3_url)
```
jrbourbeau, line 66: This line is currently failing with 

```
self = <scipy.io._netcdf.netcdf_file object at 0x13c58ac90>
filename = 's3://coiled-runtime-ci/test-scratch/test_nwm-9d96d682af1a4ba28a0b596c49dea501', mode = 'w', mmap = None, version = 2
maskandscale = False

    def __init__(self, filename, mode='r', mmap=None, version=1,
                 maskandscale=False):
        """Initialize netcdf_file from fileobj (str or file-like)."""
        if mode not in 'rwa':
            raise ValueError("Mode must be either 'r', 'w' or 'a'.")

        if hasattr(filename, 'seek'):  # file-like
            self.fp = filename
            self.filename = 'None'
            if mmap is None:
                mmap = False
            elif mmap and not hasattr(filename, 'fileno'):
                raise ValueError('Cannot use file object for mmap')
        else:  # maybe it's a string
            self.filename = filename
            omode = 'r+' if mode == 'a' else mode
>           self.fp = open(self.filename, '%sb' % omode)
E           FileNotFoundError: [Errno 2] No such file or directory: 's3://coiled-runtime-ci/test-scratch/test_nwm-9d96d682af1a4ba28a0b596c49dea501'

../../../mambaforge/envs/benchmarks/lib/python3.12/site-packages/scipy/io/_netcdf.py:247: FileNotFoundError
```

tests/geospatial/test_rechunking.py
jrbourbeau: Noting that @hendrikmakait and I chatted about this offline earlier today. We both agreed that this is a fine start and we can always refactor as we start adding more workflows and have a better picture of what needs we'll have. So go with this approach for now and change things later as needed.

tests/geospatial/test_zonal_average.py
jrbourbeau: Yes, but this isn't actually needed after all. It's writing out a small, in-memory Xarray dataset to a netcdf file. I'm fine calling that out of scope for this benchmark.

jrbourbeau:
Here's an example of the "small" version running https://cloud.coiled.io/clusters/590277/account/dask-engineering/information. It takes ~18 minutes and costs ~$0.60. Running the "large" version now.

tests/geospatial/test_zonal_average.py
```diff
@@ -0,0 +1,58 @@
...
+        if scale == "small":
+            # 6.03 TiB
+            time_range = slice("2020-01-01", "2020-12-31")
+        else:
+            # 252.30 TiB
+            time_range = slice("1979-02-01", "2020-12-31")
+        subset = ds.zwattablrt.sel(time=time_range)
+
+        counties = rioxarray.open_rasterio(
+            s3.open("s3://nwm-250m-us-counties/Counties_on_250m_grid.tif"),
```
dcherian, line 41: You could parameterize on two files here:
```
            s3.open("s3://nwm-250m-us-counties/HUC12s_on_1000m_grid.tif"),
```
this is 100_000 watersheds instead of 3000 counties

jrbourbeau: Do you have the path for the other file?

jrbourbeau: Ah, I take it it's `s3://nwm-250m-us-counties/HUC12s_on_1000m_grid.tif`?

dcherian: ah actually this isn't on the right grid yet. I'll get back to you on this.

dcherian: OK here you go: `"s3://nwm-250m-us-counties/HUC12s_on_250m_grid.tif"` It *should* slot right in.

jrbourbeau: Awesome, thanks @dcherian. I'm going to circle back to this as I'd like to merge this PR in so I can make use of some of the fixtures here in another PR

Issue #1548, opened 2024-09-16
[Tracking] Implement Geospatial Benchmarks


hendrikmakait:
# Benchmarks

* [x] Rechunking (#1546) 
* [ ] Downscaling
* [ ] Satellite-data processing 
* [ ] Zonal averaging (https://github.com/coiled/benchmarks/pull/1547)
* [ ] Regridding
* [ ] Climatology

Discussion issue: https://github.com/coiled/benchmarks/discussions/1545

PR #1550, opened 2024-09-16
Add satellite image processing benchmark

jrbourbeau:
xref https://github.com/coiled/benchmarks/issues/1548

### DIFF
--- a/tests/geospatial/test_satellite_filtering.py
+++ b/tests/geospatial/test_satellite_filtering.py
@@ -1,4 +1,5 @@
 import datetime
+import os
 
 import fsspec
 import geojson
@@ -44,7 +45,7 @@ def harmonize_to_old(data):
 
     old = data.sel(time=slice(cutoff))
 
-    to_process = list(set(bands) & set(data.band.data.tolist()))
+    to_process = list(set(bands) & set(data.data_vars.tolist()))
     new = data.sel(time=slice(cutoff, None)).drop_sel(band=to_process)
 
     new_harmonized = data.sel(time=slice(cutoff, None), band=to_process).clip(offset)
@@ -55,6 +56,7 @@ def harmonize_to_old(data):
 
 
 def test_satellite_filtering(
+    az_url,
     scale,
     client_factory,
     cluster_kwargs={
@@ -68,7 +70,12 @@ def test_satellite_filtering(
     },
 ):
     with client_factory(
-        **scale_kwargs[scale], **cluster_kwargs
+        **scale_kwargs[scale],
+        env={
+            "AZURE_STORAGE_ACCOUNT_NAME": os.environ["AZURE_STORAGE_ACCOUNT_NAME"],
+            "AZURE_STORAGE_SAS_TOKEN": os.environ["AZURE_STORAGE_SAS_TOKEN"],
+        },
+        **cluster_kwargs,
     ) as client:  # noqa: F841
         catalog = pystac_client.Client.open(
             "https://planetarycomputer.microsoft.com/api/stac/v1",
@@ -113,11 +120,12 @@ def test_satellite_filtering(
             groupby="solar_day",
         )
         # See https://planetarycomputer.microsoft.com/dataset/sentinel-2-l2a#Baseline-Change
-        ds = harmonize_to_old(ds)
+        # FIXME
+        # ds = harmonize_to_old(ds)
 
         # Compute humidity index
         humidity = (ds.B08 - ds.B11) / (ds.B08 + ds.B11)
 
         result = humidity.groupby("time.month").mean()
-        result = result.persist()
+        result.to_zarr(az_url)
         wait(result)
--- a/tests/geospatial/test_satellite_filtering.py
+++ b/tests/geospatial/test_satellite_filtering.py
@@ -1,0 +1,117 @@
+import datetime
+
+import fsspec
+import geojson
+import planetary_computer
+import pystac_client
+import stackstac
+import xarray as xr
+from distributed import wait
+
+
+def harmonize_to_old(data):
+    """
+    Harmonize new Sentinel-2 data to the old baseline.
+
+    Parameters
+    ----------
+    data: xarray.DataArray
+        A DataArray with four dimensions: time, band, y, x
+
+    Returns
+    -------
+    harmonized: xarray.DataArray
+        A DataArray with all values harmonized to the old
+        processing baseline.
+    """
+    cutoff = datetime.datetime(2022, 1, 25)
+    offset = 1000
+    bands = [
+        "B01",
+        "B02",
+        "B03",
+        "B04",
+        "B05",
+        "B06",
+        "B07",
+        "B08",
+        "B8A",
+        "B09",
+        "B10",
+        "B11",
+        "B12",
+    ]
+
+    old = data.sel(time=slice(cutoff))
+
+    to_process = list(set(bands) & set(data.band.data.tolist()))
+    new = data.sel(time=slice(cutoff, None)).drop_sel(band=to_process)
+
+    new_harmonized = data.sel(time=slice(cutoff, None), band=to_process).clip(offset)
+    new_harmonized -= offset
+
+    new = xr.concat([new, new_harmonized], "band").sel(band=data.band.data.tolist())
+    return xr.concat([old, new], dim="time")
+
+
+def test_satellite_filtering(
+    scale,
+    client_factory,
+    cluster_kwargs={
+        "workspace": "dask-engineering-azure",
+        "region": "westeurope",
+        "wait_for_workers": True,
+    },
+    scale_kwargs={
+        "small": {"n_workers": 10},
+        "large": {"n_workers": 100},
+    },
+):
+    with client_factory(
+        **scale_kwargs[scale], **cluster_kwargs
+    ) as client:  # noqa: F841
+        catalog = pystac_client.Client.open(
+            "https://planetarycomputer.microsoft.com/api/stac/v1",
+            modifier=planetary_computer.sign_inplace,
+        )
+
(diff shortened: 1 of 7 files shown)


jrbourbeau (commented):
Still TODO: 

- [x] Try with `odc.stac` instead of `stackstac`
- [ ] Write output to Zarr instead of persisting to cluster memory


@guillaumeeb @Kirill888 does this look like it captures a common use case for folks processing satellite images?

tests/geospatial/test_satellite_filtering.py
```diff
@@ -0,0 +1,117 @@
...
+import fsspec
+import geojson
+import planetary_computer
+import pystac_client
+import stackstac
+import xarray as xr
+from distributed import wait
+
+
+def harmonize_to_old(data):
```
guillaumeeb, line 12: Interesting, I knew that there was some version change, but didn't know that this could affect the processing. This adds some real problematics, but also complexifies the benchmark, this might no be needed...

tests/geospatial/test_satellite_filtering.py
```diff
@@ -0,0 +1,117 @@
...
+
+        search = catalog.search(
+            collections=["sentinel-2-l2a"],
+            intersects=area_of_interest,
+            datetime=time_of_interest,
+        )
+        items = search.item_collection()
+
+        # Construct Xarray Dataset from stack items
+        ds = stackstac.stack(items, assets=["B08", "B11"], chunksize=(400, 1, 256, 256))
```
guillaumeeb, line 107: So here you can complexify and select other bands for other indices like NDVI, NDSI, NDWI.

tests/geospatial/test_satellite_filtering.py
```diff
@@ -0,0 +1,117 @@
...
+):
+    with client_factory(
+        **scale_kwargs[scale], **cluster_kwargs
+    ) as client:  # noqa: F841
+        catalog = pystac_client.Client.open(
+            "https://planetarycomputer.microsoft.com/api/stac/v1",
+            modifier=planetary_computer.sign_inplace,
+        )
+
+        # GeoJSON for region of interest is from https://github.com/isellsoap/deutschlandGeoJSON/tree/main/1_deutschland
```
guillaumeeb, line 78: How much Sentinel 2 tiles does that represent? Do you have an idea?

tests/geospatial/test_satellite_filtering.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        items = search.item_collection()
+
+        # Construct Xarray Dataset from stack items
+        ds = stackstac.stack(items, assets=["B08", "B11"], chunksize=(400, 1, 256, 256))
+        # See https://planetarycomputer.microsoft.com/dataset/sentinel-2-l2a#Baseline-Change
+        ds = harmonize_to_old(ds)
+
+        # Compute humidity index
+        humidity = (ds.sel(band="B08") - ds.sel(band="B11")) / (
```
guillaumeeb, line 112: It would be prettier to have `(ds.B08 - ds.B11) / (ds.B08 + ds.B11)`, but I guess this is the way stackstack is working, and not a big deal.

tests/geospatial/test_satellite_filtering.py
```diff
@@ -0,0 +1,117 @@
...
+            ds.sel(band="B08") + ds.sel(band="B11")
+        )
+        result = humidity.groupby("time.month").mean()
+        result = result.persist()
```
guillaumeeb, line 116: I would also love to see the plot and if it represent something, typically seasonality at least.

guillaumeeb (commented):
Thanks @jrbourbeau, this is clearly what I had in mind, and I think it's a very good basic benchmark on imagery, and underlying infrastructure (here Microsoft Planetary Computer). Just made some comments, but more questions than remarks!

Were you able to run this? Do you have an idea of the data volume you read? Did you have problem with Datacube being to sparse?

cc @TomAugspurger

tests/geospatial/test_satellite_filtering.py
jrbourbeau: `search.item_collection()` returns 2835 items. Is that what you're asking about, or something else?

tests/geospatial/test_satellite_filtering.py
jrbourbeau: Yeah, you're right this has to do with what `stackstac` returns (data array). `odc.stac` returns a dataset with the structure you're proposing. Thoughts on `stackstac` vs `odc.stac`?

tests/geospatial/test_satellite_filtering.py
jrbourbeau: +1

tests/geospatial/test_satellite_filtering.py
jrbourbeau: Do you have a suggestion on which indicies are most common? I see NDVI come up a lot. Is there any computation difference between these indicies, or are there all just relatively straightforward reductions like we already have here?

tests/geospatial/test_satellite_filtering.py
guillaumeeb: That was more a question about the number of Sentinel 2 zone in UTM format, but nevermind, it's already a good answer. If my calculation is correct, this means already half a TB of Data to load.

tests/geospatial/test_satellite_filtering.py
guillaumeeb: > Thoughts on stackstac vs odc.stac?

Nope, there is a thread about that on Pangeo Discourse I think. Both have probably advantages and drawbacks...

tests/geospatial/test_satellite_filtering.py
guillaumeeb: Yeah, I see NDVI as the HelloWorld of Satellite image processing. There is no computation difference on all if these indices, it's always the same formula, you only switch the bands. You probably already know, but NDVI vor Vegetation, NDSI for Snow, NDWI for Water.

TomAugspurger:
Are these benchmarks running in a stable cloud / region? It'd be nice to find a dataset in the same region if possible (to cut down on egress costs and speed up the I/O portion of the benchmark, which probably isn't relevant for dask).

On stackstac vs. odc-stac, the main things to be aware of are

1. they build task graphs differently. Beyond just DataArray vs. Dataset, I think that odc-stac loading includes a `groupby` stage to ensure that all of the pixels from the same time end up in the same pixel plane (where "same time" is configurable, so that a scene captured a few seconds later can be considered the same if you want).
2. odc-stac will automatically use overviews if you're requesting lower-resolution data (but not relevant here, since you don't pass `resolution=`)

I'll give this workload a shot today or tomorrow and will report back.

tests/geospatial/test_satellite_filtering.py
```diff
@@ -0,0 +1,117 @@
...
+            "coordinates": coordinates,
+        }
+
+        # Get stack items
+        if scale == "small":
+            time_of_interest = "2024-01-01/2024-09-01"
+        else:
+            time_of_interest = "2015-01-01/2024-09-01"
+
+        search = catalog.search(
```
TomAugspurger, line 99: Does Coiled have a spot where we could store a small parquet file? As written, this is hitting the search endpoint every time the benchmark runs, which isn't exercising Dask at all and has the potential to throw errors.

We could instead do the search once and use [stac-geoparquet](https://github.com/stac-utils/stac-geoparquet) to cache that result.

jrbourbeau:
> Are these benchmarks running in a stable cloud / region?

Right now this is running in `westeurope` on Azure, which should be where the underlying data is stored, but we can run in any region on AWS, GCP, or Azure. 

> I'll give this workload a shot today or tomorrow and will report back.

That'd be great. I'm happy to chat generally about this. Also, let me know if you need access to a Coiled workspace that's configured Azure.

tests/geospatial/test_satellite_filtering.py
jrbourbeau: Yeah, I may need to do a little setup to give the CI runner access to an Azure bucket, but we can definitely do that

phofl: I assume that the parquet file is small enough that we can also put it into s3 if that’s easier?

jrbourbeau: Ah, yeah, fair point

jrbourbeau:
Okay, so here's notebook (https://gist.github.com/jrbourbeau/900b602d19fe8087cafc0490b5c26f68) that runs the same computation using `odc.stac`. Here's the specific `odc.stac.load` call

```python
resolution = 10
SHRINK = 4
resolution = resolution * SHRINK

ds = odc.stac.load(
    items,
    chunks={},
    patch_url=planetary_computer.sign,
    resolution=resolution,
    crs="EPSG:3857",
    groupby="solar_day",
)
``` 

where I use things like `groupby="solar_day"`, which I saw used in a couple of examples I found. This seems to produce a much smaller graph and is more performant in general.

hendrikmakait:
I've updated the benchmark to use `odc-stac`.

Open TODOs:

* [ ] Write result to blob storage
* [ ] Cache search result as `stac-geoparquet` file

tests/geospatial/test_satellite_filtering.py
hendrikmakait: > So here you can complexify and select other bands for other indices like NDVI, NDSI, NDWI.

Is it common to calculate multiple of these at the same time?

hendrikmakait:
This PR is ready for review. I haven't bothered too much with caching data from Planetary Computer yet since I saw some occasional permissions issues when data seems to have gotten stale, but we can tackle that when necessary.

phofl (approved):

PR #1551, opened 2024-09-18
Add benchmark for NetCDF --> Zarr cloud-optimization

jrbourbeau:
NetCDF datasets being slow/not scaling well has come up a lot. This PR adds a new benchmark that loads the `nex-gddp-cmip6` dataset (https://registry.opendata.aws/nex-gddp-cmip6/) from AWS, which is stored as a bunch of `.nc` files, and converts that dataset to Zarr, a more modern, cloud-optimized format.

This is using `xr.open_mfdataset(..., parallel=True)` which is both common and _really_ slow when opening lots of NetCDF files, which I like because I've seen this with many users in practice. 

One thing I'm not sure about is how representative this benchmark is as is. I don't know if folks do this NetCDF --> Zarr conversion in isolation, or always in conjunction with other "cloud optimizing" steps like rechunking. 

EDIT: Here's a cluster link for the "small" version of this test https://cloud.coiled.io/clusters/594106/account/dask-engineering/information. It takes ~20 minutes and costs ~$0.75

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,7 +31,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python_version: ["3.10"]
-        pytest_args: [tests/geospatial/test_netcdf_to_zarr.py --ignore=tests/tpch]
+        pytest_args: [tests --ignore=tests/tpch]
         extra-env: [""]
         name_prefix: [tests]
         include:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,7 +31,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python_version: ["3.10"]
-        pytest_args: [tests --ignore=tests/tpch]
+        pytest_args: [tests/geospatial/test_netcdf_to_zarr.py --ignore=tests/tpch]
         extra-env: [""]
         name_prefix: [tests]
         include:
--- a/tests/geospatial/test_cloud_optimize.py
+++ b/tests/geospatial/test_cloud_optimize.py
@@ -62,10 +62,11 @@ def test_cloud_optimize(
             models = models[:1]
             variables = variables[:1]
         elif scale == "medium":
-            # 715 files. One model and all variables.
+            # 390 files. Two models and two variables.
             # Currently fails after hitting 20 minute idle timeout
-            # sending `to_zarr` graph to the scheduler.
-            models = models[:1]
+            # sending large graph to the scheduler.
+            models = models[:2]
+            variables = variables[:2]
         else:
             # 11635 files. All models and variables.
             pass
--- a/tests/geospatial/test_cloud_optimize.py
+++ b/tests/geospatial/test_cloud_optimize.py
@@ -7,7 +7,7 @@ def test_cloud_optimize(
     s3_url,
     client_factory,
     cluster_kwargs={
-        "workspace": "dask-engineering",
+        "workspace": "dask-benchmarks",
         "region": "us-west-2",
         "wait_for_workers": True,
     },
@@ -81,6 +81,7 @@ def test_cloud_optimize(
         print(f"Processing {len(files)} NetCDF files")
 
         # Load input NetCDF data files
+        # TODO: Reduce explicit settings once https://github.com/pydata/xarray/issues/8778 is completed.
         ds = xr.open_mfdataset(
             files,
             engine="h5netcdf",
--- a/tests/geospatial/test_cloud_optimize.py
+++ b/tests/geospatial/test_cloud_optimize.py
@@ -86,6 +86,9 @@ def test_cloud_optimize(
             engine="h5netcdf",
             combine="nested",
             concat_dim="time",
+            data_vars="minimal",
+            coords="minimal",
+            compat="override",
             parallel=True,
         )
 
--- a/tests/geospatial/test_cloud_optimize.py
+++ b/tests/geospatial/test_cloud_optimize.py
@@ -1,8 +1,7 @@
 import xarray as xr
-from dask.utils import format_bytes
 
 
-def test_netcdf_to_zarr(
+def test_cloud_optimize(
     scale,
     s3,
     s3_url,
@@ -81,6 +80,7 @@ def test_netcdf_to_zarr(
         files = [s3.open(f) for f in file_list]
         print(f"Processing {len(files)} NetCDF files")
 
+        # Load input NetCDF data files
         ds = xr.open_mfdataset(
             files,
             engine="h5netcdf",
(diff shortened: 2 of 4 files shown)


tests/geospatial/test_netcdf_to_zarr.py
```diff
@@ -0,0 +1,93 @@
...
+
+        ds = xr.open_mfdataset(
+            files,
+            engine="h5netcdf",
+            combine="nested",
+            concat_dim="time",
+            parallel=True,
+        )
+        print(f"Converting {format_bytes(ds.nbytes)} from NetCDF to Zarr")
+        ds.to_zarr(s3_url)
```
jrbourbeau, line 93: @maxrjones thanks for pointing to this dataset over in https://github.com/coiled/benchmarks/discussions/1545#discussioncomment-10681943. Does this looks like what you've seen in the wild?

.github/workflows/tests.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python_version: ["3.10"]
-        pytest_args: [tests --ignore=tests/tpch]
+        pytest_args: [tests/geospatial/test_netcdf_to_zarr.py --ignore=tests/tpch]
```
jrbourbeau, line 34: TODO: Revert before merging

tests/geospatial/test_netcdf_to_zarr.py
jrbourbeau: I've gone ahead and added a rechunking step (from "pancake" to "pencil" chunks), which seems to be pretty common when cloud-optimizing a NetCDF dataset to Zarr

maxrjones: Yeah, I think it's right that when cloud-optimizing it's usually best to include a rechunking step. But, there's many cases in which people avoid the cloud-optimization step and do subsequent analyses on the data loaded directly from the original NetCDF files.

tests/geospatial/test_cloud_optimize.py
```diff
@@ -0,0 +1,96 @@
...
+        files = [s3.open(f) for f in file_list]
+        print(f"Processing {len(files)} NetCDF files")
+
+        # Load input NetCDF data files
+        ds = xr.open_mfdataset(
+            files,
+            engine="h5netcdf",
+            combine="nested",
+            concat_dim="time",
+            parallel=True,
```
dcherian, line 89:
```suggestion
            parallel=True, data_vars="minimal", coords="minimal", compat="override",
```

These *may* be needed for decent perf, I haven't looked at the files to be sure

tests/geospatial/test_netcdf_to_zarr.py
jrbourbeau: > Yeah, I think it's right that when cloud-optimizing it's usually best to include a rechunking step.

Awesome, thanks for confirming that's usually the case 👍 

> But, there's many cases in which people avoid the cloud-optimization step and do subsequent analyses on the data loaded directly from the original NetCDF files

I guess in this case, the "subsequent analyses" is just writing to Zarr. Do you think this still captures user pain well? FWIW my experience is using `xr.open_mfdataset(..., parallel=True)` + any other step performs very poorly

tests/geospatial/test_cloud_optimize.py
dcherian: ^ yes, they are basically mandatory

jrbourbeau: Cool, I'll take them for a spin and update here. Also, should those be the default in `xarray`?

dcherian: https://github.com/pydata/xarray/issues/8778

tests/geospatial/test_cloud_optimize.py
```diff
@@ -0,0 +1,99 @@
...
+
+        if scale == "small":
+            # 130 files (152.83 GiB). One model and one variable.
+            models = models[:1]
+            variables = variables[:1]
+        elif scale == "medium":
+            # 715 files. One model and all variables.
+            # Currently fails after hitting 20 minute idle timeout
+            # sending `to_zarr` graph to the scheduler.
+            models = models[:1]
```
hendrikmakait, line 68: Is there a difference here in how we scale with respect to models and variables? I'm wondering if it would make more sense to pick a subset (larger than one) of both models and variables instead.

tests/geospatial/test_cloud_optimize.py
```diff
@@ -0,0 +1,99 @@
...
+        file_list = []
+        for model in models:
+            for variable in variables:
+                data_dir = f"s3://nex-gddp-cmip6/NEX-GDDP-CMIP6/{model}/historical/r1i1p1f1/{variable}/*.nc"
+                file_list += [f"s3://{path}" for path in s3.glob(data_dir)]
+        files = [s3.open(f) for f in file_list]
+        print(f"Processing {len(files)} NetCDF files")
+
+        # Load input NetCDF data files
+        ds = xr.open_mfdataset(
```
hendrikmakait, line 84: nit: 
```suggestion
        # TODO: Reduce explicit settings once https://github.com/pydata/xarray/issues/8778 is completed.
        ds = xr.open_mfdataset(
```

tests/geospatial/test_cloud_optimize.py
```diff
@@ -0,0 +1,99 @@
+import xarray as xr
+
+
+def test_cloud_optimize(
+    scale,
+    s3,
+    s3_url,
+    client_factory,
+    cluster_kwargs={
+        "workspace": "dask-engineering",
```
hendrikmakait, line 10:
```suggestion
        "workspace": "dask-benchmarks",
```

hendrikmakait (approved):
Thanks, @jrbourbeau, this is good to go once the workspace has been adjusted. I have one question regarding scaling this benchmark but that's non-blocking.

tests/geospatial/test_cloud_optimize.py
jrbourbeau: 👍

tests/geospatial/test_netcdf_to_zarr.py
maxrjones: > I guess in this case, the "subsequent analyses" is just writing to Zarr. Do you think this still captures user pain well? FWIW my experience is using xr.open_mfdataset(..., parallel=True) + any other step performs very poorly

Yeah, it could make sense to focus first on the simplest operations that perform poorly. You mention that going from pancake to pencil chunks is a common step that performs poorly. That's true and is the motivation for the rechunker library. The worst performance would likely be seen when combining operations that perform optimally on pancake oriented chunks with operations that perform optimally on churro oriented chunks.

jrbourbeau: > pancake oriented chunks...churro oriented chunks

💯

PR #1552, opened 2024-09-20
Add benchmark for climatology

### LINKED ISSUE #1549: Climatology

hendrikmakait:
Closes #1549 

This benchmark doesn't even run at the small scale yet.

### DIFF
--- a/tests/geospatial/test_climatology.py
+++ b/tests/geospatial/test_climatology.py
@@ -112,7 +112,7 @@ def test_rechunk_map_blocks(
         ds = ds[variables].sel(time=time_range)
 
         ds = ds.drop_vars([k for k, v in ds.items() if "time" not in v.dims])
-        pencil_chunks = {"time": -1, "longitude": "128", "latitude": "128"}
+        pencil_chunks = {"time": -1, "longitude": "auto", "latitude": "auto"}
 
         working = ds.chunk(pencil_chunks)
         hours = xr.DataArray(range(0, 24, 6), dims=["hour"])
@@ -143,7 +143,7 @@ def test_highlevel_api(
         "workspace": "dask-benchmarks-gcp",
         "region": "us-central1",
         "wait_for_workers": True,
-        "idle_timeout": "2h",
+        "idle_timeout": "1h",
     },
     scale_kwargs={
         "small": {"n_workers": 10},
--- a/tests/geospatial/test_climatology.py
+++ b/tests/geospatial/test_climatology.py
@@ -1,0 +1,117 @@
+"""This benchmark is a port of the climatology computation implemented in
+https://github.com/google-research/weatherbench2/blob/47d72575cf5e99383a09bed19ba989b718d5fe30/scripts/compute_climatology.py
+with the parameters
+
+FREQUENCY = "hourly"
+HOUR_INTERVAL = 6
+WINDOW_SIZE = 61
+STATISTICS = ["mean"]
+METHOD = "explicit"
+"""
+
+import numpy as np
+import pytest
+import xarray as xr
+from coiled.credentials.google import CoiledShippedCredentials
+
+
+def compute_hourly_climatology(
+    ds: xr.Dataset,
+) -> xr.Dataset:
+    hours = xr.DataArray(range(0, 24, 6), dims=["hour"])
+    window_weights = create_window_weights(61)
+    return xr.concat(
+        [compute_rolling_mean(select_hour(ds, hour), window_weights) for hour in hours],
+        dim=hours,
+    )
+
+
+def compute_rolling_mean(ds: xr.Dataset, window_weights: xr.DataArray) -> xr.Dataset:
+    window_size = len(window_weights)
+    half_window_size = window_size // 2  # For padding
+    stacked = xr.concat(
+        [
+            replace_time_with_doy(ds.sel(time=str(y)))
+            for y in np.unique(ds.time.dt.year)
+        ],
+        dim="year",
+    )
+    stacked = stacked.fillna(stacked.sel(dayofyear=365))
+    stacked = stacked.pad(pad_width={"dayofyear": half_window_size}, mode="wrap")
+    stacked = stacked.rolling(dayofyear=window_size, center=True).construct("window")
+    return stacked.weighted(window_weights).mean(dim=("window", "year"))
+
+
+def create_window_weights(window_size: int) -> xr.DataArray:
+    """Create linearly decaying window weights."""
+    assert window_size % 2 == 1, "Window size must be odd."
+    half_window_size = window_size // 2
+    window_weights = np.concatenate(
+        [
+            np.linspace(0, 1, half_window_size + 1),
+            np.linspace(1, 0, half_window_size + 1)[1:],
+        ]
+    )
+    window_weights = window_weights / window_weights.mean()
+    window_weights = xr.DataArray(window_weights, dims=["window"])
+    return window_weights
+
+
+def replace_time_with_doy(ds: xr.Dataset) -> xr.Dataset:
+    """Replace time coordinate with days of year."""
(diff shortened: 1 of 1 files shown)


tests/geospatial/test_climatology.py
```diff
@@ -0,0 +1,120 @@
...
+    working = ds.chunk(pencil_chunks)
+    hours = xr.DataArray(range(0, 24, 6), dims=["hour"])
+    daysofyear = xr.DataArray(range(1, 367), dims=["dayofyear"])
+    template = (
+        working.isel(time=0)
+        .drop_vars("time")
+        .expand_dims(hour=hours, dayofyear=daysofyear)
+        .assign_coords(hour=hours, dayofyear=daysofyear)
+    )
+    working = working.map_blocks(compute_hourly_climatology, template=template)
```
hendrikmakait, line 111: I wonder if people typically compute climatology with complex-enough calculations that warrant the use of `map_blocks` or typically use calculations that leverage Xarray's higher-level API.

dcherian: People use rechunk+map_blocks because it usually would not work otherwise. It's basically "manual fusion of tasks" to not overwhelm the scheduler. And convenient because you can use Xarray API in there.

I *strongly* recommend running a version of this without those steps. Ideally, these tricks should not be necessary for this calculation.

hendrikmakait: Just to make sure: What I'm hearing you say is: The computations people generally execute for calculating climatology, be it `mean`, `std`, `quantile` (or `SEEP`?) can be expressed in Xarray's high-level API and do not fundamentally require a custom function mapped over all blocks. 

In that case, I fully agree that we should avoid `map_blocks` but express these calculations in Xarray. We should probably benchmark both a computation based on a decomposable aggregation like `mean` and one based on a holistic aggregation like `quantile`.

dcherian: > The computations people generally execute for calculating climatology,

Yes. generally. The trick is to look at the function being mapped: in this case it's all Xarray ops and no for loops. That indicates the map_blocks is a workaround for some Xarray and/or dask/distributed inefficiency.

> We should probably benchmark both a computation based on a decomposable aggregation like mean and one based on a holistic aggregation like quantile.

Yes totally agree with adding quantile. Dask will force a rechunk :)

PS: What is `SEEP`?

PPS: I notice this is based on a Xarray-beam workload, it may be that rechunk+map_blocks is the only way to express this in that framework.

hendrikmakait: > Yes. generally. The trick is to look at the function being mapped: in this case it's all Xarray ops and no for loops. That indicates the map_blocks is a workaround for some Xarray and/or dask/distributed inefficiency.

Thanks for the clarification.

> PS: What is `SEEP`?

All I can offer is https://github.com/google-research/weatherbench2/blob/47d72575cf5e99383a09bed19ba989b718d5fe30/scripts/compute_climatology.py#L147-L175. Maybe @shoyer can elaborate on `SEEP` and how common similar calculations are? Just looking at the code, this can probably be expressed in Xarray's high-level API as well.

> PPS: I notice this is based on a Xarray-beam workload, it may be that rechunk+map_blocks is the only way to express this in that framework.

Yes, that's what it's based on, but not what we should be limited by.

tests/geospatial/test_climatology.py
```diff
@@ -0,0 +1,120 @@
...
+def select_hour(ds: xr.Dataset, hour: int) -> xr.Dataset:
+    """Select given hour of day from Datset."""
+    # Select hour
+    ds = ds.isel(time=ds.time.dt.hour == hour)
+    # Adjust time dimension
+    ds = ds.assign_coords({"time": ds.time.astype("datetime64[D]")})
+    return ds
+
+
+@pytest.mark.client("compute_climatology")
```
jrbourbeau, line 77: I think this is missing an entry in the cluster yaml file. Mind switching to the `client_factory` approach similar to other geo benchmarks like this

https://github.com/coiled/benchmarks/blob/624979c99639e3d24040ab30273b8f381a6a2b59/tests/geospatial/test_zonal_average.py#L14-L23

My guess is we'll want different cluster sizes for small vs. large runs

tests/geospatial/test_climatology.py
shoyer: SEEPS is a score for precipitation forecasting. In this context, it's example of needing to calculate quantiles for every location.

In my opinion, rechunk + map_blocks is not just a performance optimization. In the simple cases, something like `ds.groupby(['time.month', 'time.hour']).mean()` would suffice, but there are also many calculations that one would like to perform at each location that are impractical to perform with Xarray's high-level API, e.g.,
- Calculating a weighted rolling-window average over day of year (this example from WeatherBench2).
- Calculate exact quantiles (what is the 99.9% extreme weather event?)
- Multi-variate linear regression (what is the climate trend at this location?)

You can look at the high-level Xarray code for calculating climatology from WeatherBench2. With some cleverness you could maybe optimize away the explicit Python loops, but it's intrinsically very complex. Despite the expense of rechunking, I have a hard time imagining any distributed computing framework getting better performance with series of explicit groupby and rolling operations. For operations like exact quantiles and linear regression I'm not even sure a direct distributed calculation is even possible.

So sure, let's include a simple climatology calculation implemented with `groupby(...).mean()` as an alternative (which I [also wrote](https://github.com/google/xarray-beam/blob/86c8f5b4f6c99324ee134a0e5668daa6ff87f407/examples/era5_climatology.py) for xarray-beam), but in my experience rechunk + map_blocks is the more common pattern.

dcherian: Yes, I agree with testing both use cases.

In my experience, less experienced users usually try without rechunk+map_blocks first, and then grudgingly resort to that. It'd be nice to model that interaction and improve things if possible.

shoyer: > In my experience, less experienced users usually try without rechunk+map_blocks first, and then grudgingly resort to that. It'd be nice to model that interaction and improve things if possible.

Yes, rechunk+map_blocks is definitely not the obvious solution. It would be quite nice if we could automatically optimize high-level Xarray code to use a rechunk + map_blocks style implementation when it will be more efficient.

hendrikmakait: I agree that we won't beat hand-optimized code. However, a major goal of these benchmarks is to understand how we can improve the performance of code that relies on high-level APIs to improve the end-user experience. We all seem to agree that it would be great if users wouldn't have to worry about rewriting their high-level code in a `rechunk+map_blocks`-style code if it can be avoided. So, wherever possible, I think we should try and benchmark code without optimization and analyze where and how this falls apart.

dcherian: I'll also add that we want `rechunk+map_blocks` to *always* succeed, even if slow, so that it can be a dependable fallback, particularly when the algorithms get more complicated. This would be a major improvement over the dask.array experience today.

hendrikmakait: @dcherian: Could you elaborate on that last comment? In what situations does that combination _not_ succeed today?

I can imagine this failing if the block sizes or the temporary memory footprint of the mapped function end up being too large. However, this seems like a situation that is beyond the control of higher-level APIs.

dcherian: It will fail whenever rechunking fails... agree that after that, the blockwise is pretty stable, which is why people use this solution in combination with some kind of batching to avoid the rechunking failure. I don't have an example at hand. This is anecdotal experience from looking at lot of non-expert code at NCAR.

hendrikmakait:
This should be good for a review now. (I expect that we will iterate on this once we have improved on some things within Xarray/Dask.)

jrbourbeau (approved):
Thanks @hendrikmakait

PR #1553, opened 2024-09-20
Use `dask-benchmarks` workspace

hendrikmakait:


hendrikmakait:
> I take it this is for tracking purposes?

You're correct! We'll eventually also have dedicated workspaces for Azure and GCP set up and should migrate the benchmarks running on those cloud providers at that point as well.

PR #1555, opened 2024-09-27
Use dedicated `dask-benchmarks-gcp` workspace

hendrikmakait:


jrbourbeau (approved):
Thanks @hendrikmakait

PR #1557, opened 2024-10-09
Add regridding benchmark

### LINKED ISSUE #1556: Regridding

hendrikmakait:
Closes #1556

### DIFF
--- a/AB_environments/AB_sample.conda.yaml
+++ b/AB_environments/AB_sample.conda.yaml
@@ -25,7 +25,7 @@ dependencies:
   - fsspec ==2024.6.1
   - s3fs ==2024.6.1
   - gcsfs ==2024.6.1
-  - pyarrow ==17.0.0
+  - pyarrow ==15.0.0
   - jupyterlab ==4.2.4
   - lz4 ==4.3.3
   - ipywidgets ==8.1.3
@@ -51,6 +51,9 @@ dependencies:
   - bokeh ==3.5.1
   - gilknocker ==0.4.1
   - openssl >1.1.0g
+  - rioxarray ==0.17.0
+  - h5netcdf ==1.3.0
+  - xesmf ==0.8.7
   # End copy-paste
 
   - pip:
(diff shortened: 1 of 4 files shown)


mrocklin:
Do we have enough context here to add this to the benchmark post?  If you give me bullet points and an image (if one makes sense) I'm happy to write up words.

hendrikmakait:
@aulemahal: Thanks for the input. I'll add a follow-up issue to look into some of the suggestions for increasing the complexity of this workload.

> IIUC, these benchmarks are more geared towards dask ? 

It is, but it's also aimed at reflecting real workloads. Would more complex methods also result in more complex computations or just in more complex weight generation? 

> We do have some code to make the weights generation in parallel but I would say it is still experimental and of limited scope.

Please let us know if we can help with anything from a Dask perspective.

@mrocklin:
> Do we have enough context here to add this to the benchmark post? If you give me bullet points and an image (if one makes sense) I'm happy to write up words.

I'll whip something up. At first glance, this benchmark seems to do alright; it's mostly an embarrassingly parallel computation. Performance and # of tasks could probably look better but that's already a lot better than some of the other benchmarks.

aulemahal:
> Would more complex methods also result in more complex computations or just in more complex weight generation?

Mostly more complex weight generation, which is totally on the ESMF side, so partly in C/Fortran I think. Maybe two very different grids (curvilinear ones for example) and more complex methods would make for weights with more connected nodes, but I don't think this would affect the computation so much.

hendrikmakait:
@slevang, thanks for the additional input! Would you be interested in contributing a benchmark implemented with `xarray-regrid`?

AB_environments/AB_sample.conda.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
   - fsspec ==2024.6.1
   - s3fs ==2024.6.1
   - gcsfs ==2024.6.1
-  - pyarrow ==17.0.0
+  - pyarrow ==15.0.0
```
phofl, line 28: why?

phofl (commented):
small comment, otherwise lgtm

AB_environments/AB_sample.conda.yaml
hendrikmakait: to keep this in sync with `environment.yml`

phofl: Ah sorry, unrelated change...

phofl (approved):

phofl:
thx

PR #1561, opened 2024-10-14
Don't store cluster dumps for benchmarks

hendrikmakait:
I don't think it's necessary that we store cluster dumps for every benchmark, so I've removed that functionality.


phofl (approved):
I think I have never looked at those, so fine by me

fjetter (approved):

PR #1562, opened 2024-10-14
Add missing bottleneck package to environments

hendrikmakait:
`bottleneck` is used in `tests/geospatial/test_climatology.py::test_highlevel_api`


phofl (approved):

phofl:
Generally fine, but this might have impact on xarray reducers more generally too

PR #1566, opened 2024-10-22
Fix output chunks

hendrikmakait:
#1565 fixed the wrong test.

### DIFF
--- a/tests/geospatial/test_climatology.py
+++ b/tests/geospatial/test_climatology.py
@@ -158,6 +158,7 @@ def test_highlevel_api(
         # Load dataset
         ds = xr.open_zarr(
             "gs://weatherbench2/datasets/era5/1959-2023_01_10-wb13-6h-1440x721.zarr",
+            chunks={"time": "auto"},
         )
 
         if scale == "small":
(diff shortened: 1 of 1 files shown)


tests/geospatial/test_climatology.py
```diff
@@ -157,6 +158,7 @@ def test_highlevel_api(
         # Load dataset
         ds = xr.open_zarr(
             "gs://weatherbench2/datasets/era5/1959-2023_01_10-wb13-6h-1440x721.zarr",
+            chunks={"time": "auto"},
```
phofl, line 161: is this intended?

tests/geospatial/test_climatology.py
```diff
@@ -128,8 +129,8 @@ def test_rechunk_map_blocks(
         pancake_chunks = {
             "hour": 1,
             "dayofyear": 1,
-            "latitude": ds.chunks["latitude"],
-            "longitude": ds.chunks["longitude"],
+            "latitude": original_chunks["latitude"],
+            "longitude": original_chunks["longitude"],
```
hendrikmakait, line 133: I'm confused, isn't this on `main` already?

tests/geospatial/test_climatology.py
hendrikmakait: good catch

phofl (approved):

PR #1569, opened 2024-10-23
Allow collection of performance reports for geospatial benchmarks

hendrikmakait:
To use this, execute benchmarks with `--benchmark --performance-report`


phofl (approved):

phofl:
thanks

PR #1570, opened 2024-10-23
Fix cleanup of object-store directories

hendrikmakait:
"Folders" might not exist in object stores, so let's not fail when that's the case because no file has been added at that location.


phofl (approved):

phofl:
thanks

PR #1571, opened 2024-10-23
Use `client.wait_for_workers` instead of `wait_for_workers` kwarg

hendrikmakait:
This is a workaround for https://github.com/coiled/platform/issues/103


phofl (approved):

phofl:
Thanks

PR #1573, opened 2024-10-25
Add landsat data preparation benchmark

phofl:


tests/geospatial/test_landsat_analysis.py
```diff
@@ -0,0 +1,128 @@
...
+
+        if scale == "small":
+            bands = ["cloud_qa", "red", "green", "blue"]
+            box = (
+                10.305842676154782,
+                3.12634712296393502,
+                14.497958439404897,
+                5.052677334304235,
+            )
+        elif scale == "large":
```
hendrikmakait, line 51: We haven't really thought too much about scaling in the past. What's the significance of these scaling dimensions? Is it larger an increase data size and more (embarassingly parallel) variables?

hendrikmakait (changes requested):
Please separate query execution from workload definition (following the example of https://github.com/coiled/benchmarks/pull/1579.

PR #1576, opened 2024-10-29
Use standard dataset in regrid

hendrikmakait:
This uses the full ERA5 dataset with 37 levels and 1h granularity. This matches the annotated data sizes.


phofl (approved):

hendrikmakait:
Nope, this just changed the amount of processed data

PR #1581, opened 2024-11-04
Recalibrate scales of regridding benchmark

hendrikmakait:
For this benchmark, the most interesting scaling dimension is the size of the weight matrix which depends on the output resolution. We should, therefore, scale it along this dimension.

cc @phofl


fjetter (approved):
LGTM

PR #1589, opened 2024-11-11
Benchmark geospatial workload submission

hendrikmakait:


phofl:
What's the idea here? Checking how long it takes until the scheduler does something?

hendrikmakait:
> What's the idea here? Checking how long it takes until the scheduler does something?

Yeah, the idea is to have the ability to benchmark how long it takes for the computation to show up on the scheduler without running the entire (possibly long) workload and using fewer resources.

PR #1594, opened 2024-11-13
Don't parallelize geospatial benchmarks

hendrikmakait:
I've run into CPU limits when running multiple tests on `medium` scale in parallel, so let's avoid running benchmarks simultaneously for now.

### DIFF
--- a/.github/workflows/geospatial.yml
+++ b/.github/workflows/geospatial.yml
@@ -69,7 +69,6 @@ jobs:
       run: |
         pytest --benchmark \
             tests/geospatial -m geo_execution \
-            -n 4 --dist loadscope \
             --scale ${{ inputs.scale }} \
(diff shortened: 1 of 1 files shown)


phofl:
what type of cpu limits?

availability?

hendrikmakait:
Coiled User CPU limits for now:

```
FAILED tests/geospatial/test_climatology.py::test_highlevel_api[execution] - coiled.errors.ServerError: 
User CPU-core limit reached.

	Limit:          1000 CPUs
	Already Using:  808 CPUs
	Requested:      404 CPUs (for 101 instances)
```

I have no clue what actual limits on the individual cloud accounts are, so I'd rather get one set of benchmarks down in a little more time and then figure out parallelism afterward.

PR #1602, opened 2024-11-14
Fix report step on CI

hendrikmakait:


tests/tpch/conftest.py
```diff
@@ -30,9 +30,6 @@ def pytest_addoption(parser):
     parser.addoption("--cloud", action="store_false", dest="local", help="")
     parser.addoption("--restart", action="store_true", default=True, help="")
     parser.addoption("--no-restart", action="store_false", dest="restart", help="")
-    parser.addoption(
-        "--performance-report", action="store_true", default=False, help=""
-    )
     parser.addoption(
```
jrbourbeau, line 33: Not clear to me why this is being removed, but that's okay. Just wanted to make sure this is intentional is all

.github/workflows/tests.yml
```diff
@@ -262,7 +262,7 @@ jobs:
           script: |
             const workflow_url = `https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}`
             const issue_body = `[Workflow Run URL](${workflow_url})`
-            github.issues.create({
+            github.rest.issues.create({
                 owner: context.repo.owner,
                 repo: context.repo.repo,
                 body: issue_body,
```
jrbourbeau, line 268: Don't need to do anything here, but just wanted to note that the original version of this code has been moved into it's own GitHub action https://github.com/xarray-contrib/issue-from-pytest-log which is what we're now using in `dask/dask`

https://github.com/dask/dask/blob/98331028f8a51d404f8c9856d49ba726322c4e03/.github/workflows/upstream.yml#L76-L86

jrbourbeau: To be clear, this is mostly just an FYI

tests/tpch/conftest.py
hendrikmakait: This PR included https://github.com/coiled/benchmarks/pull/1599.

PR #1608, opened 2024-11-15
Auto-chunking for ERA5 loading

hendrikmakait:
This PR defaults to using `auto` chunks along the time dimension when loading the ERA5 dataset in geospatial benchmarks. Due to the small size of native inputs (4 MiB), this results in smaller graphs and (mostly) runtime improvements. Here are the runtimes for the small scale:

auto-chunking

The performance regression on `test_highlevel_api` is curious and should be investigated. Unfortunately, we do not measure task graph sizes, so I can't provide any numbers here.

### DIFF
--- a/tests/geospatial/workloads/climatology.py
+++ b/tests/geospatial/workloads/climatology.py
@@ -72,6 +72,7 @@ def rechunk_map_blocks(
     # Load dataset
     ds = xr.open_zarr(
         "gs://weatherbench2/datasets/era5/1959-2023_01_10-wb13-6h-1440x721.zarr",
+        chunks={"time": "auto"},
     )
 
     if scale == "small":
@@ -122,6 +123,7 @@ def highlevel_api(
     # Load dataset
     ds = xr.open_zarr(
         "gs://weatherbench2/datasets/era5/1959-2023_01_10-wb13-6h-1440x721.zarr",
+        chunks={"time": "auto"},
     )
 
     if scale == "small":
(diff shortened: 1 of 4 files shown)


tests/geospatial/workloads/climatology.py
```diff
@@ -72,6 +72,7 @@ def rechunk_map_blocks(
     # Load dataset
     ds = xr.open_zarr(
         "gs://weatherbench2/datasets/era5/1959-2023_01_10-wb13-6h-1440x721.zarr",
+        chunks={"time": "auto"},
```
jrbourbeau, line 75: This the a case where I think the naive user-code is what we want. This may be _slightly_ better but we want to optimize what most users will have in practice

phofl:
it is as far as I understand, auto means whatever chunks zarr has because our auto chunking was kind of stupid in the past

hendrikmakait:
We don't have to merge this PR for now, but it showcases what @phofl and I believe would be a better default in general. This PR can be used to highlight benefits or complications on the way to establishing this. The biggest benefit is likely in graph sizes at larger scales, but we don't measure this and still run into a variety of other errors so that workloads still don't work reliably end-to-end.

jrbourbeau:
> it showcases what @phofl and I believe would be a better default in general

Ah, I see. I hadn't realized this was being used to help motivate a new default. Thanks for clarifying that

dcherian:
You guys are seeing that classic problem. There's no good chunking that works well for both problems, indeed i'd recommend running each workload with two orthogonal chunking strategies. The thing to fix is to make sure the workload runs in any case, even if its slow.

hendrikmakait:
Closing as stale

PR #1613, opened 2024-11-25
Use string paths with ``open_mfdataset``

jrbourbeau:
We can just use simple string paths now (instead of manually opening files) with the latest ``xarray`` release


hendrikmakait (approved):
Thanks, @jrbourbeau!

PR #1630, opened 2024-12-16
Update miniconda to fix CI

hendrikmakait:
https://github.com/coiled/benchmarks/actions/runs/12349802479

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,6 +71,7 @@ jobs:
           condarc-file: ci/condarc
           python-version: ${{ matrix.python_version }}
           environment-file: ci/environment.yml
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up environment
-        uses: conda-incubator/setup-miniconda@v3.0.4
```
hendrikmakait, line 67: This is the relevant change, the rest is just pinning consistently to the minor version.

hendrikmakait: I have overestimated the capabilities of this action. Consistently pinning to v3 now.

phofl (approved):

PR #1646, opened 2025-01-07
Fix tests matrix

hendrikmakait:
Apparently we need to specify _all_ values: https://github.com/coiled/benchmarks/actions/runs/12635100154/workflow


hendrikmakait:
Looks like this works as expected: https://github.com/coiled/benchmarks/actions/runs/12651384181

PR #1648, opened 2025-01-09
Purge `dask-expr`

hendrikmakait:
Follow-up to https://github.com/dask/dask/pull/11623
Closes https://github.com/coiled/benchmarks/issues/1647

The only thing I haven't purged is the `dask_expr_version` column from the runs.

### DIFF
--- a/tests/tpch/conftest.py
+++ b/tests/tpch/conftest.py
@@ -7,7 +7,6 @@ from contextlib import nullcontext
 
 import coiled
 import dask
-import dask_expr
 import distributed
 import filelock
 import pytest
@@ -142,7 +141,6 @@ def tpch_database_table_schema(request, testrun_uid, scale, query, local):
         originalname=request.node.originalname,
         path=str(request.node.path.relative_to(TEST_DIR)),
         dask_version=dask.__version__,
-        dask_expr_version=dask_expr.__version__,
         distributed_version=distributed.__version__,
         python_version=".".join(map(str, sys.version_info)),
         platform=sys.platform,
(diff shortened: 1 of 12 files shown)


tests/tpch/conftest.py
```diff
@@ -142,7 +141,6 @@ def tpch_database_table_schema(request, testrun_uid, scale, query, local):
         originalname=request.node.originalname,
         path=str(request.node.path.relative_to(TEST_DIR)),
         dask_version=dask.__version__,
-        dask_expr_version=dask_expr.__version__,
```
fjetter, line 145: this is a DB change. I'm not sure if any of our CI jobs reacts to this. Technically, this would require a DB migration (but I'm also fine with just setting this to null)

hendrikmakait: If I understand correctly, this would just set the `TPCHRun.dask_expr_version` to `None`, which is allowed: https://github.com/coiled/benchmarks/blob/615ba99c901745b58042f287ef73ddd2dc169b00/benchmark_schema.py#L86. To change the actual data model, we do indeed need a migration and should remove the linked line from the ORM model.

hendrikmakait: > I'm not sure if any of our CI jobs reacts to this.

If you're concerned with CI jobs accessing the attribute, this PR should remove all of those instances.

fjetter: ok, that's what I wanted. Thanks for checking

fjetter (approved):

phofl (approved):

